Issue #9, opened 2020-02-26
Upload a file to a PAS package


hannahkates:
Here's a link to the existing Applicant Portal if it might help inform approaches: https://dcppfsuat2.dynamics365portals.us/applicants/overview/application-status-new/

bfreeds:
Komal gave me a walkthrough yesterday of the Artifacts/Documents features in CRM and in the Microsoft Applicant Portal.  
# Artifacts 
> Artifacts are a data construct in CRM that represent documents.  There are two types of artifacts (project-related and package-related).    

### ZAP CRM  
#### Package Artifacts
1. View
![image](https://user-images.githubusercontent.com/5316367/75799831-a248df00-5d46-11ea-9c95-97cbea018540.png)
1. Edit
Click on existing Package's artifact
![image](https://user-images.githubusercontent.com/5316367/75810366-746c9600-5d58-11ea-9a99-8a05bc15aaca.png)
Planner can edit by changing values in these fields
![image](https://user-images.githubusercontent.com/5316367/75810427-90703780-5d58-11ea-8d32-7013cc003fbd.png)

1. Create 
  - Creating a "DCP Document", a document a planner would make
  - ![image](https://user-images.githubusercontent.com/5316367/75810890-72ef9d80-5d59-11ea-96fa-5c6231f8da0a.png)

1.  Delete
Clicking "Deactivate" in the menu bar gives this modal & options for deletion.  Will want to make sure our api query filters out for anything that's been deactivated.
![image](https://user-images.githubusercontent.com/5316367/75811996-708e4300-5d5b-11ea-8444-15c19cdc7309.png)

### Existing Applicant Portal
#### Package Artifacts 
1. View
1. Edit
![image](https://user-images.githubusercontent.com/5316367/75800604-d4a70c00-5d47-11ea-86c9-a405ee2f902b.png)  


The following fields are read-only:  
  - `Version`
  - `Project`
  - `Package`
  - `Package Type` and `File Creator` have dropdowns the user can click, but there aren't any options to select from.  
    - Are there other cases when there would be choose-able options here?
  - `Documents`
    - Can click add file, select file, and save.
    - This uploads to Sharepoint with the following URL:
[...]

PR #30, opened 2020-03-19
9 upload doc to pas package

godfreyyeung:
This PR/commit acts as a reference to address #9 

This is the minimal set of modules, controllers and services to connect to CRM and make a document upload.

Most of it was taken from `labs-zap-api` and stripped of unecessary imports.


server/src/document/document.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+    } else if ( entityName === 'dcp_package' ) {
+      const packageGUID = instanceId;
+
+      const packageRecord = (await CRMWebAPI.get(`dcp_packages?$select=dcp_name&$filter=dcp_packageid eq '${packageGUID}'&$top=1`))['value'][0];
+
+      const packageName = packageRecord.dcp_name;
+
+      const folderName = `${packageName}_${packageGUID.replace(/\-/g,'').toUpperCase()}`;
+
+      uploadDocResponse = await CRMWebAPI.uploadDocument('dcp_artifacts', packageGUID, folderName, file.originalname, encodedBase64File, true, headers);
```
godfreyyeung: The first parameter to uploadDocument should be `dcp_packages`

godfreyyeung:
> This is awesome! Would you mind adding a more in-depth description on the PR? Having trouble knowing where to start/would love an explanation of how this was implemented file-by-file etc. Even if it is the same as zap-api, I think it would help a lot in reviewing

can do!

server/src/app.module.ts
```diff
@@ -1,10 +1,55 @@
...
 @Module({
-  imports: [],
+  imports: [
+    ConfigModule,
+    AuthModule,
+    DocumentModule,
+    TypeOrmModule.forRootAsync({
+      useFactory: (config: ConfigService) => ({
+        type: 'postgres',
+        url: config.get('DATABASE_URL'),
```
allthesignals, line 22: This doesn't need to be in here. This is no database for this API.

server/src/config/config.module.ts
```diff
@@ -0,0 +1,13 @@
+import { Module } from '@nestjs/common';
+import { ConfigService } from './config.service';
+
+@Module({
+  providers: [
+    {
+      provide: ConfigService,
+      useValue: new ConfigService(`development.env`),
```
allthesignals, line 8: Unnecessary template literal.

server/src/app.controller.ts
```diff
@@ -1,12 +1,38 @@
...
+    } catch (e) {
+      if (e instanceof HttpException) {
+        res.status(401).send({ errors: [e] });
+      } else {
+        console.log(e);
+
+        res.status(500).send({ errors: [e] });
+      }
+    }
+  }
```
allthesignals, line 36: Why is this here? If this is for document upload.

server/src/auth.middleware.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    try {
+      req.session = await this.authService.validateCurrentToken(token);
+
+      next();
+    } catch (e) {
+      proceedNoAuth(res, next);
+    }
+  }
+}
```
allthesignals, line 26: The PR says it is intended to address document upload (which was a research task, not really something to build out), but this is authentication logic. This seems unrelated to the issue at hand.

server/src/auth/auth.module.ts
```diff
@@ -0,0 +1,54 @@
...
+    setTimeout(async () => {
+      console.log('generating cookie...');
+
+      const response = await superagent.get(`http://localhost:3000/login?accessToken=${mockJWT}`);
+      const { header: { 'set-cookie': [, token] } } = response;
+
+      console.log(`SKIP_AUTH is true! The cookie token is ${token}. Add this to your request headers.`);
+    }, 3000);
+  }
+}
```
allthesignals, line 54: This was lifted from ZAP API to assist with development but is it really necessary for right now?

server/src/contact/contact.entity.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @Column()
+  fullname: string;
+
+  @Column()
+  statecode: string;
+
+  @Column()
+  statuscode: string;
+}
```
allthesignals, line 19: Again, this is a TypeORM object, which is specifically meant for modeling SQL database objects. There is no SQL database for this project. We wouldn't even need to use this.

server/src/contact/contact.service.ts
```diff
@@ -0,0 +1,29 @@
...
+  async findOne(opts: any): Promise<Contact> {
+    return this.contactRepository.findOneOrFail(opts);
+  }
+
+  async findByEmail(email): Promise<Contact> {
+    return this.activeContacts()
+      .andWhere('emailaddress1 ILIKE :email', { email })
+      .getOne();
+  }
+}
```
allthesignals, line 29: None of this would be necssary or woudl work because we aren't using a SQL data ase for this. THis is all TypeORM stuff which is specific to SQL datbases, not WEb API

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+      MSCRMCallerID: this.config.get('CRM_ADMIN_SERVICE_USER')
+    };
+
+    // encode base64
+    const encodedBase64File = Buffer.from(file.buffer).toString('base64');
+
+    let uploadDocResponse = {};
+
+    if ( entityName === 'dcp_package' ) {
+      const packageGUID = instanceId;
```
allthesignals, line 59: Why make another variable for this? If you need to alias it, you can do that in the destructuring above:

```
      body: {
        instanceId: packageGUID,
```

I think a comment would be fine, though.

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+    } = request;
+
+    const headers = {
+      MSCRMCallerID: this.config.get('CRM_ADMIN_SERVICE_USER')
+    };
+
+    // encode base64
+    const encodedBase64File = Buffer.from(file.buffer).toString('base64');
+
+    let uploadDocResponse = {};
```
allthesignals, line 56: Why is this defined up here? If the await fails below, it will throw a regular error anyway. And if needs to be a default object, that could be handled in a clearer way. I would just define a variable below

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    if ( entityName === 'dcp_package' ) {
+      const packageGUID = instanceId;
+
+      const packageRecord = (await CRMWebAPI.get(`dcp_packages?$select=dcp_name&$filter=dcp_packageid eq '${packageGUID}'&$top=1`))['value'][0];
+
+      const packageName = packageRecord.dcp_name;
```
allthesignals, line 63: You could destructure this:

`const { dcp_name: packageName } = (await ...`

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const folderName = `${packageName}_${packageGUID.replace(/\-/g,'').toUpperCase()}`;
+
```
allthesignals, line 66: I would have put stuff like this in a separate function below `generateFolderName(dcp_name, instanceId)`

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+      uploadDocResponse = await CRMWebAPI.uploadDocument('dcp_package', packageGUID, folderName, file.originalname, encodedBase64File, true, headers);
```
allthesignals, line 67: Why not destructure attrs out of file, too? We do it for the request object above, and file is provided right there in the method signature.

godfreyyeung:
No longer relevant to main stream of development work. Will pull out document research work for later feature branches. Thank you @allthesignals  for all the feedback!

PR #46, opened 2020-04-06
add login Sass module

andycochran:
This PR adds a Sass module for the login styles. 

When the login component is built out it will need to create a list item in the main menu like this: 

```
<li class="auth--container">
  <FaIcon @icon="user-circle" @prefix="fas" @size="2x" class="auth--icon" />
  <div>
    <span class="auth--name">
      email@example.org
    </span>
    <a href="/" class="auth--logout-button">
      Sign Out
    </a>
  </div>
</li>
```


andycochran:
> Why this naming convention? `modules/_m-login.scss ` . The underscore, plus 'm', plus dash seem redundant because they are organized under the modules folder which indicates that they're modules.

The leading underscore tells Sass not to process the file. 

I suppose we could ditch the "m-". It's just a convention I picked up for indicating the purpose of a Sass file (m=module, l=layout, t=typography). But yeah, it's not necessary. 

> What is the difference between `.auth--` prefix and something like, say, `.site-header`?

This is just a way of namespacing classes so it's obvious that the child elements are part of the component. And `.component--element` has less specificity than `.component .element`. 

I'll look into the Sass warning messages…

allthesignals:
@andycochran makes sense, thank you!  your call on naming, but imo less is more.

Issue #38, opened 2020-04-03
Frontend: Login functionality


godfreyyeung:
Handle redirects from NYCID, hit the backend API /login endpoint, store the returned cookie. 

Look examples of using ember-simple-auth in labs-zap-search

godfreyyeung:
In progress of writing /users endpoint

Issue #17, opened 2020-03-12
Create "My Projects" user dashboard


hannahkates:
AB#10519

## User Story:
As an applicant team member, I need to see all of my active projects and see which ones I can take action on (i.e. submit a PAS).

## Display Logic:

**Project shows up in the left "To DO" column with an "Edit Pre-Application Statement" button if**
PAS package `statuscode IN ('Saved', 'Package Preparation')`

**Project shows up in the right "Working on it..." column with a "View Pre-Application Statement" button if**
PAS package `statuscode IN ('Submitted', 'Under Review', 'Revision Required')`
^ all statuses except for 'Mistake'

**Project shows up in the right "Working on it..." column with no button if** 
Project is missing PAS data because the PAS package isn't visible or hasn't been created yet

**If user isn't assigned to any projects**
Page should display “You are not currently assigned to a DCP project. Please email ZAP_Feedback_DL@planning.nyc.gov for further assistance.” 

## [Figma Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP-Wireframes?node-id=0%3A1)
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/33a6f149-e556-410a-97a4-7dd84e45e534)

bfreeds:
@hannahkates @andycochran

PR #52, opened 2020-04-06
17:  Applicant Projects Dashboard

bfreeds:
This PR addresses #17 by:
- Creating a /projects route
- Sorting projects data based on statuscode (in the projects controller)

**To Do**
- Data comes from mirage scenario.  Needs to be hooked up to api
- Breadcrumb is hard-coded.  Needs to be dynamically generated.
- _Note:_ The header login doesn't make sense in the screenshots below (this will be shown for a user when they've logged in) -- being addressed in PR #38.

**With Projects in each list**
image

**With Empty Projects Lists**
image


client/app/templates/projects.hbs
```diff
@@ -0,0 +1,89 @@
+<div class="site-main large-padding-bottom">
+    <section class="grid-container large-padding-top large-padding-bottom">
+
+    <div class="grid-x grid-padding-x">
+        {{!-- TODO:  Make this breadcrumb dynamic with application routing --}}
```
allthesignals, line 5: "TODO" is a great way to earmark something quickly!

PR #55, opened 2020-04-06
Create login functionality

allthesignals:
This commit creates a sign-in component, which encapsulates all the login behavior. Tests are created for this, including acceptance and integration.

A new package, window mock, is included to assist with testing behavior that depends on redirects.

# Conflicts:
#	client/app/authenticators/zap-api-authenticator.js
#	client/app/templates/application.hbs


client/app/models/user.js
```diff
@@ -0,0 +1,5 @@
+import Model from '@ember-data/model';
+
+export default class UserModel extends Model {
+
```
godfreyyeung, line 4: Will need attributes:
- contactId <-- camelcased here, I think
- emailaddress1

client/app/router.js
```diff
@@ -9,4 +9,5 @@ export default class Router extends EmberRouter {
 
 Router.map(function() { // eslint-disable-line
   this.route('projects');
```
godfreyyeung, line 11: wondering if we can nest authenticated routes under 
```
  this.route('authenticated', { path: '' }, function() {
```

client/app/models/user.js
allthesignals: yup! not camelcase tho

client/app/router.js
allthesignals: great idea. enhancement for later?

client/app/templates/application.hbs
```diff
@@ -5,15 +5,26 @@
...
+      <li>
+        {{#if this.session.isAuthenticated}}
+          <button {{on "click" this.invalidateSession}}
+            href="#"
+            type="button"
+          >
+            Logout
+          </button>
+        {{else}}
+          <SignInButton />
```
bfreeds, line 25: @allthesignals I think the layout in the sign-in button component needs to be adjusted:
image

allthesignals: bah crap. it needs to be a button element bc of accessibbility. can't be anchor.

allthesignals: @andycochran need some help with this — this is styled for anchor tags, but we need to use button elements bc you can't bind actions to anchor elements. thoughts?

I'm going to move into new issue so we can move on

PR #58, opened 2020-04-07
38 login integration

allthesignals:
This PR integrates the client and the server, and adds a few necessary features for local development. To review more easily, I'd suggest going commit by commit, starting [here](https://github.com/NYCPlanning/labs-applicant-portal/pull/58/commits/db9da170ed8062dac7f97bfdc30ab3fe9e291ceb).

It enables compression, JSON:API parsing, and CORS, and allows the Ember app to accept a `HOST=` environment variable: `HOST=https://locahost:3000 yarn run start:ssl`.

You will notice that SSL and HTTPS are present! This PR enables the ability to develop locally with SSL. This is becoming more and more necessary as Chrome and others tighten restrictions on cross-origin cookies over non secure domains.

See instructions:

> # SSL Setup in Local Development
> In order for cookies to work across both the client and the server during local development, we must have SSL setup for our development environment. Here's how to set that up:
> 
> 1. Follow this step-by-step guide: https://www.freecodecamp.org/news/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec/. It will generate some files.
> 
> 2. Grab two files from step 1, `server.key` and `server.crt`. Copy and paste both files into BOTH `ssl/` folders under /client and /server.
> 
> 3. Run the server as you normally would. With SSL, the ember app should be run as `yarn run start:ssl`.
> 
> ## Why?
> Google Chrome is beginning to enforce some restrictions on particular uses of cookies in cross-origin contexts. In our case, our server provides a cookie from one domain and is expected to be sent across subsequent requests via another. The only way to simulate this behavior as it works on a production environment is to run our local development servers with SSL enabled.
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,13 @@
 # Labs Applicant Portal
+
+# SSL Setup in Local Development
+In order for cookies to work across both the client and the server during local development, we must have SSL setup for our development environment. Here's how to set that up:
+
+1. Follow this step-by-step guide: https://www.freecodecamp.org/news/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec/. It will generate some files.
+
+2. Grab two files from step 1, `server.key` and `server.crt`. Copy and paste both files into BOTH `ssl/` folders under /client and /server.
+
+3. Run the server as you normally would. With SSL, the ember app should be run as `yarn run start:ssl`.
+
+## Why?
+Google Chrome is beginning to enforce some restrictions on particular uses of cookies in cross-origon contexts. In our case, our server provides a cookie from one domain and is expected to be sent across subsequent requests via another. The only way to simulate this behavior as it works on a production environment is to run our local development servers with SSL enabled.
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ In order for cookies to work across both the client and the server during local 
 
 2. Grab two files from step 1, `server.key` and `server.crt`. Copy and paste both files into BOTH `ssl/` folders under /client and /server.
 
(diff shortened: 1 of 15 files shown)


client/app/adapters/application.js
```diff
@@ -3,4 +3,13 @@ import ENV from '../config/environment';
 
 export default class ApplicationAdapter extends JSONAPIAdapter {
   host = ENV.host;
+
+  // TODO: find better approach, this is a private method.
+  // see https://github.com/emberjs/data/issues/6413
+  // we must override so it includes the cookie
```
bfreeds, line 9: Really helpful comment here - appreciate the link to the issue.

README.md
```diff
@@ -1 +1,13 @@
 # Labs Applicant Portal
+
+# SSL Setup in Local Development
+In order for cookies to work across both the client and the server during local development, we must have SSL setup for our development environment. Here's how to set that up:
+
+1. Follow this step-by-step guide: https://www.freecodecamp.org/news/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec/. It will generate some files.
+
+2. Grab two files from step 1, `server.key` and `server.crt`. Copy and paste both files into BOTH `ssl/` folders under /client and /server.
+
+3. Run the server as you normally would. With SSL, the ember app should be run as `yarn run start:ssl`.
```
bfreeds, line 10: For future feature/developer experience improvement,  I wonder if we can use mkcert and docker-compose to automatically generate the certs needed for local development
https://github.com/FiloSottile/mkcert

Issue #14, opened 2020-03-12
Update deployment pipeline for QA team


hannahkates:
## User story:
As a QA analyst, I need an isolated, stable environment (no unannounced code changes) in which to perform QA testing.

Code should be manually deployed from Labs' development environment into the QA Team's environment at agreed upon intervals when code changes are ready for testing.

- Create branches on front end and API
- Rename branches
- Create netlify deployments and URLs

Diagram: https://docs.google.com/presentation/d/1rfC-jxB5Rh-d8gnP9CmDM_iVbeuKJ0ppDPSYLmz6HWI/edit#slide=id.g715440f704_0_0

bfreeds:
From @hannahkates, we need five environments / deployment pipelines:

| Git Branch        | Domain           | CRM Environment  |
| ------------- |-------------:| -----|
| feature branches      | (Netlify auto generated) | UAT2 |
| develop      | develop.applicants.planning.nyc.gov      |   UAT2 |
| qa | qa.applicants.planning.nyc.gov      |    UAT2 |
| staging | staging.applicants.planning.nyc.gov      |    UAT |
| master | applicants.planning.nyc.gov      |    PROD|

Our code will flow:
`feature` > `develop` > `qa` > `staging` > `master`

bfreeds:
- [ ] Domains & Subdomains Created
- [ ] Heroku Pipelines Configured for API Server deployments
- [ ] Netlify Configuration for frontend web app

bfreeds:
Created Heroku app in the "staging" pipeline for qa.  I don't see a way in Heroku to change the pipeline configurations from their existing buckets ("Review," "Staging", "Production"), so this will be an app within their "Staging" pipeline that auto deploys from the `qa` branch.   

I kept all the same config from the current staging deploy except for the `CRM_HOST` variable, changing `uat2` => `uat` in the url.

bfreeds:
Getting the user-friendly subdomains setup I think depends on #57.  In the meantime, we can use the default netlify URLs for qa.  

Backend API deployment pipelines are configured in Heroku.
Frontend deploys will be configured when #66 is merged.

Then we'll need to make a PR to merge `develop` => `qa` to trigger the initial deploy and make sure the qa team has the most up-to-date code to test from.  The URL will be:
**qa--applicant-portal.netlify.com**

After that happens and I can confirm that the deployments & urls are as expected I'll update the README with deployment documentation.

allthesignals:
[...]

PR #66, opened 2020-04-07
14:  Add netlify configuration

bfreeds:
This PR addresses #14 by adding the netlify configuration for:
- production
- staging
- qa
- develop

The `HOST` URLs referred to in this `netlify.toml` are heroku apps that have already been setup through the Heroku UI.  

image

Note:  Currently the Heroku develop and staging apps are both identical (didn't want to break it before the demo).

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,20 @@
+[build]
+  base= "client/"
+  command = "yarn build --environment=production"
+
+# production
+[context.master]
+  environment = { HOST="https://applicant-portal.herokuapp.com/" }
+
+# staging
+[context.staging]
+  environment = { HOST="https://applicant-portal-staging.herokuapp.com" }
+
+# qa team
+[context.qa]
+  environment = { HOST="https://applicant-portal-qa.herokuapp.com" }
+
+# develop
+[context.develop]
+  environment = { HOST="https://applicant-portal-develop.herokuapp.com" }
+
--- a/client/netlify.toml
+++ b/client/netlify.toml
@@ -1,8 +1,0 @@
-[build]
-command = "HOST=https://applicant-portal-staging.herokuapp.com yarn build --environment=production"
-
-[context.master]
-command = "HOST=https://applicant-portal-staging.herokuapp.com yarn build --environment=production"
-
-[context.develop]
-command = "HOST=https://applicant-portal-staging.herokuapp.com yarn build --environment=production"


bfreeds:
@allthesignals I was going to ask about that - reading the Netlify docs it sounded like the netlify.toml had to be in the git repository root.  If it can be in `client` then definitely makes more sense there, as it impacts the client deployment.  

I was mentioning Heroku just because that's the `host` config in the netlify contexts.

allthesignals:
> @allthesignals I was going to ask about that - reading the Netlify docs it sounded like the netlify.toml had to be in the git repository root. If it can be in `client` then definitely makes more sense there, as it impacts the client deployment.
> 
> I was mentioning Heroku just because that's the `host` config in the netlify contexts.

I see... it does seem to work, but maybe we should switch it. That's how it is configured in CEQR app: https://github.com/NYCPlanning/ceqr-app/blob/develop/netlify.toml

bfreeds:
Found that because in the UI we're telling Netlify to use `client` as the base directory it'll pick it up.  I think it keeps things more straightforward -- having something that's meant to deploy the client in the client folder, and also keeps the root of the repo uncluttered.  

I think for ceqr we should move netlify config into the "frontend" folder.

Updated this draft PR to keep the file in `client`
image

bfreeds:
@allthesignals after rereading the netlify [file config documentation](https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file), it seems like it is the convention to keep this file in the root of the repo.  

Backtracking on what I previously said, I think there's value in having it at root because it does crossover between frontend and backend responsibilities (with the `HOST` environment variable for the various environments being tied to the Heroku pipelines/deploys).

Note:  any config in `netlify.toml` will override what's setup in the UI.

netlify.toml
```diff
@@ -0,0 +1,20 @@
+[build]
+  base= "client/"
+  command = "yarn build --environment=production"
+
+# production
+[context.master]
+  environment = { HOST="https://applicant-portal.herokuapp.com/" }
```
allthesignals, line 7: Sorry, what is this syntax? I don't quite get how it works. Looks cool, though!

bfreeds: @allthesignals this is [TOML syntax](https://github.com/toml-lang/toml#inline-table), following the example Netlify has in their "file-based config" documentation.  

I reorganized the file to keep things that are consistent across deployments in the `[build]` section, so it's clear the only difference between deploys is the Heroku API deployment they each use.

PR #67, opened 2020-04-07
add loading state to projects view

andycochran:
image


allthesignals:
> $ eslint .
> 
> /home/circleci/repo/client/app/models/project.js
>   11:27  warning  There should be no line break before or after '='    operator-linebreak
>   13:17  error    Expected parentheses around arrow function argument  arrow-parens
>   19:12  error    Unnecessary 'else' after 'return'                    no-else-return
> 
> /home/circleci/repo/client/config/environment.js
>   7:11  warning  'getHost' was used before it was defined       no-use-before-define
>   8:20  warning  'getOAuthHost' was used before it was defined  no-use-before-define
> 
> /home/circleci/repo/client/tests/unit/adapters/application-test.js
>   9:9  warning  'adapter' is never reassigned. Use 'const' instead  prefer-const
> 
> ✖ 6 problems (2 errors, 4 warnings)
>   2 errors and 2 warnings potentially fixable with the `--fix` option.

PR #69, opened 2020-04-07
Fix landing page sign-in button, redirect to nycid

bfreeds:
This PR is a hot fix to have the big orange sign in button on the landing page to redirect users to nyc id.

It copies the logic used in the sign-in button in an index.js controller (left a todo to make this DRY, but wanted to get this hot fix in for demo).


allthesignals:
TODO: let's skillshare on how to identify test cov weaknesses and add a test

Issue #70, opened 2020-04-07
Update email address users send questions/feedback to


bfreeds:
From demo:
>We should use the ZAP_Support_DL for the contact info

bfreeds:
@hannahkates toward the end of the demo it seemed like this became less clear - can you confirm which email should be used?  Open PR for this change if it does need to be switched to `zap_suppport_dl`

hannahkates:
@bean75 Do you have a final decision on whether we should provide zap_suppport_dl or zap_feedback_dl for the email?

bean75:
Yeah - the feedback one is ok.
________________________________

hannahkates:
Okay thanks - closing this issue.

PR #71, opened 2020-04-07
70: Update user feedback email to “ZAP_Support_DL…”

bfreeds:
This PR addresses #70, fixing the email address we use to get user feedback and questions based on feedback during demo.


bfreeds:
Need confirmation that this is desired before merging this PR.

PR #74, opened 2020-04-09
Revert netlify build command syntax for HOST env

bfreeds:
Following the Netlify file config for passing branch context environment settings, the redirection when a user logs-in broke.

Looking at network logs, it appears that the HOST environment may be not getting set correctly (it’s Heroku’s develop app for qa and staging branches).

This commit would go back to the syntax we had previously, putting HOST directly in the build command for each context.

Network requests show that this app is pointing to the `develop` Heroku app -- but it should be hitting the appropriate QA heroku app.
image

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,20 +1,19 @@
 [build]
   base= "client/"
-  command = "yarn build --environment=production"
 
 # production
 [context.master]
-  environment = { HOST="https://applicant-portal.herokuapp.com/" }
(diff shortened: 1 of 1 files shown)


allthesignals:
If I understand what you're saying correctly, I don't think this has to do with the Netlify configuration, but NYC.ID. We have to configure separate NYC.ID accounts for each domain because the login redirect is defined in the service account itself.

PR #75, opened 2020-04-09
Map in the NYCID client IDs

allthesignals:
This PR tells the client which NYC.ID URI client ID to use when making the login the button.

Each environment needs a separate account because you have to configure each account with the correct redirect URL.

@bfreeds one thing I'm still not clear on is the difference between the `develop` branch and `staging` branch?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,17 +4,16 @@
 
 # production
 [context.master]
-  environment = { HOST="https://applicant-portal.herokuapp.com/" }
+  environment = { HOST="https://applicant-portal.herokuapp.com/", NYCID_CLIENT_ID='' }
 
 # staging
 [context.staging]
-  environment = { HOST="https://applicant-portal-staging.herokuapp.com" }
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -4,17 +4,24 @@
 
 # production
 [context.master]
-  environment = { HOST="https://applicant-portal.herokuapp.com/" }
+  environment = {
```
bfreeds: From the [TOML language docs](https://github.com/toml-lang/toml#inline-table):
> it is strongly discouraged to break an inline table onto multiples lines. If you find yourself gripped with this desire, it means you should be using standard tables.

If we want to split this on multiple lines, we should do it like this, which probably makes more sense as we may end up having to add more environment variables in the future:
```toml
[context.master.environment]
  HOST="..."
  NYCID_CLIENT_ID="..."
```
(oh I see you brought them on the same line.  I think the curly brackets need to be on the same line, too)

PR #76, opened 2020-04-09
Add setup and configuration for local development

allthesignals:
This commit changes some environment settings for client and server which allow us to run a local SSL server. Readme is updated.


allthesignals:
> 🎉🎉🎉
> 
> Neat solution, @allthesignals -- like that this also simplifies the dev setup. To get this working locally did you also have to add `applicant-portal-local` to NYC ID as a service account?

Woops! Yep. Let me add you to that.

`applicant-portal-local` is a new NYC.ID account.

PR #77, opened 2020-04-10
Enforce auth

allthesignals:
This PR enforces auth by throwing an exception when `contactid` is specifically not found. Before, it was only checking if `session` was visible.

This also has the browser display error messages.

### DIFF
--- a/server/src/app.controller.ts
+++ b/server/src/app.controller.ts
@@ -36,6 +36,12 @@ export class AppController {
     }
   }
 
+  @Get('/logout')
+  async logout(@Res() res: Response) {
+    res.clearCookie('token')
+      .send({ message: 'Login successful!' });
+  }
+
   @Get('/users')
   async getUser(@Session() session, @Res() res, @Query('me') me) {
     const { contactId } = session;
--- a/server/src/app.controller.ts
+++ b/server/src/app.controller.ts
@@ -39,7 +39,7 @@ export class AppController {
   @Get('/logout')
   async logout(@Res() res: Response) {
     res.clearCookie('token')
-      .send({ message: 'Login successful!' });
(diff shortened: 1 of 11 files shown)


server/src/app.controller.ts
```diff
@@ -36,6 +36,12 @@ export class AppController {
     }
   }
 
+  @Get('/logout')
+  async logout(@Res() res: Response) {
+    res.clearCookie('token')
+      .send({ message: 'Login successful!' });
```
allthesignals, line 42: This is the wrong message...

bfreeds, line 42: Note:  update this message

server/src/http-exception.filter.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Request, Response } from 'express';
+
+@Catch(HttpException)
+export class HttpExceptionFilter implements ExceptionFilter {
+  catch(exception: HttpException, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+    const status = exception.getStatus();
+
+    response
```
bfreeds, line 11: Made to fit json api spec:   https://jsonapi.org/format/#errors

server/src/http-exception.filter.ts
```diff
@@ -0,0 +1,17 @@
+import { ExceptionFilter, Catch, ArgumentsHost, HttpException } from '@nestjs/common';
+import { Request, Response } from 'express';
+
+@Catch(HttpException)
```
bfreeds, line 4: Note:  here's the docs on nest exception filters & error handling.
https://docs.nestjs.com/exception-filters#exception-filters

Issue #73, opened 2020-04-08
Update logic for placing projects in dashboard sections sections


hannahkates:
AB#10519

## User Story:
As an applicant team member, I need to see all of my active projects and see which ones I can take action on (i.e. submit a PAS).

## Display Logic:

Wireframes are provided for reference at the bottom of this GitHub issue.

**Project shows up at the top with an "Edit Pre-Application Statement" button if**
- Project
  - [x] `statecode = 'Active'` (filtered server-side. See #64)
  - [x] `dcp_visibility IN ('Applicant Only', 'General Public')` (filtered server-side See #64)
- Project applicant team
  - [x] `statuscode = 'Active'` (filtered server-side. See #64)
- PAS package 
  - [ ] `statuscode = 'Package Preparation'`
  - [x] `dcp_visibility IN ('Applicant Only', 'General Public')` (filtered server-side. See See #64 )
- [ ] Sorted alphabetically by project name

**Project appears in bottom section if**
- Project
  - [x] `statecode = 'Active'` (filtered server-side. See #64)
  - [x] `dcp_visibility IN ('Applicant Only', 'General Public')` (filtered server-side. See #64)
- Project applicant team
  - [x] `statuscode = 'Active'` (filtered server-side. See See #64)
- [ ] Sorted alphabetically by project name

Within the projects listed in the bottom section, display "View Pre-Application Statement" button if
  - [ ] PAS package `statuscode IN ('Submitted', 'Under review', 'Reviewed – No Revisions Required', 'Reviewed – Revisions Required')` all statuses except for 'Mistake'
    - Or, `statuscode NOT IN ('Package Preparation', 'Mistake')`
  - [x] PAS package `dcp_visibility IN ('Applicant Only', 'General Public')` (filtered server-side. See See #64 )

**If user isn't assigned to any projects**
Page should display a message. See wireframes.

## [Figma Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP-Wireframes?node-id=0%3A1)
![MicrosoftTeams-image (1).png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/083eca19-b3bd-4894-a7f7-e4a07d8422d3)

trbmcginnis:
Notes on what was talked about during demo: https://docs.google.com/document/d/17rNCMWqlupVkuxIrT_XCvZYS99z1H5469RzDbPMNPZQ/edit?usp=sharing

[...]

PR #84, opened 2020-04-13
73 update project list logic

godfreyyeung:
Addresses #73

### DIFF
--- a/client/app/components/projects/planning-project-card.hbs
+++ b/client/app/components/projects/planning-project-card.hbs
@@ -13,7 +13,13 @@
     {{@project.dcpApplicantCustomerValue}}
   </p>
 
-  {{#if (eq @type 'planning-project')}}
-    <a href="/pas.html" class="button hollow tiny">View Pre-Application Statement</a>
+  {{#if this.showViewPas}}
+    <a
+      href="/pas.html"
+      class="button hollow tiny"
+      data-test-view-pas
+    >
+      View Pre-Application Statement
+    </a>
   {{/if}}
 </li>
--- a/client/app/components/projects/planning-project-card.js
+++ b/client/app/components/projects/planning-project-card.js
@@ -1,0 +1,42 @@
+import Component from '@glimmer/component';
+
+const PACKAGE_STATUS_CODES = {
+  REVIEWED_NO_REVISIONS_REQUIRED: "Reviewed - No Revisions Required",
+  REVIEWED_REVISIONS_REQUIRED: "Reviewed - Revisions Required",
+  SUBMITTED: "Submitted",
+  UNDER_REVIEW: "Under Review",
+}
+
+const PACKAGE_VISIBILITY_CODES = {
+  APPLICANT_ONLY: 717170002,
+  GENERAL_PUBLIC: 717170003,
+}
+
+/**
+  * Generates a  project card listed under "Planning is working on it..."
+  * @param      {Project Model}  project
+  */
+export default class ProjectsPlanningProjectCardComponent extends Component {
+  /**
+  * @return      {bool} true if this card shows the "View Pre-Application Statement" button
+  */
+  get showViewPas() {
+    return this.args.project.pasPackages.some((projectPackage) => {
+      if (
+        [
+          PACKAGE_STATUS_CODES.SUBMITTED,
+          PACKAGE_STATUS_CODES.UNDER_REVIEW,
+          PACKAGE_STATUS_CODES.REVIEWED_NO_REVISIONS_REQUIRED,
+          PACKAGE_STATUS_CODES.REVIEWED_REVISIONS_REQUIRED,
+        ].includes(projectPackage.statuscode)
+        && [
+          PACKAGE_VISIBILITY_CODES.APPLICANT_ONLY,
+          PACKAGE_VISIBILITY_CODES.GENERAL_PUBLIC
+        ].includes(projectPackage['dcp-visibility'])
+      ) {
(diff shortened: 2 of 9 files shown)


client/app/controllers/projects.js
```diff
@@ -1,41 +1,80 @@
...
+  APPLICANT_ONLY: 717170002,
+  GENERAL_PUBLIC: 717170003,
+};
 
 export default class ProjectsController extends Controller {
   queryParams = ['email'];
 
   // TODO: organize this business logic as computed properties on the projects model
   // projects for applicant to do
   get applicantProjects () {
```
bfreeds, line 27: Update this naming to be consistent across PR.

client/app/controllers/projects.js
```diff
@@ -1,41 +1,80 @@
...
+          return false;
+        }
         return true;
-      } return false;
+      });
     });
   }
+
+
+  @sort('applicantProjects', function(projectA, projectB) {
```
bfreeds, line 65: As per PR review discussion, please add a "TODO: " note to flag a future enhancement on the sorting logic to account for instances when project titles start with a number (ex. 123 should come _before_ 8 -- which this alphabetical sorting doesn't do yet).

client/app/controllers/projects.js
```diff
@@ -61,7 +61,11 @@ export default class ProjectsController extends Controller {
     });
   }
 
-
+  // TODO: Possibly improve this sort to consider the house numbers
+  // in front of addresses. One possible implementation is that
+  // Addresses should first be sorted alphabetically by their street name,
+  // and secondly by their house number. (Currently, the sort is alphabetical
+  // even across the house number, so "123" < "8" === true)
```
bfreeds, line 68: @godfreyyeung super helpful, thank you!

client/app/components/projects/planning-project-card.hbs
```diff
@@ -13,7 +13,13 @@
     {{@project.dcpApplicantCustomerValue}}
   </p>
 
-  {{#if (eq @type 'planning-project')}}
-    <a href="/pas.html" class="button hollow tiny">View Pre-Application Statement</a>
+  {{#if this.showViewPas}}
```
allthesignals, line 16: Great!

client/app/components/projects/planning-project-card.js
```diff
@@ -0,0 +1,42 @@
...
+  * @return      {bool} true if this card shows the "View Pre-Application Statement" button
+  */
+  get showViewPas() {
+    return this.args.project.pasPackages.some((projectPackage) => {
+      if (
+        [
+          PACKAGE_STATUS_CODES.SUBMITTED,
+          PACKAGE_STATUS_CODES.UNDER_REVIEW,
+          PACKAGE_STATUS_CODES.REVIEWED_NO_REVISIONS_REQUIRED,
+          PACKAGE_STATUS_CODES.REVIEWED_REVISIONS_REQUIRED,
```
allthesignals, line 30: Beautiful

client/app/controllers/projects.js
```diff
@@ -1,41 +1,84 @@
 import Controller from '@ember/controller';
+import { sort } from '@ember/object/computed';
+
+// TODO: Get theses codes from a Constants service
```
allthesignals, line 4: My only objection to using a service would be that services are stateful objects that get initialized on app bootup, and they would be overkill for this because these constants are static. We might just keep them in a util file somewhere and import where it's needed. Thoughts?

client/app/controllers/projects.js
```diff
@@ -1,41 +1,84 @@
...
+
+const PACKAGE_VISIBILITY_CODES = {
+  APPLICANT_ONLY: 717170002,
+  GENERAL_PUBLIC: 717170003,
+};
 
 export default class ProjectsController extends Controller {
   queryParams = ['email'];
 
   // TODO: organize this business logic as computed properties on the projects model
```
allthesignals, line 25: I would emphasize this TODO here for future reference. These computeds here don't depend on any other context outside of the model itself, so could probably just live there.

client/app/components/projects/planning-project-card.js
```diff
@@ -0,0 +1,42 @@
+import Component from '@glimmer/component';
+
+const PACKAGE_STATUS_CODES = {
```
allthesignals, line 3: Can you export these, too:

`export const PACKAGE_STATUS_CODES = {`

then import from somewhere else:

`import { PACKAGE_STATUS_CODES } from '../projects/planning-project-card';`

client/mirage/factories/project.js
```diff
@@ -1,33 +1,130 @@
...
+  PACKAGE_PREPARATION: 'Package Preparation',
+  CERTIFIED: 'Certified',
+  FINAL_APPROVAL: 'Final Approval',
+  REVIEWED_NO_REVISIONS_REQUIRED: 'Reviewed - No Revisions Required',
+  REVIEWED_REVISIONS_REQUIRED: 'Reviewed - Revisions Required',
+  WITHDRAWN: 'Withdrawn',
+  SUBMITTED: 'Submitted',
+  UNDER_REVIEW: 'Under Review',
+  MISTAKE: 'Mistake',
+};
```
allthesignals, line 13: Not sure if I"m missing something, but you can export these from the original controller class file like so:

`export const PACKAGE_STATUS_CODES = {`

Then you'd be able to import them here:

`import { PACKAGE_STATUS_CODES } from '../find/correct/path/whatever-file-name-is';`

PR #85, opened 2020-04-13
79: redirect user to projects page when they visit index if already logged in

### LINKED ISSUE #79: Signed in users should not see root page
@ksinghprod noticed that the site title link takes signed in users to the root page where they see the sign in prompt, despite already being signed in. We should either redirect signed in users from `/` to `/projects` or display alternative content. IDK what that content should be, so let's go with a redirect for now.

trbmcginnis:
Redirects user to projects page when visiting index route if already logged in

Closes #79

### DIFF
--- a/client/tests/acceptance/user-can-login-test.js
+++ b/client/tests/acceptance/user-can-login-test.js
@@ -60,4 +60,25 @@ module('Acceptance | user can login', function(hooks) {
     assert.ok(find('[data-test-applicant-error-message="contact-not-assigned"'));
     assert.equal(find('[data-test-error-response="code0"]').textContent.trim(), 'code: NO_CONTACT_FOUND');
   });
+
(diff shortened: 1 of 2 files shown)


client/tests/acceptance/user-can-login-test.js
```diff
@@ -37,4 +37,25 @@ module('Acceptance | user can login', function(hooks) {
...
   });
+
+  test('If use already logged in, revisiting index reroutes them to projects page', async function (assert) {
+    // user logs in
+    this.server.create('user');
+    this.server.get('/login', (schema, request) => {
+      assert.equal(request.queryParams.accessToken, 'a-valid-jwt');
+    });
+    window.location.hash = '#access_token=a-valid-jwt';
+    await visit('/login');
```
allthesignals, line 48: Random thought: maybe we could turn this into a test helper?

PR #92, opened 2020-04-15
83 Handle login errors, like for missing crm contact

godfreyyeung:
## TL;DR;

I'm proposing that in Nest, we write exceptions in this format:

```
     const responseBody = {
        'code': 'APPLICANT_PORTAL_SPECIFIC_CODE',
        'message': 'Brief details of internal error',
      }

      throw new HttpException(responseBody, HttpStatus.UNAUTHORIZED);
```

And the frontend can then use the `error.response.code` property (`APPLICANT_PORTAL_SPECIFIC_CODE`) to conditionally display more user friendly messages.

---
This PR modifies the frontend to display error messages for errors occuring during user login.
It also adjusts how the Applicant Portal API sends error responses to support more robust error reporting on the frontend.

In the frontend, this PR introduces an Ember error page for the login route. The error page
is automatically displayed by Ember if it detects an exception or
bad response in the route. The exception or bad response is supplied
to the template as the model.

The page also implicitly logs a user out through the DoLogout
component.

In the backend, the PR modifies the AuthService.generateNewToken() response body to provide an Applicant Portal API-specific error code when a Contact cannot be found for the provided email or ID. This short error code will help the frontend more robustly display a user-friendly message when no contact is found.

By default, responses from Nest appear as such:
```
{
  "statusCode": 403,
  "message": "Forbidden"
}
```

Somewhere along the way, likely in the Nest HTTP Exception filter, the response is wrapped up inside an error object within an array. The error object takes this form 

```
{
  response: ..., 
  status: ....,
  message: .... 
}
```

Where
  - the `response` property takes the form (string or object) of whatever is passed in to the first ("response") argument of the HTTPException constructor. 
 - the `status` property is derived from the status code passed in to the second ("status")  argument of the HTTPException constructor.
  - and the `message` is either
    a. A copy of the `response` argument, if that argument is a simple string.
    b. The value of the "message" property within `response`, if `response` is an object. 
    c. Derived from the "status" argument, if `response` is an object without the `message` property. 

We need to supply an object to the `response` property to add an "Applicant Portal Error Code" within it, so we also need to add the `message` property within `response` property to set the response message (otherwise, `message` will be display a generic "HTTP Exception" string). This PR thus proposes to discard use of the `error.message` property, and instead use its identical source,  `error.response.message`. 

Previously:
image

Now: 
image

### DIFF
--- a/client/app/controllers/login-error.js
+++ b/client/app/controllers/login-error.js
@@ -1,0 +1,11 @@
+import Controller from '@ember/controller';
+
+export default class LoginErrorController extends Controller {
+  /**
+    * @return      {bool} true if login route yields error response
+    *                     about Contact not existing in CRM.
+    */
+  get contactNotAssigned() {
+    return this.model.errors.some((error) => error.response.code === 'NO_CONTACT_FOUND');
+  }
+}
--- a/client/app/templates/login-error.hbs
+++ b/client/app/templates/login-error.hbs
@@ -1,0 +1,42 @@
+<div class="cell">
+  <div class="grid-container">
+    <div class="grid-x grid-padding-x grid-padding-y align-middle">
+      <div class="cell large-6 large-offset-3 text-center">
+        <h2>Oops!</h2>
+
+        {{#if this.contactNotAssigned}}
+          <p data-test-applicant-error-message="contact-not-assigned">
+            You are not currently assigned to a DCP project. Please email&nbsp;
+            <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a>&nbsp;
+            for further assistance.
+          </p>
+        {{else}}
+          <p>
+            Something went wrong.
+          </p>
+        {{/if}}
+
+        {{#each @model.errors as |error idx|}}
+          {{#if error.detail}}
+            <code class="callout display-block text-left red-muted" data-test-error-message={{idx}}>{{error.detail}}</code>
+          {{else}}
+            {{#each-in error.response as |key value|}}
+              <code class="callout display-block text-left red-muted" data-test-error-response={{concat key idx}}>
+                {{key}}: {{value}}
+              </code>
+            {{/each-in}}
+            <code class="callout display-block text-left red-muted">
+              status: {{error.status}}
+            </code>
+          {{/if}}
+        {{/each}}
+
+        <DoLogout />
+
+      </div>
+      <div class="cell large-10 large-offset-1 xlarge-8 xlarge-offset-2  text-center">
+        <p class="">Please contact City Planning at <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a> or 212-720-3300.</p>
+      </div>
+    </div>
+  </div>
+</div>
--- a/client/tests/acceptance/user-can-login-test.js
+++ b/client/tests/acceptance/user-can-login-test.js
@@ -1,9 +1,12 @@
 import { module, test } from 'qunit';
-import { visit, click, currentURL } from '@ember/test-helpers';
+import {
+  visit, click, currentURL, find,
+} from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import window, { setupWindowMock } from 'ember-window-mock';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 import { currentSession } from 'ember-simple-auth/test-support';
+import { Response } from 'ember-cli-mirage';
 
 module('Acceptance | user can login', function(hooks) {
   setupApplicationTest(hooks);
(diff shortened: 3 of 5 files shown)


godfreyyeung:
Still need to update it to use the copy specified in the Azure portal. Standby.

client/app/controllers/login-error.js
```diff
@@ -0,0 +1,11 @@
+import Controller from '@ember/controller';
+
+export default class LoginErrorController extends Controller {
+  /**
+    * @return      {bool} true if login route yields error response
+    *                     about Contact not existing in CRM.
+    */
+  get contactNotAssigned() {
+    return this.model.errors.some((error) => error.response === 'CRM user not found. Please make sure your e-mail or ID is associated with an assignment.');
```
allthesignals, line 9: Hmmm.. I think this could get unwieldy over time. What if the language of the original response error changes? Maybe the backend needs to provide error codes (CONTACT_NOT_ASSIGNED) that could be used for this sort of logic.

Also, AFAICT, I have never seen this emit more than one error before... I'm not sure which circumstances allow that. The JSON:API spec requires an error document to allow multiple though.

godfreyyeung: Oh yeah, I was wary of the same. I think error codes would be fantastic! 
We've been tasked to provide more user friendly error messages, and unfortunately this is the only way I've found detect if it is specifically a CRM missing contact error.

Should we make a request of this to the CRM team? Hoping they'll have ability to modify CRM error codes.

If they can, they could even just change the CRM `error.response` value to " You are not currently assigned to a DCP project.....", so that we don't have to write special handling in our frontend code.

allthesignals: No, I mean that our API would provide our own error codes — for example, where the text for this error is emitted, it might return a separate property as part of the response:
```
{
  message: "error message"
  code: "CONTACT_NOT_ASSIGNED"
}
```

These are not errors that come from CRM anyway. They come from our own app. CRM's error just says "Record not found" or something.

godfreyyeung: Gotcha -- I shoulda caught this earlier! I have a good idea of how to go about this now.

client/app/controllers/login-error.js
```diff
@@ -0,0 +1,11 @@
+import Controller from '@ember/controller';
+
+export default class LoginErrorController extends Controller {
+  /**
+    * @return      {bool} true if login route yields error response
+    *                     about Contact not existing in CRM.
+    */
+  get contactNotAssigned() {
+    return this.model.errors.some((error) => error.response.code === 'NO_CONTACT_FOUND');
```
bfreeds, line 9: Is `NO_CONTACT_FOUND` the error code CRM sends to our API, or are we defining "NO_CONTACT_FOUND" as the error code that our API creates?

godfreyyeung: The latter -- an error code that our API creates!

bfreeds: Nevermind!
> And the frontend can then use the error.response.code property (APPLICANT_PORTAL_SPECIFIC_CODE) to conditionally display more user friendly messages.

client/app/templates/login-error.hbs
```diff
@@ -0,0 +1,42 @@
...
+            {{#each-in error.response as |key value|}}
+              <code class="callout display-block text-left red-muted" data-test-error-response={{concat key idx}}>
+                {{key}}: {{value}}
+              </code>
+            {{/each-in}}
+            <code class="callout display-block text-left red-muted">
+              status: {{error.status}}
+            </code>
+          {{/if}}
+        {{/each}}
```
allthesignals, line 32: What do you think about this:

What if we put all of it into the loop here? For example, you could say: 

```handlebars
  {{#if (eq error.code 'NO_CONTACT_FOUND')}}
    {{!-- markup here --!}}
  {{else}}
    {{!-- generic error templates here --!}}
  {{/if}}
```

thoughts? it just seems like that computed is a little too much code for what it's actaully doing.

client/tests/acceptance/user-can-login-test.js
```diff
@@ -37,4 +40,24 @@ module('Acceptance | user can login', function(hooks) {
...
   });
+
+  test('User sees error message if no CRM contact is found for their email', async function (assert) {
+    const accessToken = 'a-valid-jwt';
+    this.server.create('user');
+    this.server.get('/login', () => new Response(401, { some: 'header' }, {
+      errors: [{
+        response: {
+          code: 'NO_CONTACT_FOUND',
+          message: 'This message is nice but does not affect frontend logic',
```
bfreeds, line 51: This message _is_ nice, if useless 🤣

client/app/templates/login-error.hbs
godfreyyeung: Would this go on lines 7-17, or 23-30 ?

If in 23-30, is the idea to only display the user-friendly message, instead of both the user-friendly message and the API message?

client/tests/acceptance/user-can-login-test.js
godfreyyeung: Loooove that y'all caught this. Makes me happy.

godfreyyeung:
> Really nice PR, @godfreyyeung -- having clean error handling is so helpful.

Thank you for reviewing! 🙇 🙏

client/app/templates/login-error.hbs
allthesignals: Nope, I'm saying remove `this.contactNotAssigned` completely, including the computed, but use the markup there and put the code i have written in 23-30. 

The idea is to put the relevant markup inside the loop that's already there to keep that markup in a place where a loop is already happening.

Issue #87, opened 2020-04-13
When a user is not assigned to any project, the displayed message is not as per the requirement 


ksinghprod:
Browser: Chrome 

Expected Result: when the user logs in with valid portal account and has a corresponding contact record which is active AND not associated to any active ZAP project with applicant or public visibility, the portal must display the following text:

“You are not currently assigned to a DCP project.”

Actual Result: the displayed text is "you do not currently have any projects that require your response" 

![image](https://user-images.githubusercontent.com/57362229/79158881-17c3b880-7da5-11ea-93eb-60bc383c132d.png)

![image](https://user-images.githubusercontent.com/57362229/79158931-2f9b3c80-7da5-11ea-8ffa-365e9270ff5d.png)

hannahkates:
Confirmed, this is the screen a user will see if they don't have any projects assigned to them in CRM. @bean75 Do you have a strong preference on language for this? We can update it to say "You are not currently assigned to a DCP project." instead.

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/5de452b0-7500-49bb-900e-89f18c3ccd8b)

bean75:
I think the Nothing to do is fine - but update the line below to indicate that:
You are not currently assigned to any active projects. Please reach out to ZAP_Feedback_DL@planning.nyc.com if you believe this is an error.
________________________________

hannahkates:
@godfreyyeung The text for this is off. The "You are not currently assigned to a DCP project." text should only show up when the user isn't assigned to ANY projects, not just zero projects in the "to do" portion of the page. See screenshot example below. In this example, the user should see "You do not currently have any projects that require your response"

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/b9464cb6-2c15-4c6b-a735-a31efa65c3b2)

godfreyyeung:
No To-Do projects:
```
Nothing to do.
You do not currently have any projects that require your response
```

No Active Applicant-Only or General Public projects whatsoever (Needs clarification):
```
Nothing to do.
[You are not currently assigned to a DCP project / You are not currently assigned to any active projects.] 

Please reach out to ZAP_Feedback_DL@planning.nyc.com if you believe this is an error.
```

bean75:
Nothing to do.
You are not currently assigned to any active DCP projects.

[...]

PR #93, opened 2020-04-16
87 Update text displayed when user has no assigned projects

godfreyyeung:
Updates the text as per Azure ticket mentioned in #87

### DIFF
--- a/client/tests/acceptance/user-sees-projects-of-all-types-test.js
+++ b/client/tests/acceptance/user-sees-projects-of-all-types-test.js
@@ -1,5 +1,5 @@
-import { module, test, skip } from 'qunit';
-import { visit, findAll } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+import { visit, find, findAll } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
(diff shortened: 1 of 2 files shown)


client/tests/acceptance/user-sees-projects-of-all-types-test.js
```diff
@@ -27,7 +27,10 @@ module('Acceptance | user sees projects of all types', function(hooks) {
     assert.equal(findAll('[data-test-view-pas]').length, 5);
   });
 
-  skip('Page should display non-assigned message if no projects');
+  test('Page should display non-assigned message if no projects', async function(assert) {
+    await visit('/projects');
+    assert.ok(find('[data-test-not-assigned]'));
+  });
```
allthesignals, line 33: At first, I wondered where the data setup was, but then I realized it implicitly provides no projects because none get created!

If I realllly wanted to get pick, though, I might've tried this:

```
this.server.createList('project', 0);
```

I'm sure mirage doesn't allow for that, though, haha.

godfreyyeung: ohhh yes. I'm always a fan of being explicit. hmmmmm interesting idea

Issue #88, opened 2020-04-13
Applicant is not able to see the project status of their project 


ksinghprod:
Browser: Chrome 
Project ID: 2018M0365

Expected Result: Applicant should be able to see the following the details about their project on the landing pageL

Project Name (default sort)
Project ID
Project Status (Active or On-Hold)
Primary Applicant Name
Packages

Actual Result: User is unable to see the project status of their project 

![image](https://user-images.githubusercontent.com/57362229/79162001-763f6580-7daa-11ea-9b9c-7517e65840c9.png)

![image](https://user-images.githubusercontent.com/57362229/79161757-07faa300-7daa-11ea-883b-ac33d6990ee0.png)

hannahkates:
@bean75 Thoughts on whether it's still beneficial to whether a project is "on hold"? This distinction hasn't come up in our recent convos about the wireframes. Since all projects in CRM are currently "on hold" we may want to hold off on implementing that since it could alarm users submitting PAS packages.

bean75:
I think we should still proceed with it - and actually it is best that they see that everything is on pause.
________________________________

ksinghprod:
This issue is now fixed 

![image](https://user-images.githubusercontent.com/57362229/79794649-9e861180-8320-11ea-88c8-dcd2d910f056.png)

bfreeds:
Thank you @bean75 & @ksinghprod!

PR #96, opened 2020-04-16
88:  Show if project is "On-Hold"

bfreeds:
This PR addresses #88 by showing an "On Hold" message on a project card if the `statuscode` = "On-Hold".

Ex.
image

### DIFF
--- a/client/app/components/projects/applicant-project-card.hbs
+++ b/client/app/components/projects/applicant-project-card.hbs
@@ -5,6 +5,15 @@
 >
   <h3 class="no-margin">
     {{@project.dcpProjectname}}
+
+    {{#if (eq @project.statuscode "On-Hold")}}
+      <span class="text-red text-weight-normal"
+            data-test-project-on-hold
+      > 
+        (On Hold) 
+      </span>
+    {{/if}}
+
     <small class="text-weight-normal">{{@project.id}}</small>
   </h3>
 
--- a/client/app/components/projects/planning-project-card.hbs
+++ b/client/app/components/projects/planning-project-card.hbs
@@ -5,6 +5,15 @@
 >
   <h3 class="no-margin">
     {{@project.dcpProjectname}}
+
+    {{#if (eq @project.statuscode "On-Hold")}}
+      <span class="text-red text-weight-normal"
+            data-test-project-on-hold
+      > 
+        (On Hold) 
+      </span>
+    {{/if}}
+
     <small class="text-weight-normal">{{@project.id}}</small>
   </h3>
 
--- a/client/mirage/factories/project.js
+++ b/client/mirage/factories/project.js
@@ -21,6 +21,8 @@ const PACKAGE_VISIBILITY_CODES = {
 };
 
 export default Factory.extend({
+  statuscode: 'Active', // default project statuscode
+
   applicant: trait({
     dcpDcpProjectDcpPackageProject(i) {
       const statuses = [
@@ -132,5 +134,9 @@ export default Factory.extend({
 
   noPackages: trait({
     dcpDcpProjectDcpPackageProject: [],
+  }),
+
+  onHold: trait({
+    statuscode: 'On-Hold',
   }),
 });
--- a/server/src/projects/projects.controller.ts
+++ b/server/src/projects/projects.controller.ts
@@ -1,6 +1,6 @@
-import { 
-  Controller, 
-  Get, 
+import {
+  Controller,
+  Get,
   Query,
   Res,
   HttpException,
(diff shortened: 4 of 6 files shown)


client/app/components/projects/applicant-project-card.hbs
```diff
@@ -5,6 +5,15 @@
...
   <h3 class="no-margin">
     {{@project.dcpProjectname}}
+
+    {{#if (eq @project.statuscode "On-Hold")}}
+      <span class="text-red text-weight-normal"
+            data-test-project-on-hold
+      > 
+        (On Hold) 
+      </span>
+    {{/if}}
```
allthesignals, line 15: This is great.

Could even have some new computeds on the project model:

`{{#if (eq @project.isOnHold)}}`

Love it.

client/app/components/projects/planning-project-card.hbs
```diff
@@ -5,6 +5,15 @@
...
     {{@project.dcpProjectname}}
+
+    {{#if (eq @project.statuscode "On-Hold")}}
+      <span class="text-red text-weight-normal"
+            data-test-project-on-hold
+      > 
+        (On Hold) 
+      </span>
+    {{/if}}
+
```
allthesignals, line 16: What is the difference between applicant-project-card and planning-project-card components? I'm confused what these are doing and why're they're separate things. They share a lot of template logic.

client/mirage/factories/project.js
```diff
@@ -133,4 +135,8 @@ export default Factory.extend({
   noPackages: trait({
     dcpDcpProjectDcpPackageProject: [],
   }),
+
+  onHold: trait({
+    statuscode: 'On-Hold',
+  }),
```
allthesignals, line 141: Fabulous.

server/src/projects/projects.controller.ts
```diff
@@ -58,7 +58,7 @@ export class ProjectsController {
   // Serializes an array of objects into a JSON:API document
   serialize(records, opts?: object): Serializer {
     const ProjectsSerializer = new Serializer('projects', {
-      attributes: ['dcp_projectname', 'dcp_name', 'statecode', 'dcp_visibility', '_dcp_applicant_customer_value', 'dcp_dcp_project_dcp_projectapplicant_Project', 'dcp_dcp_project_dcp_package_project'],
+      attributes: ['dcp_projectname', 'dcp_name', 'statecode', 'statuscode', 'dcp_visibility', '_dcp_applicant_customer_value', 'dcp_dcp_project_dcp_projectapplicant_Project', 'dcp_dcp_project_dcp_package_project'],
```
allthesignals, line 61: No idea what to do when/if this list gets "too" long (in terms of just visually how it appears in the text editor). I don't think it's too long now (i don't even think our linter complains about that stuff).

client/app/components/projects/applicant-project-card.hbs
bfreeds: @allthesignals I originally created the project card components, when the markup between the two was more different.  They have since become basically the same, and totally on the same page -- one of the small refactors I was hoping to get to before the end of this sprint is to create one project card component that's reusable across the lists.

bfreeds: Also interested to chat about the idea of moving that to a computed -- I'm not sure I understand Ember data models yet to know what benefits that'd get us!

client/app/components/projects/planning-project-card.hbs
bfreeds: @allthesignals yea they're basically the same now & was going to refactor, just wanted to get this into QA and then take a step back and make that consolidation.

server/src/projects/projects.controller.ts
bfreeds: @allthesignals I had originally autoformatted this with Prettier and it picked this up and made it more readable -- I pulled that back out to make the git diff easier to read in this PR, but take a look at this commit to see how Prettier would make our code more readable, and take away some of these judgement calls about spacing/formatting with some very (imo) sensible default about what makes code readable.
https://github.com/NYCPlanning/labs-applicant-portal/pull/96/commits/16ba20f0e3d4d758c1e2a24f398c1bc19add1dd6

client/app/components/projects/applicant-project-card.hbs
allthesignals: Hmm yeah we can talk. Computed properties can be made on anything, including ember data models, because models themselves extend the object model.

It would get us closer to the convention of keeping business logic out of templates, although it's so minor here it doesn't matter. I have, however, seen it get pretty out of hand, though — lots of nested conditionals in templates.

Issue #18, opened 2020-03-12
Create editable PAS form page


hannahkates:
[PAS package CRM fields with character limits and option sets](https://app.zenhub.com/files/243359669/ba553d91-203b-42ca-868b-5818301b41c3/download)
[Draft Data Dictionary](https://app.zenhub.com/files/243359669/349cdea9-7714-4603-8c17-dea0d091206e/download)

## User Story
- Applicant users need to provide details about their proposed project in the PAS package, to help DCP understand their proposal and advance in the application process.
- DCP staff who receive the PAS packages use the data collected to determine staff assignments on the project application.

## Front End
This issue captures the full template and data model needed for the PAS form inputs. We should start by building out a template with all the simplest PAS form inputs (ones that are written directly to the PAS form entity), leaving placeholders for the more complicated features. That way we can add the more complex features individually as subsequent issues.

### [Figma Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP-Wireframes?node-id=0%3A1)
^ Rely on wireframes for all text, field names, and option sets.

### Checklist of form sections for this issue:
- [ ] Project Name
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/15e95410-eec3-46a4-9997-f3497f7a6ce1)
- [ ] Project Area ![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/5896e97c-cd17-4264-9e65-bc34d2ad9e7e)
- [ ] Proposed Development Site
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/69999040-e04f-409c-8339-1b75849b281b)
- [ ] Project Description
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/009e3642-05db-46b2-b102-997c398bd244)

**Complex form features to ignore in this issue and just put in placeholders for:**
- Applicant team and ability to add new team members #20
- Project area geography #21
- Land use actions and ability to add new actions #22
- File upload #23
- Functional progress checklist #24

## Implementation Notes
@allthesignals: This issue implies setting up routing for the form views. This also means setting up the model.

We might consider using ember-m3 to better support mutation within nested structures.

PR #100, opened 2020-04-22
Add PASForm/Package routes and models in the frontend

allthesignals:
Addresses #18 by setting up the routes and initial models for an editable pas form page.

Includes changes to the way package types are handled in mirage: instead of defining traits on the project factory, those are now handled through the package factory.

Lastly, makes it easier to run `ember s` from inside the root of the project and fake-login.


client/app/router.js
```diff
@@ -12,4 +12,9 @@ Router.map(function() { // eslint-disable-line
   this.route('projects');
   this.route('login');
   this.route('logout');
+
+  this.route('packages', function() {
```
godfreyyeung, line 16: Make sense

client/app/models/package.js
```diff
@@ -0,0 +1,18 @@
+import Model, { attr, belongsTo } from '@ember-data/model';
+
+export default class PackageModel extends Model {
+  @belongsTo('project')
+  project;
+
+  @belongsTo('pas-form')
```
bfreeds, line 7: Should this relationship be that `package belongsTo pas-form`, or `package hasOne pas-form`?

client/mirage/factories/project.js
```diff
@@ -49,121 +29,35 @@ export default Factory.extend({
...
-      ];
+    afterCreate(project, server) {
+      server.create('package', { project }, 'planningNoViewPASButton');
     },
   }),
 
-  noPackages: trait({
-    dcpDcpProjectDcpPackageProject: [],
-  }),
+  // are these used? are they even correct?
```
godfreyyeung, line 59: yeah, I think we can delete this actually.

client/config/environment.js
```diff
@@ -72,9 +72,15 @@ function getHost(environment) {
 }
 
 function getOAuthHost(environment) {
-  const { NYCID_CLIENT_ID } = process.env;
+  const { NYCID_CLIENT_ID, HOST } = process.env;
 
-  if (environment === 'development') {
+  // plain local development, no local backing server
+  if (environment === 'development' && !HOST) {
+    return 'http://localhost:4200/login#access_token=test';
```
godfreyyeung, line 79: note to self: This works bc it hits mirage

client/mirage/factories/project.js
```diff
@@ -35,6 +60,7 @@ export default Factory.extend({
 
       return [
         {
+          id: `applicant-${i}`,
```
godfreyyeung: @allthesignals  Is this the actual id format that the API will assign package models?

client/app/models/project.js
```diff
@@ -7,11 +7,12 @@ export default class ProjectModel extends Model {
 
   @attr dcpApplicantCustomerValue;
 
-  @attr({ defaultValue: () => [] }) dcpDcpProjectDcpPackageProject;
+  @hasMany('package', { async: false })
+  packages;
```
godfreyyeung, line 11: I think we also need to change this on the backend, in the serializer in `project.controller.ts`.

bfreeds:
> I would prefer if we also introduced the backend changes for introducing the package model here, as well. After `yarn run start`, the project list filtering is broken -- projects are not receiving any related packages.

I agree with this ^, and would change this PR's subject to be "18 add pasForm model and routes" or something to indicate that this is addressing a part of the original issue - since #18 encompasses the UI for an editable form, too.

trbmcginnis:
> > I would prefer if we also introduced the backend changes for introducing the package model here, as well. After `yarn run start`, the project list filtering is broken -- projects are not receiving any related packages.
> 
> I agree with this ^, and would change this PR's subject to be "18 add pasForm model and routes" or something to indicate that this is addressing a part of the original issue - since #18 encompasses the UI for an editable form, too.

Aren't we only focusing on frontend isolated tests now? Running `ember s` and looking at the test suite rather than running client & server together? If we want to broaden our approach and include a full-on comprehensive backend + frontend change, then maybe we should standardize that across all the other issues we're working on and not just build frontend components but also include the new required endpoints/API changes -- up to ya'll in terms of how you want to go about it, just want to make sure we're all on the same page

client/app/router.js
godfreyyeung, line 16: Actually, should this be `package` singular, since this route is editing one specific package?

client/app/models/package.js
allthesignals: Thanks! You're right, logically it would be a hasOne. But you could also say the same for a pas-form: it has one package. There's a 1-to-1 relationship between the two.

The `hasMany` and `belongsTo` verbiage is borrowed from Rails' ActiveRecord Association API, which was designed for abstracting how relational databases work. In the case of relational databases, this makes sense: there's a real difference between belongsTo and hasOne because one expects a foreign key, the other infers it. 

However, in Ember Data, these APIs were not designed for relational databases, or any specific database in general. It's just linking together the in-memory relationships between JSON objects, which may be by their nature nested and bi-directional.

All that to say, there is no `hasOne` in Ember Data — in this case, `belongsTo` does the same thing, and a 1-to-1 relationship should use `belongsTo` on both sides.

allthesignals:
> This looks great, @allthesignals. I think the `server/yarn.lock` is accidentally committed here. Also want to confirm the pasForm relationship before merging.

@bfreeds hmm, this I do think this should be committed because it signifies that perhaps a minor or patch version drift occurred in the dependency tree. Thoughts?

FWIW, I do tend to use the root-level `yarn run start` a lot, but I also use the client/server CLIs individually as well.

client/mirage/factories/project.js
allthesignals: Nope, this is from an old commit. I made changes along the way to make tests work. Think of it as an incremental step.

client/app/router.js
allthesignals: I think this should be plural. A singular package is actually handed in the subroute, packages.edit. Keeping this as "packages" follows the convention in case there is an index subroute here, and multiple packages are needed.

allthesignals:
> I would prefer if we also introduced the backend changes for introducing the package model here, as well. After `yarn run start`, the project list filtering is broken -- projects are not receiving any related packages.
> 
> I'd understand if it's broken up into another PR, too.

@godfreyyeung  Hmmm, okay. Can you elaborate more? I might want to put this in another PR because I'm already running up against the 500 diff count (minus yarn.lock) limit for our pull requests.

From what I can tell, the app isn't actually quite "broken" — there aren't really errors. It just doesn't receive any packages.

I'll try to keep this in mind from now on, although, by requiring these respective backend changes, I do think this reduces the advantage of having Ember Mirage installed at all.

On the other hand, it does put the `develop` branch in a "broken" state, as far as looking at its full integration.

If I can get another PR up for the backend today, could I get an approval?

allthesignals:
> > I would prefer if we also introduced the backend changes for introducing the package model here, as well. After `yarn run start`, the project list filtering is broken -- projects are not receiving any related packages.
> 
> I agree with this ^, and would change this PR's subject to be "18 add pasForm model and routes" or something to indicate that this is addressing a part of the original issue - since #18 encompasses the UI for an editable form, too.

You're right, good point! Done.

client/app/models/project.js
allthesignals: Addressed this comment in the main PR thread, see response there :)

client/app/router.js
godfreyyeung: sounds good 👍

godfreyyeung:
> > I would prefer if we also introduced the backend changes for introducing the package model here, as well. After `yarn run start`, the project list filtering is broken -- projects are not receiving any related packages.
> > I'd understand if it's broken up into another PR, too.
> 
> @godfreyyeung Hmmm, okay. Can you elaborate more? I might want to put this in another PR because I'm already running up against the 500 diff count (minus yarn.lock) limit for our pull requests.
> 
> From what I can tell, the app isn't actually quite "broken" — there aren't really errors. It just doesn't receive any packages.
> 
> I'll try to keep this in mind from now on, although, by requiring these respective backend changes, I do think this reduces the advantage of having Ember Mirage installed at all.
> 
> On the other hand, it does put the `develop` branch in a "broken" state, as far as looking at its full integration.
> 
> If I can get another PR up for the backend today, could I get an approval?

sounds good to me!

client/app/models/package.js
bfreeds: Awesome, thanks for clarifying and explaining that @allthesignals, totally makes sense :-)

bfreeds:
> @bfreeds hmm, this I do think this should be committed because it signifies that perhaps a minor or patch version drift occurred in the dependency tree. Thoughts?

That makes sense to me -- was just flagging because I thought maybe you had other work in the server directory that wasn't part of this PR (since I wasn't seeing any changes to `server/package.json`).

PR #102, opened 2020-04-24
Refactor CRM calls using more expressive ORM

allthesignals:
This PR refactors the code for handling projects by centralizing all the project-specific responsibilities into the projects service instead (before, some of it was in the controller).

Additionally, this PR makes it simpler to add spacing to our OData queries so they're easier to read.

If you look at the commit history, you'll notice I went from this:
image

To this:
image

I felt like the first option was adding more abstraction that was actually just making things more confusing. The second approach seems to add clarity without introducing too much additional stuff.


server/src/crm/crm.service.ts
```diff
@@ -32,158 +31,158 @@ export class CrmService {
...
+  // most likely means it's a date. we want the date formatting that
+  // json stringify provides.
+  if (typeof value === 'object') {
+    const stringyDate = JSON.stringify(value).replace(/"/g, "'");
 
-    return newObj;
+    typeSafeValue = `${stringyDate}`;
   }
 
-  async get(query, maxPageSize = 100, headers= {}) {
```
godfreyyeung, line 108: I was hoping to migrate the _utils/crm-web-api.ts file into this crm.service.ts file so that those functions can be part of the nest framework, and we would have less ad-hoc looking util files. Any reason to reintroduce  `_utils/crm-web-api.ts` and move functions like `get()` out of this service?

godfreyyeung: Oh I see,  get() was moved higher up in this file and it's now wrapping the crm-web-api.ts get() .

allthesignals: My original reasoning was that since we didn't write any of that code, we should isolate it somewhere and not touch it. But I think I see your point. I'll change it back and I think it'll make for a nicer looking diff.

allthesignals: Also wil encourage us to make the code better!

allthesignals: @godfreyyeung yep but I did take your advice and move these methos back.

I think if we're to touch these methods more (we will), we _have_ to refactor them. They're very weird. Thoughts?

server/src/crm/crm.service.ts
```diff
@@ -74,38 +95,7 @@ export class CrmService {
...
-      else {
-        newObj[name] = dataObj[name];
-      }
-    }
-
-    return newObj;
-  }
-
-  async get(query, maxPageSize = 100, headers= {}) {
+  async _get(query, maxPageSize = 100, headers = {}): Promise<any> {
```
godfreyyeung, line 98: I like this!

server/src/crm/crm.service.ts
godfreyyeung: I see your point about intentionally isolating strange code. I agree, if we modify it more we should do some heavy duty refactoring.

server/src/projects/projects.service.ts
```diff
@@ -3,33 +3,86 @@ import {
...
-    const applicantActiveStatusCode = '1';
-    const projectActiveStateCode = '0';
-    const projectVisibilityApplicantOnly = '717170002';
-    const projectVisibilityGeneralPublic = '717170003';
-
-    // Todo: Migrate this query to use oData service and write tests for this 
-    try  {
-      results = await this.crmService.get(`dcp_projects?$filter=dcp_dcp_project_dcp_projectapplicant_Project/any(o:o/_dcp_applicant_customer_value%20eq%20%27${contactId}%27%20and%20o/statuscode%20eq%20${applicantActiveStatusCode})%20and%20(dcp_visibility%20eq%20${projectVisibilityApplicantOnly}%20or%20dcp_visibility%20eq%20${projectVisibilityGeneralPublic})%20and%20statecode%20eq%20${projectActiveStateCode}&$expand=dcp_dcp_project_dcp_package_project,dcp_dcp_project_dcp_projectapplicant_Project($filter=%20statuscode%20eq%20${applicantActiveStatusCode})`);
+    try {
+      const { records } = await this.crmService.get('dcp_projects', `
```
godfreyyeung, line 24: ❤️

server/src/contact/contact.service.ts
```diff
@@ -47,20 +48,18 @@ export class ContactService {
...
+          and statuscode eq ${ACTIVE_CODE}
+        &$top=1
+      `);
+
+      return firstRecord;
     } catch(e) {
       const errorMessage = `Error finding contact by email. ${e.message}`;
       console.log(errorMessage);
       throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
     }
```
godfreyyeung, line 63: We should maybe also move the Contact Serializer out of app.controller.ts and into contact.service.ts, for consistency with project.service.ts

godfreyyeung:
> This PR refactors the code for handling projects by centralizing all the project-specific responsibilities into the projects service instead (before, some of it was in the controller).
> 
> Additionally, this PR makes it simpler to add spacing to our OData queries so they're easier to read.
> 
> If you look at the commit history, you'll notice I went from this:
> image
> 
> To this:
> image
> 
> I felt like the first option was adding more abstraction that was actually just making things more confusing. The second approach seems to add clarity without introducing too much additional stuff.

This is so awesome!! 🔥

server/src/contact/contact.service.ts
allthesignals: Yes!! Great catch thanks @godfreyyeung . Will do in follow up PR.

server/src/crm/crm.service.ts
```diff
@@ -49,7 +70,7 @@ export class CrmService {
     return "Error";
   }
 
-  private dateReviver (key, value) {
```
bfreeds, line 52: Not a blocker but I'd suggest we keep these `private` Typescript keywords for safety, especially since this is code that we didn't write ourselves.

PR #103, opened 2020-04-24
23 file upload component core frontend functionality

godfreyyeung:
Thinking of opening this up for official PR review and Merge at this state, and tackling the below Todos in future PRs. This should help avoid a massive PR down the road. 

At this stage, the component has a stable frontend UI and tests. It can
  - [x] Select files
  - [x] Deselect a file for upload
  - [x] Mark files for deletion
  - [x] Unmark files for deletion
  - [x] Execute POST upload requests for all selected files
  - [x] Execute DELETE requests for all to-be-deleted files
  - [x] save() method to initiate processing all to-be-uploaded and to-be-deleted files.
--- 

Todo/Redo for future issues:
- [x] convert tests to use this.server.create and store.query
- [x] use qunit-dom
- [ ] Rework File Manager to sit on Package models, instead of being managed inside File Managers service
  - [x] Make sure File Manager creation happens on model init, instead of in a getter.

Core feature remaining todos:

- [ ] Set up upload and delete backend endpoints
- [ ] isDirty computed on File Manager service

### DIFF
--- a/client/app/components/packages/attachments.hbs
+++ b/client/app/components/packages/attachments.hbs
@@ -127,6 +127,8 @@
             </div>
             <div class="cell small-1 text-right">
               <button
+                 {{on "click" (fn this.deselectFileForUpload file)}}
+                data-test-deselect-file-button={{idx}}
                 type="button"
               >
                 <FaIcon @icon="times" @prefix="fas" class="text-red" />
--- a/client/app/components/packages/attachments.hbs
+++ b/client/app/components/packages/attachments.hbs
@@ -77,7 +77,7 @@
       </ul>
       <hr>
       <ul class="no-bullet">
-        {{#each this.fileManager.deleteFiles as |file idx|}}
+        {{#each this.fileManager.filesToDelete as |file idx|}}
           <li
             class="text-justify"
           >
@@ -108,7 +108,7 @@
         {{/each}}
       </ul>
       <ul class="no-bullet">
-        {{#each this.fileManager.uploadFiles.files as |file idx|}}
+        {{#each this.fileManager.filesToUpload.files as |file idx|}}
           <li
             class="text-justify"
           >
--- a/client/app/components/packages/attachments.hbs
+++ b/client/app/components/packages/attachments.hbs
@@ -1,6 +1,12 @@
 <h3>
   Attached Documents
-  <a class="section-anchor" id="uploads" data-magellan-target="uploads"></a>
+  <a
+    href="#"
+    class="section-anchor"
+    id="uploads"
+    data-magellan-target="uploads"
+  >
+  </a>
 </h3>
 
 <p>
@@ -9,15 +15,26 @@
 
 <div class="site-main grid-x grid-margin-x">
   <div class="cell small-6">
-    REQUIRED FOR ALL PROJECTS:
+    <b>REQUIRED FOR ALL PROJECTS:</b>
     <ul>
-      <li>Official Zoning Sectional Map</li>
-      <li>Project Area Photographs</li>
-      <li>Tax Map(s)</li>
+      <li><a href="#">Tax Map(s)</a> — if project area has more than one tax block submit a separate map for each tax block</li>
+      <li><a href="#">Official Zoning Map</a> — circle the project area</li>
+      <li><a href="#">Land Use Map</a></li>
+      <li>Site Plan — include building heights, unit counts and square footage calculations (GSF &nbsp; ZSF) for proposed project (certain exceptions apply, consult the Lead Planner)</li>
+      <li>Photographs</li>
+      <li>Signature Form</li>
     </ul>
   </div>
   <div class="cell small-6">
-    FOR PROJECTS IN THE NYC COASTAL ZONE:
+    <b>FOR PROJECTS IN THE NYC COASTAL ZONE:</b>
+    <ul>
+      <li>Map of <a href="#">project location within the NYC coastal zone</a> including Special Area Designations</li>
+    </ul>
+    <b>FOR <a href="#">CITY MAP CHANGES</a>:</b>
+    <ul>
+      <li>Most recent Borough President’s final Section Map or most recent approved and filed Alteration Map for the project area</li>
+      <li>Site survey — recommended but not required</li>
+    </ul>
   </div>
   <div class="cell">
     <fieldset class="fieldset">
@@ -34,6 +51,7 @@
           <div class="grid-x">
             <div class="cell small-8">
               <a
+                href="#"
                 data-test-document-name={{idx}}
               >
                 {{file.name}}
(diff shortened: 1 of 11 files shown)


client/app/components/packages/attachments.js
```diff
@@ -0,0 +1,21 @@
...
+  @service
+  fileManagement;
+
+  get fileManager() {
+    return this.fileManagement.fileManagers[this.args.package.id]
+    || this.fileManagement.createFileManager(
+      this.args.package.id,
+      this.args.package.documents,
+    );
+  }
```
allthesignals, line 20: Hmm, so I recall running into some trouble with this approach in LUP portal. Do you recall?

In my experience, it can get tricky when you _create_ things inside of a _getter_ (or a computed property). The reason for that is it's a side-effect of calling the property (`myComponent.fileManager`): your native getter there, which should just _read_ some data from somewhere, actually _may_ conditionally create something.

Alternatively, could we create/lookup a file manager somewhere else and pass it into this component?

I'm guessing that the reasoning here is that we could return to the component, and we want a way to retain the same state. But I wonder if the way to handle that is to figure out what happens when the user navigates away from the form, and un-uploaded files are still in-memory. What if the component clears the file manager when it's destroyed?

client/app/services/file-management.js
```diff
@@ -0,0 +1,25 @@
...
+  ) {
+    this.fileManagers[id] = {
+      existingFiles,
+      deleteFiles: [],
+      uploadFiles: this.fileQueue.create(id),
+    };
+
+    return this.fileManagers[id];
+  }
+}
```
allthesignals, line 25: I really like that you wrapped this 3rd party file service inside a service that our app actually owns. This helps with testing a lot. I think we should make a habit of doing this :) great job

client/app/services/file-management.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default class FileManagementService extends Service {
+  @service
+  fileQueue;
+
+  // A hash of all File Managers in the app.
+  // There is one File Manager per editable Package.
+  // The key to access a File Manager is the respective
+  // package ID.
+  fileManagers = {}
```
allthesignals, line 11: I see, so if the user has multiple forms, there will need to be multiple file managers across those forms. Makes sense!

I worry a little about memory leakage here but with ES6 classes it should be good.

allthesignals: Alternatively, what do you think about this: what if we created a file manager on the PASForm model itself? It would be an ancillary part of the model class, manage its own state, and maintain a 1-to-1 mapping with the various forms... that way we don't have to manage a dictionary of references to existing file managers.

client/app/components/packages/attachments.hbs
```diff
@@ -0,0 +1,120 @@
...
+              </small>
+            </div>
+            <div class="cell small-1 text-right">
+              <button
+              >
+                <FaIcon @icon="times" @prefix="fas" class="text-red" />
+              </button>
+            </div>
+          </div>
+          </li>
```
allthesignals, line 54: These chunks of code might make good components of their own! Unclear now though... good to keep it all in one component and then see where things go :)

client/tests/integration/components/packages/attachments-test.js
```diff
@@ -0,0 +1,129 @@
...
+      @package={{this.package}}
+    />`);
+
+    await assert.equal(findAll('[data-test-document-to-be-uploaded-name]').length, 0);
+
+    await selectFiles('#FileUploader123 > input', file, file2);
+
+    await assert.equal(findAll('[data-test-document-to-be-uploaded-name]').length, 2);
+    await assert.equal(find('[data-test-document-to-be-uploaded-name="0"]').textContent.trim(), 'PAS Form.pdf');
+    await assert.equal(find('[data-test-document-to-be-uploaded-name="1"]').textContent.trim(), 'Action Changes.excel');
```
allthesignals, line 127: Check out some of these nicer DOM assertion APIs: https://github.com/simplabs/qunit-dom

client/app/components/packages/attachments.js
godfreyyeung: Hmm at this point whether navigating away from the component clears the files is an open design issue I think. But I do agree it would be better if the File Manager and the file references aren't destroyed when navigating away. If component teardown did somehow clear the file manager,  existing and to-be-deleted file references would be lost. The to-be-uploaded file queue should still remain in memory. 

I think the same would go for if File Managers sat on the Package model, and the model teardown clears the File Manager

allthesignals: Why should it remain in memory? Wouldn't that risk memory leakage? We experienced this issue during the LUP build, I think we should try to avoid it, and that's ultimately what I'm suggesting here. Component teardown wouldn't automatically clear the files, we'd have to wire that up correctly.

Either way, I think we still should not have a component that implicitly (and conditionally) creates things in memory. I think it's a confusing design because it would surprise me if I read a property from an object, and the act of observing/reading creates things in memory.

godfreyyeung:
@allthesignals  Not ignoring your comments, just want to merge this in chunks!

client/app/components/packages/attachments.js
godfreyyeung: Oh, I didn't meant that part as a prescriptive statement. I meant that if a FileManager is deleted on teardown, it doesn't necessarily also delete the referenced fileQueue. We can for sure manually destroy the fileQueue too. 

Originally thought that memory leakage shouldn't be a big issue if there's only 10-15 editable packages on the dashboard at any given time. 

Though now that I'm thinking about it, if somehow the browser cache holds on to File Managers and File Queues indefinitely (because the FileManagement service is independent of loaded packages), I see how it will leak over months/years. Hmm

godfreyyeung: Okay, I get now why it may be ideal to create the File Manager+fileQueue on Package model init and destroy both on model didDelete()

allthesignals: Ah, gotcha! Makes sense :) 

Memory leakage to me in this situation means something like file uploads somehow persisting in-memory across different packages, even if they were navigate away from. It's not that it's a big deal in terms of performance, but it could confuse us from a maintainability perspective.

client/app/components/packages/attachments.hbs
```diff
@@ -0,0 +1,156 @@
...
+    <ul>
+      <li>Map of <a href="#">project location within the NYC coastal zone</a> including Special Area Designations</li>
+    </ul>
+    <b>FOR <a href="#">CITY MAP CHANGES</a>:</b>
+    <ul>
+      <li>Most recent Borough President’s final Section Map or most recent approved and filed Alteration Map for the project area</li>
+      <li>Site survey — recommended but not required</li>
+    </ul>
+  </div>
+  <div class="cell">
```
andycochran, line 39: All this stuff doesn't need to be in the component. It can go in the `<PasForm >` component (I've already got it in my PR — https://github.com/NYCPlanning/labs-applicant-portal/pull/109). This component only needs to be the part indicated in the wireframes by the gray line: 

image

godfreyyeung: Cool, thx I will remove!

Issue #20, opened 2020-03-12
Applicant team form


hannahkates:
## User Journey:
- User must fill in the Primary Applicant form.
- User can then add additional members to the applicant team by clicking one of the "Add team member" role buttons.
- Clicking on one of the "Add team member" buttons will generate an additional form for that new team member, with the role (Applicant, Representative, Environmental Consultant) displayed in the header the way that Primary Applicant is in the screenshot.

## Unresolved Questions:
- Should the completed member forms get collapsed or stay expanded and continue to be additive?

## [Figma Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP-Wireframes?node-id=0%3A1)
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/4f2b6760-35e1-457b-a8eb-3d530738f5cd)

## Implementation Notes

The Figma wireframes have more information on the exact names of the entities.

Consider encapsulating this inside its own component that receives an array of objects. New applicant information entities get pushed into that array through this form. The specific fields are available on the dcp_applicantinformation entity.

allthesignals:
![image.png](https://images.zenhubusercontent.com/5ca23922d9fcf5278d715977/5d6621d6-a58c-426c-8232-36bc75805c93)

Need to figure out what's going on here, what's teh difference

bfreeds:
@allthesignals @hannahkates do you know which field represents the "role"?  Is this a coded value that maps onto the four role types ("Primary Applicant", "Co-Applicant", "Representative", and "Environmental Consultant"?)
![image](https://user-images.githubusercontent.com/5316367/80241240-40637200-8631-11ea-8372-367a3c1c9ff0.png)

bfreeds:
@hannahkates  on the applicant information entities, I see two "types":
![image](https://user-images.githubusercontent.com/5316367/80255863-c4c2ee80-864b-11ea-9013-fb12f5ae6bf5.png)
Is this the field that we should be using for the 4 types we are using in the app? If so will they be added as valid options?  

Note this is separate in CRM from applicant information representative, which _does not_ have a "type" field.

PR #104, opened 2020-04-24
20:  Add frontend logic for applicant team editor

bfreeds:
This pr addresses #20 by adding the frontend form logic for adding and removing applicants.  The data generated by these form elements will need to be connected to the `Save` frontend logic and api request to CRM to make sure that the applicants are updated to their correct entities (#18).
--- 
### Changes
Creates two components:
- `packages/applicant-fieldset`
  - The form for a single applicant or applicant team member
- `packages/applicant-team-editor`
  - Renders a an applicant-fieldset (for "Applicant"), gives user control to add or remove additional applicant-fieldsets

Will create one model:
- `applicant`
  - Note:  this model covers the data model for the records that will need to be added to both the `Applicant Information` table and `Applicant Representative Table` in CRM.  There is a custom attribute in this Ember Model (`targetEntity`) that gives us the key to know which array/CRM entity the object will need to be synced with. 
  - Also Note:  the `applicantType` attribute on this model _only_ applies to objects being posted to the `Applicant Information` table in CRM.  
    - If `applicantType` = `Individual`, CRM Type = `Applicant`
    - If `applicantType` = `Organization`, CRM Type = `Authorized Applicant Representative`

Adds tests for `applicant-team-editor`.

Also creates one custom scss module:
- `.fieldset-adder`
  - Extends `.callout` class to customize for this design pattern used throughout the PasForm
image

---


client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+    <div class="cell auto medium-4">
+      <label>ZIP <input type="text"></label>
+    </div>
+  </div>
+  
+  <label>Phone <input type="text"></label>
+
+  {{!-- Remove button if not Primary Applicant --}}
+  {{#if (not-eq @title "Primary Applicant") }}
```
allthesignals, line 36: Nit that I waffle on this a bit, but what do you think about having this logic inside the component class as a computed property? Generally if the logic goes beyond "if truthy", I make a computed. Sometimes that doesn't work if you have a loop, but in this case it might make sense.

client/tests/integration/components/packages/applicant-team-editor-test.js
```diff
@@ -0,0 +1,41 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import EmberObject from '@ember/object';
+
+module('Integration | Component | packages/applicant-team-editor', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
```
allthesignals, line 10: I'm going to look at some ways of thinking about components that may/may not depend on models.

One thing I'm afraid of is having a component require instances of a certain class. Will think on it some more...

client/app/components/packages/applicant-fieldset.hbs
bfreeds: @allthesignals this ended up being amended after learning more about the two applicant tables in CRM, but I see your point and agree that it would be clearer to have anything but the most basic evaluations in a computed rather than doing anything complicated in the template logic.  The way this was written is also imperative and kinda hides the business logic being implemented.

client/app/components/packages/applicant-team-editor.hbs
```diff
@@ -0,0 +1,29 @@
+<section>
+  {{log "@applicants passed to component:"}}
+  {{log @applicants}}
```
allthesignals, line 3: Don't forget to remove these

PR #106, opened 2020-04-27
18 pas form server

allthesignals:
Refactoring code (partly on @godfreyyeung advice) to make controller module own its serialization configuration.

I took this and used an [interceptor](https://docs.nestjs.com/interceptors) to deal with the JSON:API conversion step. Resource-specific configuration is handled on their respective controllers, on interceptor instantiation.

image


server/src/contact/contact.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+@UseInterceptors(new JsonApiSerializeInterceptor('contacts', {
+  attributes: ['contactid', 'emailaddress1'],
+  id: 'contactid',
+}))
+@Controller('contacts')
+export class ContactController {
+  constructor(private readonly contactService: ContactService) {}
+
+  @Get('/')
+  async getUser(@Session() session) {
```
godfreyyeung, line 14: should be named getContact instead?

client/app/authenticators/zap-api-authenticator.js
```diff
@@ -28,7 +28,7 @@ export default class ZAPAuthenticator extends BaseAuthenticator {
   // a ZAP Token. This request will fail if that cookie doesn't exist
   // https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md#using-a-dedicated-endpoint
   async _fetchUserObject() {
```
godfreyyeung, line 30: should this be named _fetchContactObject() instead?

allthesignals: yessir! good catch

server/src/contact/contact.controller.ts
allthesignals: yessir! good catch

PR #113, opened 2020-04-30
PAS Form Routes, Controllers, Models, Handlers

allthesignals:
This PR addresses another part of the PAS Form system, the ability to post and patch data into the system.

### DIFF
--- a/client/app/models/pas-form.js
+++ b/client/app/models/pas-form.js
@@ -1,9 +1,14 @@
-import Model, { belongsTo, attr } from '@ember-data/model';
+import Model, { belongsTo, hasMany, attr } from '@ember-data/model';
 
 export default class PasFormModel extends Model {
   @belongsTo('package', { async: false })
   package;
 
+  @hasMany({ async: false })
+  applicants;
+
+  @hasMany({ async: false })
+  bbls;
 
   // Project Information
   @attr('string')
@@ -207,4 +212,12 @@ export default class PasFormModel extends Model {
 
   @attr('string')
   dcpProjectattachmentsotherinformation;
+
+  async saveDirtyBbls() {
+    return Promise.all(
+      this.bbls
+        .filter((bbl) => bbl.hasDirtyAttributes)
+        .map((bbl) => bbl.save()),
+    );
+  }
 }
--- a/client/app/models/pas-form.js
+++ b/client/app/models/pas-form.js
@@ -1,7 +1,7 @@
 import Model, { belongsTo, attr } from '@ember-data/model';
 
 export default class PasFormModel extends Model {
-  @belongsTo('pas-form', { async: false })
+  @belongsTo('package', { async: false })
   package;
 
 
--- a/client/app/serializers/package.js
+++ b/client/app/serializers/package.js
@@ -1,0 +1,7 @@
+import ApplicationSerializer from './application';
+
+export default class PackageSerializer extends ApplicationSerializer {
+  attrs = {
+    pasForm: { serialize: true },
+  };
+}
--- a/client/tests/unit/serializers/application-test.js
+++ b/client/tests/unit/serializers/application-test.js
@@ -1,0 +1,23 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Serializer | application', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let serializer = store.serializerFor('application');
+
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let record = store.createRecord('application', {});
+
+    let serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
+  });
+});
--- a/client/tests/unit/serializers/application-test.js
+++ b/client/tests/unit/serializers/application-test.js
@@ -6,17 +6,17 @@ module('Unit | Serializer | application', function(hooks) {
 
   // Replace this with your real tests.
   test('it exists', function(assert) {
-    let store = this.owner.lookup('service:store');
-    let serializer = store.serializerFor('application');
+    const store = this.owner.lookup('service:store');
+    const serializer = store.serializerFor('application');
 
     assert.ok(serializer);
   });
 
   test('it serializes records', function(assert) {
-    let store = this.owner.lookup('service:store');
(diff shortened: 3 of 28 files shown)


client/app/models/applicant.js
```diff
@@ -0,0 +1,5 @@
+import Model from '@ember-data/model';
```
bfreeds, line 1: Rebase off develop to bring in the new applicant model

allthesignals: Thanks!

client/app/serializers/package.js
```diff
@@ -0,0 +1,7 @@
+import ApplicationSerializer from './application';
+
+export default class PackageSerializer extends ApplicationSerializer {
+  attrs = {
+    pasForm: { serialize: true },
+  };
+}
```
allthesignals, line 7: This doesn't need to be here, i remove this in a follow up PR...

client/tests/unit/serializers/application-test.js
```diff
@@ -0,0 +1,14 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Serializer | application', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
```
bfreeds, line 8: Does this need to be:
```js
async function(assert)
```
Or are the two methods being called on lines 9 & 10 synchronous?

client/tests/unit/serializers/package-test.js
```diff
@@ -0,0 +1,23 @@
...
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function(assert) {
+    const store = this.owner.lookup('service:store');
+    const record = store.createRecord('package', {});
+
+    const serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
```
bfreeds, line 21: What does this assertion test?  Just that there's no errors when `record.serialize()` is invoked?

client/app/models/pas-form.js
```diff
@@ -1,9 +1,14 @@
-import Model, { belongsTo, attr } from '@ember-data/model';
+import Model, { belongsTo, hasMany, attr } from '@ember-data/model';
 
 export default class PasFormModel extends Model {
-  @belongsTo
+  @belongsTo({ async: false })
```
bfreeds, line 4: Wanted to learn what the tradeoff / benefits of making our model relationships synchronous.  Found [this article](https://embermap.com/notes/83-the-case-against-async-relationships), and it seems like a good way to make things easier to reason about.

client/tests/unit/serializers/package-test.js
allthesignals: Yep! Not a very good test.

allthesignals: I am removing these serializers in a later PR. They don't need to be there.

client/app/models/pas-form.js
allthesignals: Yes! That's exactly why. I believe @andycochran also had trouble because we had one of these rels set as asynchronous, it was returning a Proxy object.

Frankly, we should go through all of our apps and set it to false. And probably add a lint rule.

server/src/packages/packages.controller.ts
```diff
@@ -1,6 +1,8 @@
-import { Controller, Get, Param, UseInterceptors } from '@nestjs/common';
+import { Controller, Get, Param, UseInterceptors, Patch, Body, Session, HttpException, HttpStatus, UsePipes, UseGuards } from '@nestjs/common';
```
bfreeds, line 1: V. nit:  when these lines get long it's more readable to break each variable onto a newline (code formatters like Prettier would do this automatically)

server/src/packages/packages.controller.ts
```diff
@@ -119,22 +121,40 @@ import { JsonApiSerializeInterceptor } from '../json-api-serialize.interceptor';
...
+    if (!pasForm) {
+      return {
+        ...projectPackage,
+      };
+    } else {
+      return {
+        ...projectPackage,
+        'pas-form': {
+          ...pasForm,
+          applicants: pasForm.dcp_dcp_applicantinformation_dcp_pasform,
```
bfreeds, line 135: Just flagging that we're going to need to (at some point) specify the two CRM tables that our `applicants` will be sent to / retrieved from.  

_Think_ the name of the other applicant array is: `dcp_dcp_applicantrepresentativeinformation_dcp_pasform` but gotta double check that.

server/src/packages/packages.controller.ts
```diff
@@ -119,22 +121,40 @@ import { JsonApiSerializeInterceptor } from '../json-api-serialize.interceptor';
...
-  @Get('/:id')
-  packages(@Param('id') id) {
+  @Get('/packages/:id')
+  getPackage(@Param('id') id) {
     return this.packagesService.getPackage(id);
   }
+
+  @Patch('/packages/:id')
+  patchPackage(@Body() body, @Param('id') id) {
+    return {
```
bfreeds, line 155: Very clean & simple solution to the challenging ID problem

server/src/packages/packages.controller.ts
allthesignals: @bfreeds  will do, can we add it to our lint rules?

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { PasFormService } from '../pas-form.service';
+
+@Controller('applicants')
+export class ApplicantsController {
+  constructor(private readonly pasFormService: PasFormService) { }
+
+  @Patch('/:id')
+  patchApplicant(@Body() body, @Param('id') id) {
+    return {
+      dcp_applicantinformationid: id,
```
bfreeds, line 11: Similar comment to above there will be also `dcp_applicantrepinformation` that has to be patched as well.

client/tests/unit/serializers/package-test.js
bfreeds: Cool - was just checking as I'm trying to understand/learn test writing.

server/src/packages/packages.controller.ts
allthesignals: Ahhh yes yes you are correct.

Thank you.

I will make a follow up issue for this. We need to merge those things. Might be messy.

server/src/packages/packages.controller.ts
allthesignals: Tried getting this to work, might need some help: https://eslint.org/docs/rules/object-curly-newline

PR #127, opened 2020-05-01
Save PAS form on save button click

bfreeds:
This PR will add the save button frontend logic.

### DIFF
--- a/client/tests/integration/components/pas-form-test.js
+++ b/client/tests/integration/components/pas-form-test.js
@@ -1,6 +1,8 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, click, fillIn } from '@ember/test-helpers';
+import {
+  render, click, fillIn, settled,
+} from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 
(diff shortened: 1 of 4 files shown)


client/tests/integration/components/pas-form-test.js
```diff
@@ -91,4 +91,22 @@ module('Integration | Component | pas-form', function(hooks) {
...
+    await render(hbs`<PasForm @package={{this.package}} />`);
+    
+    // make it dirty
+    await fillIn('[data-test-dcprevisedprojectname]', 'Some Cool New Project Name');
+
+    await this.pauseTest();
+
+    // save it
+    // HOW TO VERIFY THIS BEHAVIOR!?
+    assert.ok(await click('[data-test-save-button'));
```
allthesignals, line 110: Some ideas: https://www.ember-cli-mirage.com/docs/testing/assertions

client/app/components/pas-form.js
```diff
@@ -5,6 +5,11 @@ import { tracked } from '@glimmer/tracking';
 export default class PasFormComponent extends Component {
   @tracked package;
 
+  @action
+  save(packageInstance) {
+    packageInstance.saveDescendants();
```
allthesignals, line 10: Thinking about this some more... this makes our pas-form component class and our package model turn into an example of [inappropriate intimacy](https://refactoring.guru/smells/inappropriate-intimacy).

This is fine for now but just some food for thought later down the road :)

allthesignals: One way to deal with this is to pass down a generic "onSave" handler. The Pas-Form component could simply call 'onSave' when the button is clicked, and fire the passed handler. It could be defined as a noop by default.

bfreeds: @allthesignals totally agree with you on this -- I added a TODO comment for now.

PR #129, opened 2020-05-01
23 backend document upload... + frontend refactor...

godfreyyeung:
Adds a server endpoint for uploading documents.

Also has some frontend refactoring of file managers -- attaching them to the Packages instead of managing them through a File Management service.

Todo:
  - [ ] Conditionally instantiate the File Managers
  - [ ] Verify document upload for Packages without an existing File Location


server/src/document/document.controller.ts
```diff
@@ -0,0 +1,77 @@
...
+    const headers = {
+      // Document will be uploaded as this user
+      MSCRMCallerID: this.config.get('CRM_SERVICE_CONTACT_ID'),
+    };
+
+    let uploadDocResponse = {};
+
+    const encodedBase64File = Buffer.from(file.buffer).toString('base64');
+
+    const packageGUID = instanceId;
```
allthesignals, line 53: Teeeeensy suggestion but you can do this on line 40: 

```
      body: {
        instanceId: packageGUID,
      },
```

though, idk, maybe it should just be instance id? reading this, I didn't assume it's specific to packages.

server/src/document/document.controller.ts
```diff
@@ -0,0 +1,77 @@
...
+    uploadDocResponse = await this.documentService.uploadDocument('dcp_package',
+        packageGUID,
+        folderName,
+        file.originalname,
+        encodedBase64File,
+        true,
+        headers
+      );
+
+    response.status(200).send({"message": uploadDocResponse});
```
allthesignals, line 73: Are you able to just return the line on 64? I think we will want to avoid using the Response and Request decorators unless there's a really good reason (setting cookies).

Reason for this is that if we use Response decorators, we forgo some NestJS features (like pipes, guards, etc). 

I'd try to just return lines 64-71 and get rid of the response param.

allthesignals:
Taking a look at this

godfreyyeung:
I've set up and got the Nest tests passing. I'm confused why Codecov is regressing. It seems to indicate `file-management.js` is receiving less coverage now?

allthesignals:
@godfreyyeung yeah the codecov is griping about Ember... let me see...

client/app/components/packages/attachments.js
```diff
@@ -9,18 +8,10 @@ import { action } from '@ember/object';
...
 export default class PackagesAttachmentsComponent extends Component {
-  constructor() {
-    super(...arguments);
-
-    this.fileManager = this.fileManagement.findOrCreate(
-      this.args.package.id,
-      this.args.package.documents,
-    );
+  get fileManager() { // should be an instance of FileManager
+    return this.args.fileManager;
```
godfreyyeung, line 12: Future REDO: We could potentially get fileManager from this.args.package.fileManager

allthesignals: My thinking on this is that this is a move towards having the component class less aware of the internal structure of the package model.

If we have components that know a lot about specific methods of model, specific shapes, etc, they're a little less encapsulated, harder to test, and more prone to breakage should anything change.

Thoughts?

PR #130, opened 2020-05-02
Move breadcrumbs to root; Add load spinner for packages;

allthesignals:
This commit moves the breadcrumbs to the root of the packages route. This is needed because we want the breadcrumbs to be visible across all sub routes.

Additionally, this adds a load spinner for the packages route using the ember substate feature. This is done by adding a 'loading.hbs' template in the packages route. This route is triggered when the model is being resolved. I added this because this can be a large request, and without it, it would look like nothing is happening.

### DIFF
--- a/client/app/templates/packages/edit.hbs
+++ b/client/app/templates/packages/edit.hbs
@@ -1,7 +1,3 @@
-<Breadcrumbs />
-
-<hr class="large-margin-bottom">
-
 <Packages::PasForm::Edit @package={{@model}} />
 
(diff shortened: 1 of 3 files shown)


client/app/templates/packages/edit.hbs
```diff
@@ -1,5 +1,3 @@
-<Breadcrumbs />
-
 <hr class="large-margin-bottom">
```
andycochran, line 1: Move this hr along with the Breadcrumbs component.

PR #131, opened 2020-05-02
Delete things

allthesignals:
Delete the things!

Adds server side routes for handling delete methods


server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -58,4 +58,13 @@ export class ApplicantsController {
       return this.crmService.create('dcp_applicantinformations', allowedAttrs);
     }
   }
+
+  @Delete('/:id')
```
bfreeds, line 62: Nice, like the way nest.js gives us this.

PR #135, opened 2020-05-04
Enable e2e backend tests; Add to CI

allthesignals:
This PR enables e2e testing in CircleCI. Adds environment variable mocking in the doc upload test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,11 @@ jobs:
           name: run tests
           command: yarn test
 
+      # run tests!
+      - run:
+          name: run tests
+          command: yarn test:e2e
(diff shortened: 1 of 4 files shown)


allthesignals:
> Ohhhhhhh..... 💡 💡 💡 💡
> 
> That's so interesting. I thought e2e tests on CircleCI would still pull from CI envs.
> 
> Thanks Matt!!

Sooo I think we wanna move away from needing a dot-env file for test environments. It's just hard to deal with and introduces too many variables.

The main reason e2e wasn't running is that we didn't have it listed in the CircleCI config.

Anyway, thank you so much for adding these tests!

PR #138, opened 2020-05-04
22 land use action component

trbmcginnis:
- This PR creates a LandUseAction component where a user can select an action from the dropdown menu and then answer additional questions about that added action. 
- The PR uses a child component action-input-form to standardize form inputs since there are additional questions required by many actions (some of which are similar to other actions).

**Questions:**
- How do other engineers prefer to store lookups? Outside of the file where it's used or inside of it? I left the lookup inside the file for easy access, but am open to suggestions. 
- A lot of lookups and naming going on here! Will take any suggestions on names

**TODO:**
- Working on design update final polishing
- Double check that attributes included in the lookup are correct based on CRM fields (double checking and polishing up)

### DIFF
--- a/client/app/components/action-form-input.hbs
+++ b/client/app/components/action-form-input.hbs
@@ -1,0 +1,11 @@
+<Input
+  class="action-form-input"
+  type={{@fieldType}}
+  maxlength="200"
+  placeholder=""
+  autocomplete="off"
+  @value={{this.newVal}}
+  @key-down={{fn (mut this.newVal)}}
+  @key-up={{fn this.updateAttr @landUseActionField this.newVal}}
+  data-test-input="{{@landUseActionField}}"
+/>
--- a/client/app/components/action-form-input.js
+++ b/client/app/components/action-form-input.js
@@ -1,0 +1,14 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class ActionFormInputComponent extends Component {
+  // value that field is to be changed to
+  @tracked newVal;
+
+  // updates target attribute on key-up event
+  @action
+  updateAttr(attr, newValue) {
+    this.args.pasForm[attr] = newValue;
+  }
+}
--- a/client/app/components/land-use-action.hbs
+++ b/client/app/components/land-use-action.hbs
@@ -1,0 +1,17 @@
+<div class="large-margin-bottom">
+  <label data-test-land-use-action-picker>
+    Add a Proposed Action:
+    <PowerSelect
+      supportsDataTestProperties={{true}}
+      @placeholder="-- select an action --"
+      @searchEnabled={{false}}
+      @selected={{this.selectedLandUseAction}}
+      @options={{this.landUseActionTextOptions}}
+      @onchange={{fn (mut this.selectedLandUseAction)}}
+      @onclose={{fn this.setCount}}
+      as |landUseAction|
+    >
+      {{landUseAction}}
+    </PowerSelect>
+  </label>
+</div>
--- a/client/app/components/land-use-action.hbs
+++ b/client/app/components/land-use-action.hbs
@@ -5,13 +5,76 @@
       supportsDataTestProperties={{true}}
       @placeholder="-- select an action --"
       @searchEnabled={{false}}
-      @selected={{this.selectedLandUseAction}}
-      @options={{this.landUseActionTextOptions}}
+      @options={{this.landUseActionOptions}}
       @onchange={{fn (mut this.selectedLandUseAction)}}
-      @onclose={{fn this.setCount}}
+      @onclose={{fn this.setSelectedActionsArray}}
       as |landUseAction|
     >
-      {{landUseAction}}
+      {{landUseAction.name}}
     </PowerSelect>
   </label>
 </div>
+{{#each this.landUseActionSelections as |landUseAction|}}
+  <fieldset class="fieldset">
+    <legend>
+      <strong>{{landUseAction.name}}</strong>
+    </legend>
+    <label class="action-form-label">
+    How many {{landUseAction.name}} actions?
+      <ActionFormInput @pasForm={{@pasForm}} @fieldType="number" @landUseActionField={{landUseAction.field}}/>
+    </label>
+    {{#if (or (eq landUseAction.name "Zoning Special Permit") (eq landUseAction.name "Zoning Certification") (eq landUseAction.name "Zoning Authorization"))}}
+      <label class="action-form-label">
+        Where in the Zoning Resolution can this action be found?
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr1}}/>
+        <span class="help-text">Provide the Zoning Resolution section number. Ex. ZR Sec. 74-711</span>
+      </label>
+      <label class="action-form-label">
+        Which sections of the Zoning Resolution does this modify?
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr2}}/>
+        <span class="help-text">Provide the Zoning Resolution section number(s). Ex. ZR Sec. 42-10 and 43-17</span>
+      </label>
+    {{/if}}
+    {{#if (eq landUseAction.name "Zoning Map Amendment")}}
+      <label class="action-form-label">
+        Existing Zoning Districts:
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr1}}/>
+        <span class="help-text">ex. R7-1/C2-4</span>
+      </label>
+      <label>
+        Proposed Zoning Districts:
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr2}}/>
+        <span class="help-text">ex. C4-5X</span>
+      </label>
+    {{/if}}
+    {{#if (eq landUseAction.name "Zoning Text Amendment")}}
+      <label class="action-form-label">
+        Affected ZR Section Number:
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr1}}/>
+        <span class="help-text">Provide the Zoning Resolution section number. Ex. ZR Sec. 74-711</span>
+      </label>
+      <label class="action-form-label">
+        Affected ZR Section Title:
+        <ActionFormInput @pasForm={{@pasForm}} @fieldType="text" @landUseActionField={{landUseAction.attr2}}/>
+        <span class="help-text">Provide the Zoning Resolution section Title. Ex. EXAMPLE</span>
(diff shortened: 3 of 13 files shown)


client/app/styles/modules/_pasform.scss
```diff
@@ -5,4 +5,12 @@
...
   border-width: 2px;
   border-style: dashed;
+}
+
+.action-form-input {
+  margin: 0 0 0rem
+}
+
+.action-form-label {
+  margin: 0 0 1rem
```
andycochran, line 15: Are these styles necessary?

client/app/components/land-use-action.hbs
```diff
@@ -0,0 +1,82 @@
+<h5>Proposed Land Use Actions</h5>
+<p>What land use actions does the proposed project require? Identify all that applyand fill in the appropriate related information. Note: This is preliminary information and may change following discussion with NYC Planning.</p>
```
andycochran, line 2: This content is already in the PasForm component. The LandUseAction component should just be the dropdown and dynamic inputs.

andycochran: You could also go ahead an plug this component into the pas-form/edit.hbs, swapping out lines 55-56: 
```
{{!-- TODO: Add land-use-actions-editor component --}}
<div class="bg-red-dim xlarge-padding medium-margin-bottom">land-use-actions-editor component</div>
```

client/app/styles/modules/_pasform.scss
trbmcginnis: **Before styling:**

Screen Shot 2020-05-04 at 4 13 02 PM

**After Styling:**

Screen Shot 2020-05-04 at 4 13 38 PM

client/app/components/land-use-action.hbs
trbmcginnis: Okay good to know, I'll remove it!

client/app/components/action-form-input.hbs
```diff
@@ -0,0 +1,11 @@
+<Input
+  class="action-form-input"
+  type={{@fieldType}}
+  maxlength="200"
+  placeholder=""
+  autocomplete="off"
+  @value={{this.newVal}}
+  @key-down={{fn (mut this.newVal)}}
+  @key-up={{fn this.updateAttr @landUseActionField this.newVal}}
```
allthesignals, line 9: does the `mut` helper not work for this?

client/app/components/action-form-input.js
```diff
@@ -0,0 +1,23 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class ActionFormInputComponent extends Component {
+  // represents pasForm model
+  @tracked pasForm;
+
+  // attribute on pasForm model, e.g. 'dcpPfchangeincitymap'
+  @tracked landUseActionField;
```
allthesignals, line 10: It looks like the 3 arguments for this compnent are : 
`pasForm, fieldType, landUseActionField`

Those 3 don't need to be tracked. If you want to provide default values, see comment above :)

client/app/components/action-form-input.js
```diff
@@ -0,0 +1,23 @@
...
+
+  // value that field is to be changed to
+  @tracked newVal;
```
allthesignals, line 13: Good

client/app/components/action-form-input.js
```diff
@@ -0,0 +1,23 @@
...
+
+  // input type, e.g. "text" or "number"
+  @tracked fieldType;
+
+  // updates target attribute on key-up event
+  @action
+  updateAttr(attr, newValue) {
+    this.args.pasForm[attr] = newValue;
+  }
```
allthesignals, line 22: Ah, I think I see the problem... `args` are immutable, so you have to have this action here that updates the value. Interesting.

client/app/components/land-use-action.js
```diff
@@ -0,0 +1,100 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class LandUseActionComponent extends Component {
+  // represents pasForm model
+  @tracked pasForm;
+
+  // local variable that records results of dropdown selection
+  selectedLandUseAction = {};
```
allthesignals, line 10: Good distinction between local variable and an argument variable... 

How come we need to store state on this component? Is there some reason the model alone can't be responsible for that?

client/app/components/land-use-action.js
```diff
@@ -0,0 +1,100 @@
...
+
+  // actions selected by user in dropdown
+  @tracked landUseActionSelections = [];
+
+  // lookup to match action titles to their associated attributes
+  landUseActionOptions = [
```
allthesignals, line 16: Just a stylistic choice of mine, but I'd keep this static array as a constant up above, then just set it down here :) 

```
const LAND_USE_ACTION_OPTIONS = [ /* ... */ ];

// class stuff..

landUseActionOptions = [
  ...LAND_USE_ACTION_OPTIONS
];

```

In my opinion, it keeps the function of the class more focused :)

client/app/models/pas-form.js
```diff
@@ -21,52 +21,52 @@ export default class PasFormModel extends Model {
...
-  @attr('string')
+  @attr('number')
   dcpPfzoningspecialpermit;
 
-  @attr('string')
+  @attr('number')
   dcpZoningspecialpermitpursuantto;
 
-  @attr('string')
+  @attr('number')
```
allthesignals, line 63: Thank you!!

client/tests/integration/components/action-form-input-test.js
```diff
@@ -0,0 +1,30 @@
...
+    const projectPackage = this.server.create('package');
+
+    this.package = await this.owner.lookup('service:store').findRecord('package', projectPackage.id, { include: 'pas-form' });
+
+    this.actionField = 'dcpPfchangeincitymap';
+
+    // Template block usage:
+    await render(hbs`
+    <ActionFormInput @pasForm={{this.package.pasForm}} @fieldType="number" @landUseActionField={{this.actionField}}/>
+    `);
```
allthesignals, line 21: Just for sake of consistency, you could get rid of your variable `actionField` and just invoke like this: 

```
    await render(hbs`
    <ActionFormInput
      @pasForm={{this.package.pasForm}} @fieldType="number" @landUseActionField= "dcpPfchangeincitymap"/>
    `);
```

Just tiny nit to keep your concise test even leaner :)

client/tests/integration/components/land-use-action-test.js
```diff
@@ -0,0 +1,122 @@
...
+} from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { selectChoose } from 'ember-power-select/test-support';
+
+module('Integration | Component | land-use-action', function(hooks) {
+  setupRenderingTest(hooks);
+  setupMirage(hooks);
+
+  test('User can add an action and answer extra questions', async function(assert) {
```
allthesignals, line 17: Really like the plain-spoken test name here, written in terms of the UX

client/tests/integration/components/land-use-action-test.js
```diff
@@ -0,0 +1,122 @@
...
+    await triggerEvent('[data-test-input="dcpZoningspecialpermittomodify"]', 'keyup');
+
+    assert.equal(this.package.pasForm.dcpZoningspecialpermittomodify, 'Permit 7A');
+
+    await click('[data-test-delete-button="Zoning Special Permit"]');
+
+    assert.equal(this.package.pasForm.dcpPfzoningspecialpermit, 0);
+    assert.equal(this.package.pasForm.dcpZoningspecialpermitpursuantto, '');
+    assert.equal(this.package.pasForm.dcpZoningspecialpermittomodify, '');
+  });
```
allthesignals, line 121: This deletion test is probably a little too long — it seems like we want to prove that we can delete stuff, but are there really this many clicks involved to set it up for asserting for deletion?

client/tests/integration/components/land-use-action-test.js
```diff
@@ -0,0 +1,122 @@
...
+    this.package = await this.owner.lookup('service:store').findRecord('package', projectPackage.id, { include: 'pas-form' });
+
+    // Template block usage:
+    await render(hbs`
+      <LandUseAction @pasForm={{this.package.pasForm}}>
+      </LandUseAction>
+    `);
+
+    // Check that we can delete "Change in CityMap"
+    await selectChoose('[data-test-land-use-action-picker]', 'Change in CityMap');
```
allthesignals, line 80: Hmmm this comment refers to the multiple lines below it right?

client/tests/integration/components/land-use-action-test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    assert.dom(this.element).includesText('How many Change in CityMap actions?');
+
+    assert.ok(!this.package.pasForm.dcpPfchangeincitymap);
```
allthesignals, line 84: Assert it's still empty, yes?

client/tests/integration/components/land-use-action-test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    await triggerEvent('[data-test-input="dcpPfchangeincitymap"]', 'keyup');
+
+    assert.equal(this.package.pasForm.dcpPfchangeincitymap, 4);
+
+    await click('[data-test-delete-button="Change in CityMap"]');
+
+    assert.dom(this.element).doesNotIncludeText('How many Change in CityMap actions?');
+
+    assert.equal(this.package.pasForm.dcpPfchangeincitymap, 0);
```
allthesignals, line 96: Great, this seems like this should be the end of the test. What is the rest for? Maybe there's some other behavior you want to check? Seems like it's another part of the form. I think if that's the case, it should be a separate `test` block.

On the other hand, if you are throwing in these extra assertions just to make _extra_ sure that your component works, I'd say you should give yourself some credit: the test you have already is concise and tests for the behavior you want.

client/tests/integration/components/land-use-action-test.js
trbmcginnis: I can definitely make it shorter

client/tests/integration/components/land-use-action-test.js
trbmcginnis: Going to make this test shorter anyway, this will be removed

client/tests/integration/components/land-use-action-test.js
trbmcginnis: I'm trying to assert that it's falsey (because the default is undefined)

client/app/components/action-form-input.js
```diff
@@ -0,0 +1,23 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class ActionFormInputComponent extends Component {
+  // represents pasForm model
+  @tracked pasForm;
```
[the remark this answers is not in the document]
trbmcginnis: Amazing, will update, thanks!

Issue #121, opened 2020-05-01
"City Register File Number" input should only appear conditionally


hannahkates:
The "City Register File Number" input should only appear conditionally if they answer yes to the previous question about legal instruments.

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/ecfefd5e-b771-44ef-aa61-5f8d26802f19)

PR #143, opened 2020-05-04
121: make city register question conditional

hannahkates:
Addresses #121. The city register question should only appear if the previous dcprestrictivedeclaration answer is yes/true.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -398,16 +398,17 @@
           onClick={{fn this.updateAttr @package.pasForm "dcpRestrictivedeclaration" false}}
           data-test-dcprestrictivedeclaration-no
         /><label for="dcprestrictivedeclaration-no">No</label>
+        {{#if (eq @package.pasForm.dcpRestrictivedeclaration true)}}
+          <label>
+            <strong>City Register File Number</strong>
+            <Input
+              @type="text"
+              @value={{@package.pasForm.dcpCityregisterfilenumber}}
+              data-test-dcpcityregisterfilenumber
+            />
+          </label>
+        {{/if}}
       </fieldset>
-
-      <label>
-        <strong>City Register File Number</strong>
-        <Input
-          @type="text"
-          @value={{@package.pasForm.dcpCityregisterfilenumber}}
-          data-test-dcpcityregisterfilenumber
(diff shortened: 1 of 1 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -398,17 +398,18 @@
           onClick={{fn this.updateAttr @package.pasForm "dcpRestrictivedeclaration" false}}
           data-test-dcprestrictivedeclaration-no
         /><label for="dcprestrictivedeclaration-no">No</label>
+        {{#if (eq @package.pasForm.dcpRestrictivedeclaration true)}}
```
allthesignals, line 401: This is defined as a string in the model. I worry that the conditional won't work unless we make sure it is a real boolean and that CRM considers it a boolean.

allthesignals: Happy to do this but can't suggest a commit here 

```
  @attr('string')
  dcpRestrictivedeclaration;
```

W just need that to be a boolean:

```
  @attr('boolean')
  dcpRestrictivedeclaration;
```

hannahkates: Thanks I'll make that update

hannahkates: Actually I think it's okay to leave it as is? This question should work the same as `dcpProjectareaindustrialbusinesszone` and `dcpProjectareaindutrialzonename` which appear to be working correctly. 

[`dcpProjectareaindustrialbusinesszone` is saved as "string" on the model](https://github.com/NYCPlanning/labs-applicant-portal/blob/develop/client/app/models/pas-form.js#L128), and the template logic uses [`eq @package.pasForm.dcpProjectareaindustrialbusinesszone true`](https://github.com/NYCPlanning/labs-applicant-portal/blob/develop/client/app/components/packages/pas-form/edit.hbs#L272)

allthesignals: Ah then those are wrong because they will get serialized as strings.

We'll need to fix those as well.

allthesignals: Oops, also forgot to mention, you cna also just say `   {{#if @package.pasForm.dcpRestrictivedeclaration}}`

PR #146, opened 2020-05-05
Fix attr fields, align with CRM

allthesignals:
Updates fields to correct names; provides all PAS form fields

### DIFF
--- a/server/src/packages/packages.controller.ts
+++ b/server/src/packages/packages.controller.ts
@@ -14,13 +14,16 @@ export const PACKAGE_ATTRS = [
   'dcp_packagetype',
   'dcp_visibility',
   'versionnumber',
-  'documents',
 ];
 
 @UseInterceptors(new JsonApiSerializeInterceptor('packages', {
   id: 'dcp_packageid',
   attributes: [
     ...PACKAGE_ATTRS,
+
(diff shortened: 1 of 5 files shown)


server/src/packages/packages.controller.ts
```diff
@@ -90,7 +93,7 @@ export class PackagesController {
   @Patch('/:id')
   async patchPackage(@Body() body, @Param('id') id) {
     const allowedAttrs = pick(body, PACKAGE_ATTRS);
-
+    console.log(allowedAttrs);
```
trbmcginnis, line 96: Are you keeping this on purpose?

allthesignals: nope! good catch

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -44,7 +44,7 @@ export class ApplicantsController {
   @Post('/')
   create(@Body() body) {
     const allowedAttrs = pick(body, APPLICANT_ATTRIBUTES);
-
+    console.log(allowedAttrs);
```
trbmcginnis, line 47: Do we want to keep this?

PR #150, opened 2020-05-05
Use destroyRecord instead of store deleteRecord

allthesignals:
This uses the destroyRecord api instead of deleting through the store.

For reasons I don't know yet, it isn't picking up store.deleteRecord calls.

### DIFF
--- a/client/app/components/packages/applicant-team-editor.js
+++ b/client/app/components/packages/applicant-team-editor.js
@@ -20,7 +20,8 @@ export default class ApplicantTeamEditorComponent extends Component {
   @action
   removeApplicant(applicant) {
     // remove the applicant from the ember store
-    this.store.deleteRecord(applicant);
+    // this.store.deleteRecord(applicant);
+    applicant.destroyRecord();
 
     // remove the DOM node by removing the object from the array passed to the editor component
     this.args.applicants.removeObject(applicant);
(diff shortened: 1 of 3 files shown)


client/app/components/packages/applicant-team-editor.js
```diff
@@ -20,7 +20,8 @@ export default class ApplicantTeamEditorComponent extends Component {
   @action
   removeApplicant(applicant) {
     // remove the applicant from the ember store
-    this.store.deleteRecord(applicant);
+    // this.store.deleteRecord(applicant);
+    applicant.destroyRecord();
 
     // remove the DOM node by removing the object from the array passed to the editor component
     this.args.applicants.removeObject(applicant);
```
godfreyyeung, line 27: This smells not Ember-ish.
```
    this.args.applicants.removeObject(applicant);
```

allthesignals: Hmm what do you mean?

allthesignals: destroyRecord is a method on ember models

allthesignals: also what is "ember-ish"?

godfreyyeung: Oh I was referring to line 27. It modified the arguments to the component, as opposed to Data Down Action Up. That's interesting we have to do that manual cleanup

PR #151, opened 2020-05-05
Fix Form Anchors

andycochran:
This PR fixes the PAS Form page anchors so that the sidebar menu works, offsetting the scroll position to account for the sticky header.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -9,8 +9,11 @@
       NYC Planning understands that some projects may not be fully formed at the submission of the PAS. Give as much detailed information as possible. Failure to provide detailed, relevant information will result in the PAS being rejected.
     </p>
 
-    <section class="form-section" id="project-info">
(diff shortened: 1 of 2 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -820,37 +845,37 @@
...
           <li class="medium-margin-bottom">
-            <a href="#">
+            <a href="#project-description">
               <FaIcon @icon="chevron-circle-down" @fixedWidth={{true}} class="text-silver" />
               Project Description
             </a>
           </li>
           <li class="medium-margin-bottom">
-            <a href="#">
+            <a href="#attached-documents">
```
allthesignals, line 878: These still have the `#`

andycochran: The `href` are supposed to. The `id` are not.

allthesignals: oooyyyyyyyy duh! thanks

PR #154, opened 2020-05-05
Submit form, transition to show

allthesignals:
Adds submit button. Cleans up show template

### DIFF
--- a/client/app/components/packages/pas-form/edit.js
+++ b/client/app/components/packages/pas-form/edit.js
@@ -1,8 +1,11 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
 
 export default class PasFormComponent extends Component {
+  @service router;
+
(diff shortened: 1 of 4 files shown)


client/app/components/packages/pas-form/edit.js
```diff
@@ -14,6 +17,13 @@ export default class PasFormComponent extends Component {
     packageInstance.saveDescendants();
   }
 
+  @action
+  async submit(packageInstance) {
```
godfreyyeung, line 21: We've been using `projectPackage`. Couldn't find any existing instances of `packageInstance` as a parameter name

allthesignals: okay will change

godfreyyeung: okay, i lied, just one instance in the `edit.js` file.

but there's 47 uses of projectPackage.

this is a super nitpick :p

allthesignals: nope you're right, i didn't like the naming there either. we know it's an instance anyways!

PR #155, opened 2020-05-05
Conditionally instantiate File Managers on packages

godfreyyeung:
Only create File Managers on Applicant Packages,
not on Planning Packages

### DIFF
--- a/client/app/models/package.js
+++ b/client/app/models/package.js
@@ -2,16 +2,36 @@ import Model, { attr, belongsTo } from '@ember-data/model';
 import { inject as service } from '@ember/service';
 import FileManager from '../services/file-manager';
 
+// TODO: Get theses codes from a Constants service
+const PACKAGE_STATUS_CODES = {
+  PACKAGE_PREPARATION: 'Package Preparation',
+  CERTIFIED: 'Certified',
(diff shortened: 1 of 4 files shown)


client/app/models/package.js
```diff
@@ -2,16 +2,36 @@ import Model, { attr, belongsTo } from '@ember-data/model';
...
+      const fileQueue = this.fileQueue.create(this.id);
+
+      this.fileManager = new FileManager(
+        this.id,
+        this.documents,
+        [],
+        fileQueue,
+      );
+    }
   }
```
allthesignals, line 35: Did you see the deprecation warning link I sent? If we're going to refactor this, we should just refactor it away from the deprecated API.

godfreyyeung: Yeah, I wanted to do this in small chunks tho! I agree. 
I looked into using the afterModel hook, but felt it was too many changes in one commit

allthesignals: I'm not sure why this change is necessary. We don't need to optimize for this. I remember we talked about this, it's a lot of code to add just to add an optimization that we're not sure is even an issue to begin with.

godfreyyeung: Hmm ok, I can close this.

Issue #161, opened 2020-05-05
View page needs to list actual data (+brainstorm refactor)


hannahkates:
## What should happen:
The view-only page after PAS submission should display all the data you user just submitted

## What happened instead:
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/12f4e3e4-1b29-46fd-a0f5-d478f34cec76)

bfreeds:
Quick fix to show real data has been merged into `develop` and is in an open PR to be brought into `qa` for testing.  

Next steps will be to refine the styling and refactor for code maintainability, however once that PR is brought into QA this will not be blocking testing.

PR #165, opened 2020-05-05
Reconstruct Selected Land Use Actions on PAS form

godfreyyeung:
If PAS Form has Land Use Actions saved already, show them on edit page

Additionally, refactor to use more getters to maintain the list of action selections
and options, instead of using direct array manipulation.


client/app/components/land-use-action.js
```diff
@@ -94,4 +94,16 @@ export default class LandUseActionComponent extends Component {
     this.args.pasForm[actionToRemove.attr1] = '';
     this.args.pasForm[actionToRemove.attr2] = '';
   }
+
+  @action
+  reconstructLandUseActionSelections () {
+    const { pasForm } = this.args;
+    this.landUseActionOptions.map((option) => { //eslint-disable-line
```
allthesignals, line 101: Also, this doesn't need to be an array `.map` because `.map` returns a new array. In this situation, it should be a `forEach`.

That said, I wonder if this could be broken out a little more into a separate getter. For example, you could have a getter: 

```javascript
get answeredQuestions() {
  const { pasForm } = this.args;

  return this.landUseActionOptions.filter(option => pasForm[option.field] || pasForm[option.attr1] || pasForm[option.attr2]);
}
```

That, in addition to the sorter getter, would be nice!

client/app/components/land-use-action.js
```diff
@@ -94,4 +94,16 @@ export default class LandUseActionComponent extends Component {
     this.args.pasForm[actionToRemove.attr1] = '';
     this.args.pasForm[actionToRemove.attr2] = '';
   }
+
+  @action
+  reconstructLandUseActionSelections () {
```
allthesignals, line 99: Not too sure about this name... it's doing quite a lot here, and I think most of the things in this action could be used inside getters. In fact, once it's in getters, we probably won't need this action or the `did-insert` because we could use the getter in place of `this.landUseActionSelections` in the template:

```handlebars
{{#each this.answeredLandUseActions as |action|}}
  {{action}}
{{/each}}
```

client/app/components/action-form-input.hbs
```diff
@@ -8,4 +8,5 @@
   @key-down={{fn (mut this.newVal)}}
   @key-up={{fn this.updateAttr @landUseActionField this.newVal}}
   data-test-input="{{@landUseActionField}}"
+  {{did-insert (fn this.setNewVal @landUseActionField)}}
```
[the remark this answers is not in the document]
godfreyyeung: 💯 🔥 🎉 you're right! `mut` totally works here. refactored!

client/app/components/land-use-action.js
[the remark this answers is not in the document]
godfreyyeung: moved away from the map!

client/app/components/land-use-action.js
```diff
@@ -94,4 +94,16 @@ export default class LandUseActionComponent extends Component {
...
+    const { pasForm } = this.args;
+    this.landUseActionOptions.map((option) => { //eslint-disable-line
+      if (pasForm[option.field] || pasForm[option.attr1] || pasForm[option.attr2]) {
+        this.landUseActionOptions.removeObject(option);
+        this.landUseActionOptions.sort((a, b) => ((a.name > b.name) ? 1 : -1));
```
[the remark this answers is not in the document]
godfreyyeung: Done! :)

client/tests/acceptance/pas-form-test.js
```diff
@@ -0,0 +1,25 @@
...
+    await visit('/packages/1/edit');
+
+    assert.dom('[data-test-input="dcpPfzoningcertification"]').hasValue('21');
+    assert.dom('[data-test-input="dcpZoningpursuantto"]').hasValue('some value');
+    assert.dom('[data-test-input="dcpZoningtomodify"]').hasValue('some other val');
+    assert.dom('[data-test-input="dcpPfzoningtextamendment"]').hasValue('2');
+    assert.dom('[data-test-input="dcpAffectedzrnumber"]').hasNoValue();
+    assert.dom('[data-test-input="dcpZoningresolutiontitle"]').hasNoValue();
+    assert.dom('[data-test-input="dcpPfchangeincitymap"]').hasValue('4');
+  });
```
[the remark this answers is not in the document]
godfreyyeung: Done! :)

client/app/components/land-use-action.js
godfreyyeung: You're right, great suggestion!

client/app/components/land-use-action.js
godfreyyeung: Great suggestion!! Done!

godfreyyeung:
> Thanks for fixing this :) I commented some ideas for making the code easier to maintain for us later on...
> 
> I want to bring up another concern about the component and probably need @trbmcginnis' input on this, but is there any way we can take advantage of the two-way binding from the Input helper? For example:
> 
> ```
> <Input
>   type={{@fieldType}}
>   @value={{get @pasForm @landUseActionField}}
> />
> ```
> 
> Or something? I have a feeling Taylor ran into this already... but there might be a helper for dealing with this.
> 
> EDIT:
> 
> Another possibility for the `<LandUseAction />` component is to have it yield out preconfigured `<Input />` components, but with those other things filled out:
> 
> PasForm template:
> 
> ```
> <LandUseAction as  |action|>
>   <action.input
>     @fieldType=''text"
>     @value={{@pasForm.theActualAttr}}
>   />
> </LandUseAction>
> ```
> 
> @trbmcginnis @godfreyyeung I think it'd be really worth spending time watching this: https://embermap.com/topics/forms/our-first-contextual-form

I was thinking of this too. It seems the action-form-input should be bound directly ("downward") to the model attributes, instead of through a component newVal property. However, I think maybe the original intent was to provide some more abstraction to handle different fields and input types? 

Also think it should maybe be for another PR/refactor discussion.

allthesignals:
> However, I think maybe the original intent was to provide some more abstraction to handle different fields and input types?

Yep, I think we can still provide that abstraction through contextual components. Happy to talk through?

> Also think it should maybe be for another PR/refactor discussion.

Sounds good! Want to keep going with the refactor after this PR?

allthesignals:
@godfreyyeung this looks really great by the way!

godfreyyeung:
> > However, I think maybe the original intent was to provide some more abstraction to handle different fields and input types?
> 
> Yep, I think we can still provide that abstraction through contextual components. Happy to talk through?
> 
> > Also think it should maybe be for another PR/refactor discussion.
> 
> Sounds good! Want to keep going with the refactor after this PR?

Sounds great! I have a feeling it'll be related with the form components refactor that @bfreeds  and @andycochran  are working on for https://github.com/NYCPlanning/labs-applicant-portal/issues/161

Thanks for all your feedback! ❤️

client/app/components/land-use-action.hbs
```diff
@@ -5,9 +5,9 @@
       supportsDataTestProperties={{true}}
       @placeholder="-- select an action --"
       @searchEnabled={{false}}
-      @options={{this.landUseActionOptions}}
+      @options={{this.sortedAvailableLandUseActionOptions}}
       @onchange={{fn (mut this.selectedLandUseAction)}}
-      @onclose={{fn this.setSelectedActionsArray}}
+      @onclose={{fn this.addNewLandUseActionSelection}}
```
godfreyyeung, line 10: Flag for later: if we close the dropdown without selecting anything, we generate a blank Land Use Action input box (not tied to any PAS form attributes)

allthesignals: Good catch! Can you make the action check for this? `if this.selectedAction //...`

godfreyyeung: Good idea, done!

trbmcginnis:
Thanks @godfreyyeung this is great. Definitely in support of `@value={{get @pasForm @landUseActionField}}` I think I definitely overcomplicated the solution by trying to use the `key-down` action, was just about to go back and try to remove those but I saw that ya'll already talked about that on this PR.

allthesignals:
> Thanks @godfreyyeung this is great. Totally in support of `@value={{get @pasForm @landUseActionField}}` I think I definitely overcomplicated the solution by trying to use the `key-down` action, was thinking about ridding the app of input key-down actions and saw that ya'll already talked about that on this PR. I don't think it needs to go into another PR refactor -- but if you don't want to do it in this PR, I was going to go ahead and change that today anyway. Up to you
> 
> Not sure how necessary contextual components would be since most of the inputs are quite similar (the only difference being the pasForm field and the "type" of input). Unless we want to include the `label` in that contextual component like they did in the video? <-- but that would basically be the same amount of code space as before so don't think it's particularly beneficial

Contextual components would get us away from `@value={{get` stuff. Are you sure that using `get` as the @value there actually 2-way binds? The reason for that is 2-way binding would helps us move away from relying on key-down actions.

Contextual components would allow us to actually pass in explicit model attributes instead of doing these meta method.

godfreyyeung:
Hmm. 

> > Thanks @godfreyyeung this is great. Totally in support of `@value={{get @pasForm @landUseActionField}}` I think I definitely overcomplicated the solution by trying to use the `key-down` action, was thinking about ridding the app of input key-down actions and saw that ya'll already talked about that on this PR. I don't think it needs to go into another PR refactor -- but if you don't want to do it in this PR, I was going to go ahead and change that today anyway. Up to you
> > Not sure how necessary contextual components would be since most of the inputs are quite similar (the only difference being the pasForm field and the "type" of input). Unless we want to include the `label` in that contextual component like they did in the video? <-- but that would basically be the same amount of code space as before so don't think it's particularly beneficial
> 
> Contextual components would get us away from `@value={{get` stuff. Are you sure that using `get` as the @value there actually 2-way binds? The reason for that is 2-way binding would helps us move away from relying on key-down actions.
> 
> Contextual components would allow us to actually pass in explicit model attributes instead of doing these meta method.

Hmm. If we do go the contextual component route, we should do it in another PR.

allthesignals:
> Hmm. If we do go the contextual component route, we should do it in another PR.
The docs suggest value two-way binds still, but we'll confirm

Sorry, I don't think you're seeing my point.

I don't know if that will actually get us the two way binding. Does it know to set the value upstream? That's what I'm asking to confirm.

```
@value={{get @object @propertyName}}
```

godfreyyeung:
> > Hmm. If we do go the contextual component route, we should do it in another PR.
> > The docs suggest value two-way binds still, but we'll confirm
> 
> Sorry, I don't think you're seeing my point.
> 
> I don't know if that will actually get us the two way binding. Does it know to set the value upstream? That's what I'm asking to confirm.
> 
> ```
> @value={{get @object @propertyName}}
> ```

Yeah, I think so:
image
image

trbmcginnis:
I just committed some changes that use `@value={{mut (get @pasForm @landUseActionField)}}`. Seems to be working fine, but if you're still feeling weary about using this method ^^ I'd be happy to change these to contextual components.

client/app/components/land-use-action.js
```diff
@@ -67,31 +67,56 @@ export default class LandUseActionComponent extends Component {
...
   // (and therefore displays in the list of selected actions).
   @action
-  setSelectedActionsArray() {
-    this.landUseActionOptions.removeObject(this.selectedLandUseAction);
-    // sort options alphabetically
-    this.landUseActionOptions.sort((a, b) => ((a.name > b.name) ? 1 : -1));
+  addNewLandUseActionSelection() {
     // use unshiftObject to place new action at top of list
-    this.landUseActionSelections.unshiftObject(this.selectedLandUseAction);
+    const existsSelection = this.selectedLandUseAction.field;
```
allthesignals, line 102: Hmm `existsSelection`? Why not just `selection`? We'll know it exists if it passes the check below :)

allthesignals:
> I just committed some changes that use `@value={{mut (get @pasForm @landUseActionField)}}`. Seems to be working fine, but if you're still feeling weary about using this method ^^ I'd be happy to change these to contextual components.

No this is great! Thank you for confirming @godfreyyeung! :) I think this is the way we should go.

The intermediate value "newVal" should not be needed now.

client/app/components/land-use-action.js
godfreyyeung: True! 👌 Changed!

Issue #119, opened 2020-05-01
Pre-populate project name on PAS form


hannahkates:
Project name should be pre-populated. This form input is more complicated than we captured in the wireframes. It should be pre-populated with `dcp_projectname`. If the user makes a change to it in the form UI, we should save that revised value to `dcp_revisedprojectname`, and the form should then display `dcp_revisedprojectname` instead. 

So basically...
- If `dcp_revisedprojectname` IS NULL, display `dcp_projectname`
- If `dcp_revisedprojectname` IS NOT NULL, display `dcp_revisedprojectname`

![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/b16cb3c4-ef71-43a0-b699-d7bccd469dcf)

PR #190, opened 2020-05-06
119: pre-populate project name on pas-form with dcpProjectname

trbmcginnis:
BEFORE MERGE: might cause merge conflicts in tests

Use `package.project.dcpProjectname` as a default for `package.pasForm.dcpRevisedprojectname` on the pas-form model.

Needed to update tests to account for new reference to package.project.

Addresses #119

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -20,7 +20,7 @@
         <Input
           @type="text"
           @value={{@package.pasForm.dcpRevisedprojectname}}
-          data-test-dcprevisedprojectname
+          data-test-project-name-input
         />
       </label>
     </section>
--- a/client/app/models/pas-form.js
+++ b/client/app/models/pas-form.js
@@ -45,7 +45,12 @@ export default class PasFormModel extends Model {
 
   @attr('number') dcpProjectblock;
 
-  @attr('string') dcpRevisedprojectname;
+  @attr({
+    defaultValue(pasForm) {
+      return pasForm.package.project.dcpProjectname;
+    },
+  })
+  dcpRevisedprojectname;
 
   @attr('boolean') dcpExistingbuildingoccupied;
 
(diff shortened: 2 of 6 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -19,8 +19,9 @@
         <strong>Project Name</strong>
         <Input
           @type="text"
-          @value={{@package.pasForm.dcpRevisedprojectname}}
-          data-test-dcprevisedprojectname
+          @value={{if @package.pasForm.dcpRevisedprojectname @package.pasForm.dcpRevisedprojectname @package.project.dcpProjectname}}
+          @key-down={{fn (mut @package.pasForm.dcpRevisedprojectname)}}
```
allthesignals, line 23: I'm getting a little nervous about relying on these key-down events. How does this affect people who might right-click and paste data into the field? Wouldn't that not trigger a key-down event?

I think you should probably _bind_ the specific value here:

```
@value={{@package.pasForm.dcpRevisedprojectname}}
```

THEN in your pas-form model, add a [defaultValue](https://api.emberjs.com/ember-data/3.18/functions/@ember-data%2Fmodel/attr):

```
@attr({
  defaultValue(pasForm) {
    return pasForm.package.dcpProjectname;
  },
})
dcpRevisedprojectname;
```

See this [blog post](https://davidtang.io/2018/05/26/conditional-default-values-for-ember-data-model-attributes.html) for more examples.

It's unclear to me whether this defaultValue gets triggered when the server provides an empty string. Happy to jump on a call and work through that.

More broadly, though, my reasoning here is that this behavior is pretty fundamental to the business logic itself. Instead of keeping that critical behavior inside the if helper in a template, it's more apparent on the model level.

trbmcginnis: that's a good point, will change

allthesignals:
@trbmcginnis make sure you manually test that it works locally :) `yarn run test` from root.

Issue #110, opened 2020-04-29
Add address information next to bbl title in ProjectGeography component BBL list


trbmcginnis:
This issue is a break-off of a previous issue #21 <-- look here for extra info. In `project-geography.js` there is already some logic to get this started, comments include the associated issue. 

#### User Story
As an applicant, I want to see addresses next to the BBLs in my bbl list so that I can more easily understand my project geography

#### Description
Right now when a user submits a bbl for their project on the ProjectGeography component on the PAS Form package, a list of their project BBLs will display. There is no address attribute stored with BBLs in CRM so we do not have easy access to address information. We do have access to addresses through geosearch, though, so in order to display this info we need to query geosearch for addresses based on bbls. This can happen in the frontend or the backend. This extra querying will only be necessary for bbls that already exist in the db, because when a user submits a new bbl they will do so through the labs-search component, which provides address information on the spot -- and we can display this information for them right away. 

![image.png](https://images.zenhubusercontent.com/5ca36b1555cfe32f2f66b411/4cd7dafc-11f0-4bf7-a8b4-8d68b82f1c08)

hannahkates:
- Display address when someone adds the BBL. It's okay if the address disappears upon hard refresh.
- Turn BBL into hyperlink to ZoLa profile for that lot

PR #193, opened 2020-05-06
110: addresses next to bbl in ProjectGeography component bbl list

trbmcginnis:
**BBls list in ProjectGeography component now includes link to zola and address.** 

**Address**: When a user searches a bbl or address in the labs-search component, an object with a bbl and corresponding address is returned. A lookup is populated with these bbl-address pairs so that addresses can be displayed next to the bbl in the bbls list. NOTE: these will disappear on hard refresh. 

**Zola Link**: This PR also provides link to zola on each bbl.

Addresses #110 

Screen Shot 2020-05-06 at 5 01 29 PM

### DIFF
--- a/client/app/components/project-geography.hbs
+++ b/client/app/components/project-geography.hbs
@@ -10,7 +10,16 @@
 {{#each @bbls as |bbl idx|}}
   <fieldset class="fieldset relative">
     <legend data-test-bbl-title="{{bbl.dcpBblnumber}}">
-      <strong>{{bbl.dcpBblnumber}}</strong>
+      <a
+        href="https://zola.planning.nyc.gov/bbl/{{bbl.dcpBblnumber}}"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <strong>{{bbl.dcpBblnumber}}</strong>
+      </a>
+      {{#if bbl.temporaryAddressLabel}}
+        ({{bbl.temporaryAddressLabel}})
+      {{/if}}
     </legend>
     <fieldset>
       <legend data-test-development-site-question="{{bbl.dcpBblnumber}}-{{bbl.dcpDevelopmentsite}}">
--- a/client/app/components/project-geography.js
+++ b/client/app/components/project-geography.js
@@ -10,27 +10,23 @@ export default class ProjectGeographyComponent extends Component {
   // represents array of bbl models
   @tracked bbls;
 
-  // TODO: ISSUE #110
-  // Used to find corresponding addresses for list of bbls.
-  // Populated later with labs-search addon response.
-  // bblToAddressLookup = {};
-
   @action
   selectSearchResult(labsSearchResult) {
     // `labsSearchResult` is object with `label` (address) and `bbl` attributes.
     // labsSearchResult.bbl comes back as a number,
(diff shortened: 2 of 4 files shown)


client/app/components/project-geography.hbs
```diff
@@ -10,7 +10,16 @@
 {{#each @bbls as |bbl idx|}}
   <fieldset class="fieldset relative">
     <legend data-test-bbl-title="{{bbl.dcpBblnumber}}">
-      <strong>{{bbl.dcpBblnumber}}</strong>
+      <a
+        href="https://zola.planning.nyc.gov/bbl/{{bbl.dcpBblnumber}}"
+        target="_blank"
+        rel="noopener noreferrer"
```
allthesignals, line 16: Out of curiosity what does this do?

client/app/components/project-geography.hbs
```diff
@@ -10,7 +10,16 @@
 {{#each @bbls as |bbl idx|}}
   <fieldset class="fieldset relative">
     <legend data-test-bbl-title="{{bbl.dcpBblnumber}}">
-      <strong>{{bbl.dcpBblnumber}}</strong>
+      <a
+        href="https://zola.planning.nyc.gov/bbl/{{bbl.dcpBblnumber}}"
+        target="_blank"
```
allthesignals, line 15: Awesome!

client/app/components/project-geography.js
```diff
@@ -10,27 +10,27 @@ export default class ProjectGeographyComponent extends Component {
...
 
-    // TODO: ISSUE #110
-    // set local variables for displaying address next to bbl in list
-    // const currentAddress = labsSearchResult.label;
-    // const formattedAddress = currentAddress.replace(', New York, NY, USA', '');
-    // this.bblToAddressLookup[currentBbl] = formattedAddress;
+    // set local variables for displaying address next to bbl in bbls list
+    const currentAddress = labsSearchResult.label;
+    const formattedAddress = currentAddress.replace(', New York, NY, USA', '');
+    this.bblToAddressLookup[currentBbl] = formattedAddress;
```
allthesignals, line 33: I see! So you're storing all the address references in the a hash on the component. 

What if you stored them as a regular property on the model itself? For example, in pas-form model:

```
class PasForm extends Model {
  // other stuff...

  temporaryAddressLabel = '';
}
```

It won't have an `@attr` decorator, meaning it won't get serialized and sent to the server. It's just a property on the object.

Then, in your template, you can just reference it:

```handlebars
      {{#if bbl.temporaryAddressLabel}}
        ({{bbl.temporaryAddressLabel}})
      {{/if}}
```

Then, all you'd have to do is change this code to set it:
```javascript
  bblObjectToPush.temporaryAddressLabel = formattedAddress;
```

Thoughts? I think it's probably cleaner just because we aren't having to have the component maintain a stateful object lookup.

client/app/components/project-geography.hbs
trbmcginnis: it's a linting thing -- need to add `rel="noopener noreferrer` to get linting to pass

client/app/components/project-geography.js
trbmcginnis: ooh yes great idea, I'll update

PR #196, opened 2020-05-07
178 - display user's email address in header

andycochran:
This PR refactors the "Logout" (sic) button so that the authenticated user's email address displays next to the "Sign Out" (Yay!) button. 

image


client/app/controllers/application.js
```diff
@@ -5,6 +5,8 @@ import { inject as service } from '@ember/service';
 export default class ApplicationController extends Controller {
   @service session;
 
+  userEmail = this.session.data.authenticated.emailaddress1;
```
allthesignals, line 8: Please just access this.session.data.authenticated.emailaddress1 directly from the template.

Ideally the logic for the current user should be wrapped in its own component, with the session service injected. I don't think using a controller is getting us anything aside from having to look up where this value might be coming from.

bfreeds: Good call @allthesignals, that makes sense.

Issue #162, opened 2020-05-05
Applicant team members aren't displaying in CRM


hannahkates:
## What should happen:
When the user saves the form, the applicant team info they saved should get sent to Dynamics CRM.

## What happened instead:
The tables on Dynamics CRM PAS form page doesn't get filled in, even though the data has been successfully sent to CRM and can be retrieved by the front-end app.
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/deb546ba-cec4-47f9-8d54-c3ce19ce97ec)

## How to reproduce this bug:

1. Edit a PAS form by adding applicant team data
2. Save the form changes
3. View that PAS form in Dynamics CRM and see that the applicant tables are both blank.

allthesignals:
Here's some validation information we received from @Eabrante 

| field | Label | Required | Type | Validation | Default | Notes |
|------------------|--------------|----------|------------|------------------------------------------------------|---------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| dcp_organization | Organization | Yes | Text | Length: 100 | Blank | If 'Type' field value is selected as “Authorized Applicant Representative” then show below field on the form else hide it from form: Organization |
| dcp_name | Full Name | Yes | Read-only |  |  |  |
| dcp_firstname | First Name | Yes | Text | Length: 50 | Blank |  |
| dcp_lastname | Last Name | Yes | Text | Length: 50 | Blank |  |
| dcp_type | Type | Yes | Option Set | 1. Applicant  2. Authorized Applicant Representative | Null | From MCS Portal - Get all Applicant Information associated to current PAS Form where Type Optionset value = Applicant or Authorized Applicant Representative and Applicant Information.status = Active |
| emailaddress1 | Email | Yes | Text |  | Blank |  |
| dcp_address | Address | Yes | Text | Length: 250 | Blank |  |
| dcp_city | City | Yes | Text | Length: 80 | Blank |  |
| territorycode | State | No | Option Set | Value: US States | Null |  |
| dcp_zipcode | Zip Code | No | Numeric | Length: 5 | Blank |  |
| telephone1 | Phone | No | Numeric | Length: 10 | Blank |  |

![image.png](https://images.zenhubusercontent.com/5ca23922d9fcf5278d715977/5cdb99e6-1f13-4033-90b9-009a476a67d4)

bfreeds:
[...]

PR #201, opened 2020-05-08
162:  Refactor applicants model

bfreeds:
This PR refactors the applicants model to prep for server logic to translate frontend applicant model to the appropriate entity in CRM.  Prep work for #162.  Want to merge this as a small PR to make sure everyone has these changes in their feature branches.

The changes:
1. Rename `applicantType` to `dcpType`, to make the model consistently follow CRM naming
2. Use explicit entityType values to make it clear which entity/table in CRM they will go to.  This will make it simpler on the server side implementation.  Because this is using the explicit crm entity values, I created a computed that translates those entity names into friendly names which we use in the UI (to avoid complicated/repetitive conditionals in the template).

### DIFF
--- a/client/app/components/packages/applicant-fieldset.hbs
+++ b/client/app/components/packages/applicant-fieldset.hbs
@@ -8,24 +8,20 @@
   {{!-- if sent to CRM "Applicant Information" table, need to know "type" --}}
   {{#if (eq @applicant.friendlyEntityName "Applicant")}}
     <div class="cell medium-6" data-test-applicant-type-radio-group>
-      <Input
-        @type="radio"
-        name={{concat 'applicant-type' @elementId}}
-        @value="Applicant"
-        checked={{if (eq @applicant.dcpType "Applicant") true}}
-        onClick={{fn this.updateAttr @applicant "dcpType" "Applicant"}}
-        data-test-applicant-type-radio-individual
-      />
-      <label>Individual</label>
-      <Input
-        @type="radio"
-        name={{concat 'applicant-type' @elementId}}
-        @value="Authorized Applicant Representative"
-        checked={{if (eq @applicant.dcpType "Authorized Applicant Representative") true}}
-        onClick={{fn this.updateAttr @applicant "dcpType" "Authorized Applicant Representative"}}
-        data-test-applicant-type-radio-organization
-      />
-      <label>Organization</label>
+      <RadioButton
+        @value={{717170000}}
+        @groupValue={{@applicant.dcpType}}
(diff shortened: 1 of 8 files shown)


bfreeds:
@allthesignals the code cov is still failing on this, I think this is a case that we should make an exeption and allow this to merge despite the code cov failing -- I'm not sure how to appropriately test that one line that is not being covered, as there is nothing in the UI that would actually allow a user to hit hit the error that would be raised.  Since the line that is not being covered is defensive coding, anyways, it feels like this is a case that can be not covered by the tests.

client/app/models/applicant.js
```diff
@@ -4,15 +4,10 @@ export default class ApplicantModel extends Model {
...
-    defaultValue: 'Applicant',
-  })
-  targetEntity;
-
-  // doesn't exist on "Applicant Representative Information" CRM table
+  // only exists on dcp_applicantinformation CRM entity
+  // either 'Applicant' or 'Authorized Applicant Representative'
   @attr('string')
-  applicantType;
+  dcpType;
```
allthesignals, line 10: This should be a number

client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -1,36 +1,36 @@
...
     <div class="cell medium-6" data-test-applicant-type-radio-group>
       <Input
         @type="radio"
         name={{concat 'applicant-type' @elementId}}
-        @value="Individual"
-        checked={{if (eq @applicant.applicantType "Individual") true}}
-        onClick={{fn this.updateAttr @applicant "applicantType" "Individual"}}
+        @value="Applicant"
+        checked={{if (eq @applicant.dcpType "Applicant") true}}
+        onClick={{fn this.updateAttr @applicant "dcpType" "Applicant"}}
```
allthesignals, line 16: Mind if I refactor this to use <RadioButton/> ?

allthesignals:
@bfreeds going to merge this but tomorrow would like to talk through some ideas I have for refactoring a bit more to make more aligned with the model in CRM.

client/app/components/packages/applicant-team-editor.js
```diff
@@ -30,7 +30,7 @@ export default class ApplicantTeamEditorComponent extends Component {
   @action
   addApplicantIfEmpty() {
     if (!this.args.applicants.length) {
-      this.addApplicant('Applicant');
+      this.addApplicant();
```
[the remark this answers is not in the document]
bfreeds: It's the default of the applicant model, so passing it as an argument felt redundant.

Issue #27, opened 2020-03-13
PAS "Submit"


hannahkates:
Submission is only possible after saving. Clicking the "Submit" button on the PAS page should post to CRM all the data filled in to the PAS package form. Clicking this button should trigger the transition to the confirmation age.

## Status to update:
The PAS package form status should update to "Submitted"

## Attributes to send:
**[See wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP---Applicant-Portal?node-id=0%3A1)**

NOTE: Submitting some attributes to the PAS Package SHOULD trigger an automatic update on the dcp_project record as well as on the package. We should check this once we build the functionality.

**dcp_project fields that should get auto-updated by CRM**
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/425d7415-cbed-4f7f-966d-248716b45d10)

bean75:
Labs should not have to update the project record - just adding to the form should cause MCS to automatically update the project record. Liz could confirm but I think we picked those field s to remain because they do auto populate already.

hannahkates:
@Eabrante do you know if that's true, that submitting the PAS auto sends data to the project record? That would be great if we don't need to trigger it separately. Testing may be the only way to confirm 100% that changing the PAS status to "Submitted" actually triggers this on its own.

bean75:
@hannahkates I would assume yes. We had never planned for additional data mapping that wasn't already developed in CRM.

trbmcginnis:
@hannahkates From looking at the metadata spreadsheet I'm assuming `statuscode` is the field to update with "Submitted"?

This field has three options"

```"States:
1: Active
2: Inactive
717170000: Submitted"```

Eabrante:
> @hannahkates From looking at the metadata spreadsheet I'm assuming `statuscode` is the field to update with "Submitted"?
> 
> This field has three options:
> 
> ```
> 1: Active
> 2: Inactive
> 717170000: Submitted"
> ```

Hi @trbmcginnis , the scheme field might be 'statecode' for the package status.

Eabrante:
@trbmcginnis , nope you are right. Its 'statuscode'. 

![image](https://user-images.githubusercontent.com/43378798/81213471-64657280-8fa4-11ea-9b89-4cb8c25061e7.png)

trbmcginnis:
[...]

PR #202, opened 2020-05-08
27: package statuscode is changed to Submitted when user clicks Submit button

trbmcginnis:
package `statuscode` is set to "Submitted" when a user clicks the Submit button on the PAS Form.

Partly addresses #27

In future PRs:
- conditions for disabling Submit button

### DIFF
--- a/client/app/components/packages/pas-form/edit.js
+++ b/client/app/components/packages/pas-form/edit.js
@@ -27,7 +27,8 @@ export default class PasFormComponent extends Component {
 
   @action
   async submit(projectPackage) {
-    await projectPackage.saveDescendants();
+    projectPackage.statuscode = 'Submitted';
+    await projectPackage.save();
 
     this.router.transitionTo('packages.show', projectPackage.id);
   }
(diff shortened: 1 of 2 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -826,7 +826,7 @@
                 <button
                   type="button"
                   class="button large no-margin"
-                  {{on "click" (fn this.submit @package)}}
+                  onClick={{fn this.submit @package}}
```
allthesignals, line 829: What's wrong with the modifier?

trbmcginnis: Oh nothing's wrong, was just updating for consistency -- I can change back if you prefer the other way

allthesignals: Hmm consistent with what? The click handling above uses modifiers.

I'm not sure what's best but onClick/onclick has confused me because onclick is a native event callback. onClick seems to work too, and I don't know why, because it's camel-case. 

I think the modifier is maybe more explicit, it's clearly part of Ember and not ambiguously 'native html or some older maybe-deprecated ember api'... what do you think?

PR #204, opened 2020-05-08
Instruction improvements

hannahkates:
Need to test some conditional logic locally before this is ready for review


allthesignals:
Hmm can you pick off your commit and make a new branch off develop?

PR #205, opened 2020-05-09
163 debug missing radio btn answers

godfreyyeung:
This PR builds off #200  , so inspect the last commit for this PR's changes.

Restores radio button answers by making sure templates are matching
number properties to numbers, not strings.

Also changes input checked attribute values to "checked" instead of
true. This conforms with spec.

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

Also removes @ from input name attributes, since they are not Ember
attributes.


Hemachekuri:
working fine, closed 199 with Screen shots

Issue #174, opened 2020-05-06
Implement front-end validations


hannahkates:
On save:
- [ ] Field character limits (see wireframes)

On submit:
- [ ] Need at least one file to upload
- [ ] Conditionally required fields
- [ ] Need to have at least one applicant and the "type" of that applicant

trbmcginnis:
Connecting to the "Submit" issue so we can make sure we know which button (**Save** vs **Submit**) has which validations.

https://app.zenhub.com/workspaces/labs-sprint-5e56ebd113bedb2575e694da/issues/nycplanning/labs-applicant-portal/27

hannahkates:
@godfreyyeung this is also related to #162. The user needs to specify the type of applicant they add.

PR #206, opened 2020-05-10
Validate "saveable" and "submittable" forms

allthesignals:
Addresses #174 

Creates two sets of validations: "saveable state" and "submittable state". These are organized under the `validations` folder.

Saveable validations impose a lower requirement than submittable because they are the bare minimum required to successfully save to the server without error.

Submittable imposes higher requirements because they are the validations required for successful submission.

Based on the requirements, I've added a custom validation based on `validatePresence`. This validation only conditionally invokes validatePresence based on the value of a specified field. This is saved under the validators folder.

This PR also refactors some of the edit form to be more manageable by generating groups of radio buttons through structured data. The refactor uses ember-radio-button.

A small fix was added which throws a proper error when a package is not found.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -94,66 +94,42 @@
             <legend>
               <strong>Does the proposed Project, or portion thereof, require a NYC DEP Storm Water Construction Permit?</strong>
             </legend>
-            <Input
-              @type="radio"
-              @name="dcpproposedprojectorportionconstruction"
-              id="dcpproposedprojectorportionconstruction-yes"
-              @value='717170000'
-              checked={{if (eq this.unifiedChanges.dcpProposedprojectorportionconstruction "717170000") true}}
-              onClick={{fn this.updateAttr this.unifiedChanges "dcpProposedprojectorportionconstruction" "717170000"}}
-              data-test-dcpproposedprojectorportionconstruction-yes
-            /><label for="dcpproposedprojectorportionconstruction-yes">Yes</label>
-            <Input
-              @type="radio"
-              @name="dcpproposedprojectorportionconstruction"
-              id="dcpproposedprojectorportionconstruction-no"
-              @value='717170000'
-              checked={{if (eq this.unifiedChanges.dcpProposedprojectorportionconstruction "717170001") true}}
-              onClick={{fn this.updateAttr this.unifiedChanges "dcpProposedprojectorportionconstruction" "717170001"}}
-              data-test-dcpproposedprojectorportionconstruction-no
-            /><label for="dcpproposedprojectorportionconstruction-no">No</label>
-            <Input
-              @type="radio"
-              @name="dcpproposedprojectorportionconstruction"
-              id="dcpproposedprojectorportionconstruction-unsure"
-              @value='717170002'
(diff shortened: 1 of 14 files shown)


client/app/validations/saveable-pas-form.js
```diff
@@ -0,0 +1,104 @@
...
+    validateLength({
+      max: 50,
+      message: 'Name must be between {min} and {max} characters',
+    }),
+  ],
+
+  dcpDescriptionofprojectareageography: [
+    validateLength({
+      max: 2000,
+      message: 'Must be between {min} and {max} characters',
```
hannahkates, line 17: Since a min isn't specified, does it default to plugging in zero instead?

hannahkates:
Thank you @allthesignals!!! Does this fix #27 too?

client/app/validators/required-if-selected.js
```diff
@@ -0,0 +1,30 @@
+import {
+  validatePresence,
+} from 'ember-changeset-validations/validators';
+
+export default function validatePresenceIf(options) {
```
bfreeds, line 5: @allthesignals can you explain that this function is doing?

client/app/validations/submittable-pas-form.js
```diff
@@ -0,0 +1,58 @@
+import SaveablePasForm from './saveable-pas-form';
```
trbmcginnis, line 1: These validations also need to include the extra questions for the land use actions. E.g. if `dcpPfzoningauthorization > 0` then `dcpZoningauthorizationpursuantto` and `dcpZoningauthorizationtomodify` are required. That goes for all land use actions that have extra questions (listed out in `land-use-actions.js`

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -94,66 +94,42 @@
...
-            <Input
-              @type="radio"
-              @name="dcpproposedprojectorportionconstruction"
-              id="dcpproposedprojectorportionconstruction-unsure"
-              @value='717170002'
-              checked={{if (eq this.unifiedChanges.dcpProposedprojectorportionconstruction "717170002") true}}
-              onClick={{fn this.updateAttr this.unifiedChanges "dcpProposedprojectorportionconstruction" "717170002"}}
-              data-test-dcpproposedprojectorportionconstruction-unsure
-            /><label for="dcpproposedprojectorportionconstruction-unsure">Unsure at this time</label>
+            {{#each (array
```
bfreeds, line 100: Love these refactors to simplify the form.

client/app/validators/required-if-selected.js
allthesignals: Sorry, yes — it conditionally validates the presence of a field. For example, in the form, if I've selected "yes" to one question, and it asks a follow-up question, this validation will make that follow-up question _required_ but only if "yes" is selected.

client/app/validations/submittable-pas-form.js
allthesignals: Gotcha! Will need help on that...

client/app/validators/required-if-selected.js
bfreeds: Gotchya, thank you -- that's helpful and makes sense

client/app/validations/saveable-pas-form.js
allthesignals: Hmm... I think this is referring to character length? It'll just be a blank string.

PR #207, opened 2020-05-10
Dirty state touchups

allthesignals:
Depends on #206 

Refactors the approach to multiple changes with a much simpler approach that doesn't rely on proxies, instead uses events from e-changeset. Makes conditional validations much stabler.

Refactors several template logic to simplify.

Fixes a small bug in project geographies form in which the radio groups weren't configured correctly and so changes in unrelated radio buttons would change

Authenticates routes so that an invalidated cookie will redirect users.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -1,3 +1,5 @@
+{{log this.submittableChanges.error}}
+
 <div class="grid-x grid-margin-x">
 
     <div class="cell large-8">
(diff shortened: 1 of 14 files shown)


client/app/components/project-geography.hbs
```diff
@@ -25,59 +25,44 @@
...
-        name={{concat idx '-bbl-dev-site'}}
-        id={{concat idx '-bbl-dev-site-no'}}
-        @value={{eq bbl.dcpDevelopmentsite false}}
-        checked={{eq bbl.dcpDevelopmentsite false}}
-        {{on "click" (fn this.updateAttr bbl "dcpDevelopmentsite" false)}}
-        data-test-bbl-development-site-no="{{bbl.dcpBblnumber}}"
-      />
-      <label for="{{concat idx '-bbl-dev-site-no'}}">
+      </RadioButton>
+      <RadioButton
```
bfreeds, line 36: Using these RadioButton components is a lot cleaner.

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -1,3 +1,8 @@
+{{!-- template-lint-disable no-log --}}
+{{#each-in this.submittableChanges.error as |key value|}}
+  {{log key value.validation value.value}}
+{{/each-in}}
```
allthesignals, line 4: Left this in for future debugging purposes

PR #209, opened 2020-05-11
RadioButton refactor

andycochran:
This PR refactors the `<RadioButton>` component to follow the style guide markup standards (labels following inputs instead of wrapping them). It also passes in the optional `@radioId` arg for the `{{radio-button-input}}` component to use — which associates the labels with their inputs.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -11,19 +11,19 @@
         submitting the PAS, a prospective applicant acknowledges that they intend to file a land use application with NYC
         Planning. Submission allows staff to be assigned from appropriate divisions and project tracking to commence.
       </p>
-    
+
       <p>
         NYC Planning understands that some projects may not be fully formed at the submission of the PAS. Give as much
         detailed information as possible. Failure to provide detailed, relevant information will result in the PAS being
         rejected.
       </p>
-    
+
       <section class="form-section">
         <h3 class="section-header">
           <span id="project-info" class="section-anchor"></span>
           Project Information
         </h3>
-    
+
         <label>
           <strong>Project Name</strong>
           <Input
@@ -32,7 +32,7 @@
           />
         </label>
       </section>
-    
+
       <section class="form-section">
         <h3 class="section-header">
           <span id="applicant-team" class="section-anchor"></span>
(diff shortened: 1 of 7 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -110,7 +111,8 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpProposedprojectorportionconstruction}}
-                 data-test-dcpproposedprojectorportionconstruction={{radio.label}}
+                @radioId={{concat 'dcpproposedprojectorportionconstruction' radio.code}}
```
allthesignals, line 114: Hmm why are these necessary?

andycochran: Per the [ember-radio-button docs](https://github.com/yapplabs/ember-radio-button/blob/master/addon/components/radio-button.js#L13-L20), `radioId` is an optional arg. If you don't pass in a `radioId`, the input default to an ember id, the label's `for=""` attribut is empty, and there's no association between the label and the input.

andycochran: More specifically, labels worked ootb w/o passing this because they wrapped the inputs. But with this change in the markup (labels following inputs), we need the explicit association of a `for=""`. I suppose we could do this dynamically with the auto-generated ember element id, if you'd rather not pass explicit id args.

allthesignals: Aah gotcha, thanks! So making the label a sibling meant the inferred `for` wasn't working.

I would much rather use the auto-generated ember element ID. This way the identification is full encapsulated by the component, we don't have to manage identifiers just for radio. It's also just less template code to worry about.

andycochran: Actually, glimmer components are tagless, so there's no `elementId`. We could generate one in `radio-button.js`, but IDK if that's better than passing one explicitly in the hbs.

allthesignals: In that embermap video on forms, it talks about how to generate a guid for your component. You would import `guid` from from Ember core.

We don't need to explicitly pass an identifier to it because it doesn't even get us anything. "Explicit" is nice, but not when it adds to the maintenance overhead with no real benefit.

Lemme know if you need help!

allthesignals: I would also no that this radio button component isn't a glimmer component, it's still a regular ember component. So I think it does have an elementId.

allthesignals: Ah, it may just be that it's tagless because they set `tagName: ''`. So the generated guid path would be the way to go.

PR #213, opened 2020-05-11
212 pas bbl validation

allthesignals:
This PR adds behavior that fills out the rest of the ProjectBBL, CRM-side.

It adds lookups for mapping a borough name to a CRM-code from zap search.

needs input from @trbmcginnis

### DIFF
--- a/client/app/models/bbl.js
+++ b/client/app/models/bbl.js
@@ -1,5 +1,32 @@
 import Model, { attr, belongsTo } from '@ember-data/model';
 
+export const PROJECT_BBL_BOROUGHS = {
+  BRONX: {
+    code: 717170000,
+    label: 'Bronx',
+  },
+  BROOKLYN: {
+    code: 717170002,
+    label: 'Brooklyn',
+  },
+  MANHATTAN: {
+    code: 717170001,
+    label: 'Manhattan',
+  },
+  QUEENS: {
+    code: 717170003,
+    label: 'Queens',
+  },
+  STATEN_ISLAND: {
+    code: 717170004,
+    label: 'Staten Island',
+  },
+  CITYWIDE: {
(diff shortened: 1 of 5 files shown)


server/src/packages/pas-form/bbls/bbls.controller.ts
```diff
@@ -9,6 +9,9 @@ export const BBL_ATTRIBUTES = [
   'dcp_partiallot',
   'dcp_developmentsite',
   'dcp_bblnumber',
+  'dcp_userinputborough',
```
trbmcginnis, line 12: If these user input attributes don't exist on the database, do they just exist as local variables in the backend? When the bbl entity table gets patched, do these values just disappear since they're not saved to the db?

client/app/models/bbl.js
```diff
@@ -1,5 +1,32 @@
 import Model, { attr, belongsTo } from '@ember-data/model';
 
+export const PROJECT_BBL_BOROUGHS = {
```
trbmcginnis, line 3: It's not super clear _why_ it's necessary that you're mapping these codes -- are you sending these codes to a CRM field, e.g. `dcp_borough`? Would love a comment here explaining why it's necessary

server/src/packages/pas-form/bbls/bbls.controller.ts
allthesignals: @trbmcginnis these do exist as real attrs in CRM

the front end is responsible for filling out these particular fields

client/app/models/bbl.js
allthesignals: will do! it's necessary because zap search provides bbls as 1-5-4 numerics, and we need to map the first numeric to a particular CRM code.

PR #224, opened 2020-05-12
175 Display spinners during save or submit

godfreyyeung:
A spinner is displayed on respective buttons during save or submit.
This commit also syncs the button disabled state and spinner state.

It also removes a duplicate call to pasForm.save() by calling pasForm changesets' execute() -- instead of save() --  before calling package.saveDescendents(), which will call pasForm.save() anyway. 

changeset.execute() moves the Changeset changes down to the underlying model.

For next PR: 
 - disable form inputs when saving or submitting

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -698,11 +698,15 @@
             type="button"
             class="button expanded large"
             disabled={{not this.isSaveable}}
-            {{on "click" this.save}}
+            {{on "click" (perform this.save)}}
             data-test-save-button
           >
             {{!-- TODO: Make this button work with the action passed in @save--}}
-            Save
+            {{#if this.save.isRunning}}
+              <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
+            {{else}}
+              Save
+            {{/if}}
           </button>
           <button
             type="button"
@@ -711,8 +715,11 @@
             {{on "click" this.toggleModal}}
             data-test-submit-button
           >
-            {{!-- TODO: Disable this button if model dirty/invaild --}}
-            Submit
+            {{#if this.submit.isRunning}}
+              <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
+            {{else}}
+              Submit
+            {{/if}}
           </button>
         </p>
         {{#if this.modalIsOpen}}
@@ -727,6 +734,7 @@
                   <button
                     type="button"
                     class="button large secondary no-margin"
+                    disabled={{this.submit.isRunning}}
                     {{on "click" this.toggleModal}}
                   >
                     Continue Editing
@@ -735,11 +743,16 @@
                 <div class="cell large-6 text-right">
                   <button
                     type="button"
-                    class="button large no-margin"
-                    {{on "click" (fn this.submit)}}
+                    class="button expanded large no-margin"
+                    disabled={{this.submit.isRunning}}
+                    {{on "click" (perform this.submit)}}
                     data-test-confirm-submit-button
                   >
-                    Submit
+                    {{#if this.submit.isRunning}}
+                      <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
+                    {{else}}
+                      Submit
+                    {{/if}}
                   </button>
                 </div>
               </div>
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -130,7 +130,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpUrbanrenewalarea}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                  data-test-dcpurbanrenewalarea={{radio.label}}
               >
                 {{radio.label}}
@@ -195,7 +195,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpLanduseactiontype2}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                 data-test-dcplanduseactiontype2={{radio.label}}
               >
                 {{radio.label}}
@@ -229,7 +229,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpProjectareaindustrialbusinesszone}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                 data-test-dcpprojectareaindustrialbusinesszone={{radio.label}}
               >
                 {{radio.label}}
@@ -274,7 +274,7 @@
                 <Input
                   @type="text"
                   @value={{this.saveableChanges.dcpProjectarealandmarkname}}
-                  @changed={{fn this.validate}}
+                  @changed={{this.validate}}
                   maxlength="250"
                   data-test-dcpisprojectarealandmarkname
                 />
@@ -333,7 +333,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpRestrictivedeclaration}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                 data-test-dcprestrictivedeclaration={{radio.label}}
               >
                 {{radio.label}}
@@ -479,7 +479,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpIsinclusionaryhousingdesignatedarea}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                 data-test-dcpisinclusionaryhousingdesignatedarea={{radio.label}}
               >
                 {{radio.label}}
@@ -512,7 +512,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpDiscressionaryfundingforffordablehousing}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
                 data-test-dcpdiscressionaryfundingforffordablehousing={{radio.label}}
               >
                 {{radio.label}}
--- a/client/app/components/packages/pas-form/edit.js
+++ b/client/app/components/packages/pas-form/edit.js
@@ -3,6 +3,7 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { Changeset } from 'ember-changeset';
 import lookupValidator from 'ember-changeset-validations';
+import { task } from 'ember-concurrency';
 import SaveablePasFormValidations from '../../../validations/saveable-pas-form';
 import SubmittablePasFormValidations from '../../../validations/submittable-pas-form';
 
@@ -57,26 +58,32 @@ export default class PasFormComponent extends Component {
       isApplicantsDirty,
     } = this.pasForm;
 
-    return (isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty;
+    return !this.isUpdatingDb && ((isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty);
   }
 
   get isSubmittable() {
-    return this.submittableChanges.isValid;
+    return !this.isUpdatingDb && this.submittableChanges.isValid;
   }
 
-  @action
-  async save() {
-    await this.saveableChanges.save();
-
-    this.args.onSave();
+  get isUpdatingDb() {
+    return this.save.isRunning || this.submit.isRunning;
   }
 
-  @action
-  async submit() {
-    await this.submittableChanges.save();
+  @task(function* () {
+    yield this.saveableChanges.execute();
+    yield this.saveableChanges.rollback();
 
-    this.args.onSubmit();
-  }
+    yield this.args.onSave();
+  })
+  save;
+
+  @task(function* () {
+    yield this.submittableChanges.execute();
+    yield this.submittableChanges.rollback();
+
+    yield this.args.onSubmit();
+  })
+  submit;
 
   @action
   validate() {
@@ -87,6 +94,8 @@ export default class PasFormComponent extends Component {
 
   @action
   toggleModal() {
-    this.modalIsOpen = !this.modalIsOpen;
+    if (!this.save.isRunning) {
+      this.modalIsOpen = !this.modalIsOpen;
+    }
   }
 }
--- a/client/app/controllers/packages/edit.js
+++ b/client/app/controllers/packages/edit.js
@@ -3,13 +3,13 @@ import { action } from '@ember/object';
 
 export default class PackagesEditController extends Controller {
   @action
-  savePackage() {
-    this.model.saveDescendants();
+  async savePackage() {
+    await this.model.saveDescendants();
   }
 
   @action
-  submitPackage() {
-    this.model.submit();
+  async submitPackage() {
+    await this.model.submit();
 
     this.transitionToRoute('packages.show', this.model.id);
   }
--- a/client/tests/acceptance/user-can-click-package-edit-test.js
+++ b/client/tests/acceptance/user-can-click-package-edit-test.js
@@ -4,6 +4,7 @@ import {
   click,
   currentURL,
   settled,
+  waitFor,
 } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import { setupMirage } from 'ember-cli-mirage/test-support';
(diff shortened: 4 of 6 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -130,7 +130,7 @@
               <RadioButton
                 @value={{radio.code}}
                 @groupValue={{this.saveableChanges.dcpUrbanrenewalarea}}
-                @changed={{fn this.validate}}
+                @changed={{this.validate}}
```
allthesignals, line 133: Ah, so you're removing this because `fn` is only for when you need to apply other arguments to the function call? Plz explain. Thx!

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -697,22 +697,29 @@
           <button
             type="button"
             class="button expanded large"
-            disabled={{not this.isSaveable}}
-            {{on "click" this.save}}
+            disabled={{or (not this.isSaveable) this.isSavingOrSubmitting}}
+            {{on "click" (perform this.save)}}
```
allthesignals, line 701: good

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -697,22 +697,29 @@
           <button
             type="button"
             class="button expanded large"
-            disabled={{not this.isSaveable}}
-            {{on "click" this.save}}
+            disabled={{or (not this.isSaveable) this.isSavingOrSubmitting}}
```
allthesignals, line 700: I think the truth helper here makes this a little confusing and spread out because we already have a lot of conditional logic for determining the saveable/submittable button state  inside those getters.

Can we build this logic into `isSaveable`? Or another getter?

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -697,22 +697,29 @@
...
+              <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
+            {{else}}
+              Save
+            {{/if}}
           </button>
           <button
             type="button"
             class="button expanded large secondary"
-            disabled={{not this.isSubmittable}}
+            disabled={{or (not this.isSubmittable) this.isSavingOrSubmitting}}
```
allthesignals, line 714: Do you think this would be better up in the component class just so that we're not keeping button disabled logic in two places ( the template and the class file)?

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -727,6 +734,7 @@
                   <button
                     type="button"
                     class="button large secondary no-margin"
+                    disabled={{this.submit.isRunning}}
```
allthesignals, line 737: Is this so that the user doesn't click out of the modal after they've said they want to submit _for sure_?

client/app/components/packages/pas-form/edit.js
```diff
@@ -64,19 +68,29 @@ export default class PasFormComponent extends Component {
...
+  save;
+
+  @task(function* () {
+    yield this.submittableChanges.execute();
+
+    yield this.args.onSubmit();
+
+    yield this.submittableChanges.rollback();
+
+    this.router.transitionTo('packages.show', this.package.id);
```
allthesignals, line 91: I thought we refactored this out in another PR? Why go backwards on that progress?

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -735,11 +743,16 @@
...
+                    {{on "click" (perform this.submit)}}
                     data-test-confirm-submit-button
                   >
-                    Submit
+                    {{#if this.submit.isRunning}}
+                      <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
+                    {{else}}
+                      Submit
+                    {{/if}}
```
allthesignals, line 755: Looks really good. I worry about this modal, I think it could be moved out to its own component.

client/app/components/packages/pas-form/edit.js
allthesignals: If it's because things aren't being awaited, you probably need to make those controller methods async.

client/app/components/packages/pas-form/edit.js
```diff
@@ -64,19 +68,29 @@ export default class PasFormComponent extends Component {
...
-  async submit() {
-    await this.submittableChanges.save();
+  @task(function* () {
+    yield this.saveableChanges.execute();
 
-    this.args.onSubmit();
-  }
+    yield this.args.onSave();
+
+    yield this.saveableChanges.rollback();
```
trbmcginnis, line 80: So this resets the changeset to its defaults after the model has been save?

client/app/controllers/packages/edit.js
```diff
@@ -4,13 +4,11 @@ import { action } from '@ember/object';
...
-    this.model.saveDescendants();
+    return this.model.saveDescendants();
   }
 
   @action
   submitPackage() {
-    this.model.submit();
-
-    this.transitionToRoute('packages.show', this.model.id);
+    return this.model.submit();
```
allthesignals, line 12: These probably just need to be async/awaited so the tasks see them.

trbmcginnis:
Thanks for that PR description :) very clear

client/app/components/packages/pas-form/edit.js
```diff
@@ -64,19 +68,29 @@ export default class PasFormComponent extends Component {
...
   }
 
-  @action
-  async save() {
-    await this.saveableChanges.save();
-
-    this.args.onSave();
+  get isSavingOrSubmitting() {
+    return this.save.isRunning || this.submit.isRunning;
   }
```
allthesignals, line 73: Wondering about the naming here... I wonder if `isSettling` is a better term?

client/app/components/packages/pas-form/edit.hbs
allthesignals: It should fit fine in `this.isSubmittable` (and same for the `isSaveable` getter). Because it's true that _if it's in the process of saving, it isn't in a saveable/submittable state_ until that task has finished.

client/app/components/packages/pas-form/edit.js
godfreyyeung: Yep, from my understanding of reading the docs. Usually when we call save() it does three things:

1. moves the Changeset changes to the underlying model
2. save the underlying model (which generates network PATCH requests)
3. reset the Changeset to a clean state (so it no longer reports any changes or the fact it is dirty) 

Since we no longer call .save(), we have to do these three steps manually. rollback() takes care of step 3.  

This is where it mentions it in the docs
https://github.com/poteto/ember-changeset#execute
image
I haven't extensively dug into this and verified it though. Maybe I should take some time and log the changeset without rollback() and see if it is dirty still.

client/app/components/packages/pas-form/edit.js
```diff
@@ -64,19 +68,29 @@ export default class PasFormComponent extends Component {
...
 
-    this.args.onSubmit();
-  }
+    yield this.args.onSave();
+
+    yield this.saveableChanges.rollback();
+  })
+  save;
+
+  @task(function* () {
```
allthesignals, line 84: I'm sad they haven't fixed the syntax for this.

You used to be able to say this:

```
@task()
*submit () {
}
```

Are you sure this isn't allowed? I think it looks a little less awkward.

client/app/components/packages/pas-form/edit.js
godfreyyeung: I think one use case of execute() not clearing away the changes is that you can apply the changes to multiple records if desired

client/tests/acceptance/user-can-click-package-edit-test.js
```diff
@@ -34,12 +35,18 @@ module('Acceptance | user can click package edit', function(hooks) {
...
     await click('[data-test-submit-button]');
     await click('[data-test-confirm-submit-button]');
 
     // for some reason promises aren't being captured so we await for settled state
     await settled();
 
+    // use waitFor as a way to "wait" for the transition
+    // within the pas-form/edit Component submit() task
+    await waitFor('[data-test-show-dcprevisedprojectname]');
+
```
allthesignals, line 49: Cool.

Only thing that _could_ go wrong is if one of those promises (inside the task) takes a long time for some reason (uncaught network request). But we have been really good about mocking server calls.

client/app/components/packages/pas-form/edit.js
godfreyyeung: Yep, I just verified it. rollback() does indeed clear away Changeset.changes:
image

image

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: Yes, that's my understanding of `fn` ! It's for "binding" a function to arguments from a certain context. 
http://api.emberjs.com/ember/3.11/classes/Ember.Templates.helpers/methods/fn?anchor=fn

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: I think we need to keep isSaveable separate because it represents another state altogether -- whether someone can click Save.

What do you mean by another getter?

godfreyyeung: Oh I see what you mean by building it into isSaveable/isSubmittable.

I think it's nice to have  `this.isSavingOrSubmitting` (or something that makes more sense to everyone) 
because later we can use this to succinctly disable inputs. For example
```
<Input 
 disabled={{this.isSavingOrSubmitting}}
```

I was thinking isSavingORSubmitting to be explicit about what is happening. instead of an opaque variable name like `isProcessing`.

How about this? 
```
  get isSaveable() {
    const {
      isDirty: isPasFormDirty,
      isValid: isPasFormValid,
    } = this.saveableChanges;
    const {
      isBblsDirty,
      isApplicantsDirty,
    } = this.pasForm;
     
    const isNotPatching = !(this.save.isRunning || this.submit.isRunning)
    return ((isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty) && isNotPatching;
  }
```

client/app/components/packages/pas-form/edit.hbs
andycochran: There's nothing preventing the user from closing the modal by clicking outside of it. The div that's behind the modal (which grays out the page) also has the `toggleModal` action. Perhaps `toggleModal` needs to consider `isRunning`?

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: I think I understand what you mean now! See comment above

client/app/components/packages/pas-form/edit.hbs
allthesignals: But how? It's telling us whether or not the form is saveable, and we want to prevent saveability if it's already in the process of saving.

Another getter like 
```
get theSaveButtonIsNotDisabled() {
  return !this.isSaveable || this.isSavingOrSubmitting;
}
```

I think it works just as well to keep that conditional logic in the saveable getter, though.

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: I think for a few reasons: 
   - So they can't double-click "Submit" 
   - Consistency in UI experience/design. The UI will consistently teach the user that disabled buttons +  spinner indicates there is "network/databse activity" and there shouldn't be any more interaction.

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: Yeah, i can see that  👍

client/app/components/packages/pas-form/edit.js
godfreyyeung: Originally thought a more descriptive name would be clearer than something like isProcessing. 

Maybe isPatching would imply the database record is being modified either through a save or submit?

client/app/components/packages/pas-form/edit.hbs
andycochran: The modal already is its own component. Let's chat about what could be more isolated here. I'd love to improve the modal if we can.

client/app/components/packages/pas-form/edit.hbs
allthesignals: > I was thinking to name it isSavingOrSubmitting to be explicit about what is happening. Vs. of an opaque variable name like isProcessing.

Good point, agree.


I like your code sample `const isNotPatching = !(this.save.isRunning || this.submit.isRunning)` but I think it's easier to follow if things are written so as to avoid negatives/double negatives:

Could we make this work with `isPatching` instead? :) :) :)

client/app/components/packages/pas-form/edit.js
godfreyyeung: Oh, I didn't come access/think of this. I can give it a go!

client/app/components/packages/pas-form/edit.hbs
allthesignals: yup, okay! we can totally split it out more.

client/app/components/packages/pas-form/edit.js
godfreyyeung: The reason I thought to be it here in the task is that it needs to happen _after_ the `rollback()` call. 

If it was within `this.args.onSubmit`, the transition would occur before `rollback()`

hmm, though now it just occurred to me that I may be able to rollback() immediately after execute()

client/app/controllers/packages/edit.js
godfreyyeung: Ooo!

client/app/components/packages/pas-form/edit.js
allthesignals: Yeah, Idk, this is a bit nit from me so disregard. "patching" is very HTTP-specific, i just wonder if we can find a better than for what this is rather than a camel-case version describing the code itself ;P

client/app/components/packages/pas-form/edit.js
godfreyyeung: I'm getting an error when I try that syntax:
image

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: @andycochran  oh good point. I can add that to toggleModal

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: I think I understood and gave it a go! Thanks 💯 for this suggestion. I think it looks much cleaner after a refactor

client/app/components/packages/pas-form/edit.js
```diff
@@ -57,26 +58,32 @@ export default class PasFormComponent extends Component {
       isApplicantsDirty,
     } = this.pasForm;
 
-    return (isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty;
+    return !this.isUpdatingDb && ((isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty);
```
allthesignals, line 61: cool :) Thanks. later, we can really easily split this out :)

client/app/components/packages/pas-form/edit.js
```diff
@@ -57,26 +58,32 @@ export default class PasFormComponent extends Component {
...
+  get isUpdatingDb() {
+    return this.save.isRunning || this.submit.isRunning;
   }
 
-  @action
-  async submit() {
-    await this.submittableChanges.save();
+  @task(function* () {
+    yield this.saveableChanges.execute();
+    yield this.saveableChanges.rollback();
```
allthesignals, line 74: i don't understand this part of the changesett API surface well enough so we'll wanna be vigilant about the side effects of this change...

client/tests/acceptance/user-can-click-package-edit-test.js
```diff
@@ -34,12 +35,18 @@ module('Acceptance | user can click package edit', function(hooks) {
...
+    await waitFor('[data-test-submit-button]:not([disabled])');
     await click('[data-test-submit-button]');
     await click('[data-test-confirm-submit-button]');
 
     // for some reason promises aren't being captured so we await for settled state
     await settled();
 
+    // use waitFor as a way to "wait" for the transition
+    // within the pas-form/edit Component submit() task
+    await waitFor('[data-test-show-dcprevisedprojectname]');
```
allthesignals, line 48: Wanna flag this for @trbmcginnis and @bfreeds and @andycochran — these two methods (awaiting settled and waitFor) are some good tools for your testing toolbox. the # 1 prob you will find with weird testing issues is dealing with asynchrony — teh test assertion not realizing there are promises/tasks still resolving. these helpers will assist with that.

client/app/components/packages/pas-form/edit.js
godfreyyeung: 🙌 🙌

client/app/components/packages/pas-form/edit.js
godfreyyeung: Agree. I haven't had extensive experience working with changeset rollbacks, either.

Issue #168, opened 2020-05-06
Visual Error Handling


allthesignals:
[Azure Requirement](#)

### User Story
As an applicant and developer, I want to see possible server error messages when I save the form or submit the form so that I can ask for support and get help quickly.

### [Wireframes](#)
Provide link to Figma
Paste in screenshots

### Display Logic
Captures specifics about the data and how attributes of the data should determine a record’s styling or where a record gets displayed on the page.

### Attributes
What attributes should the API response include?
### Filter Logic
Which records should be included? Which should be excluded?

PR #225, opened 2020-05-12
168: add styling, functionality, and tests for pas-form visual error handling

trbmcginnis:
Error handling on save and submit buttons of pas-form. 

Might run into conflicts with loading spinner branch, will resolve when that's merged

Addresses #168

### DIFF
--- a/client/app/components/packages/pas-form-error.hbs
+++ b/client/app/components/packages/pas-form-error.hbs
@@ -9,7 +9,7 @@
         There was a problem saving your information.
     </h5>
       {{#each @pasForm.adapterError.errors as |error|}}
-        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.message}}</p>
+        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.details}}</p>
       {{/each}}
       <p class="text-red-dark text-small">
         Please try again. If the problem persists, please contact City Planning at <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a> or <a href="tel:212-720-3300" class="nowrap">212-720-3300</a>.
--- a/client/app/components/packages/pas-form-error.hbs
+++ b/client/app/components/packages/pas-form-error.hbs
@@ -1,0 +1,18 @@
+{{#if @pasForm.adapterError}}
+  <div class="callout alert" data-test-error-saving>
+    <h5 class="text-red-dark">
+      <FaIcon
+      @icon="exclamation-circle"
+      @size="2x"
+      class="text-red float-left medium-margin-right tiny-margin-top"
+      />
+        There was a problem saving your information.
+    </h5>
+      {{#each @pasForm.adapterError.errors as |error|}}
+        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.message}}</p>
+      {{/each}}
+      <p class="text-red-dark text-small">
+        Please try again. If the problem persists, please contact City Planning at <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a> or <a href="tel:212-720-3300" class="nowrap">212-720-3300</a>.
+      </p>
+  </div>
+{{/if}}
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -711,6 +711,8 @@
           @isSaved={{and (not this.isSaveable) (not this.save.isRunning)}}
           data-test-save-button
         />
+        {{!-- Error Handling --}}
+        <Packages::PasFormError @pasForm={{this.pasForm}}/>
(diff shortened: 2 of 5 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -704,6 +704,22 @@
...
+                    @size="2x"
+                    class="text-red float-left medium-margin-right tiny-margin-top"
+                  />
+                  There was a problem saving your information.
+                </h5>
+                <p class="text-red-dark text-small">
+                  Please try again. If the problem persists, please contact City Planning at <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a> or <a href="tel:212-720-3300" class="nowrap">212-720-3300</a>.
+                </p>
+            </div>
+          {{/if}}
```
allthesignals, line 722: Shouldn't we also display the error message?

There's an API for dealing with errors on Ember Data models: https://davidtang.io/2016/01/09/handling-errors-with-ember-data.html

In fact, you could get away with just using template logic.

Take a look at `projects-error.hbs` to see some examples... 

Also, do you think we could put these error messages  messages into a component? Take a look at the length of edit.hbs... it's already pretty long!

Gimme call if you wanna talk through approach! :)

client/app/components/packages/pas-form/edit.js
```diff
@@ -66,16 +70,26 @@ export default class PasFormComponent extends Component {
...
-
-    this.args.onSave();
+    try {
+      await this.saveableChanges.save();
+
+      this.args.onSave();
+    } catch (e) {
+      this.saveError = e;
+      console.log('saveError', e);
+    }
```
allthesignals, line 80: Good.

client/tests/acceptance/error-message-appears-when-save-fails-test.js
```diff
@@ -0,0 +1,74 @@
...
+            dcpFirstname: 'first name',
+            dcpLastname: 'last name',
+            dcpEmail: 'email',
+            dcpType: '717170000',
+          }),
+        ],
+      }),
+    });
+
+    this.server.patch('/pas-forms/:id', { errors: [{ message: 'server problem' }] }, 500); // force mirage to error
```
allthesignals, line 61: Excellent

allthesignals: Just FYI, the shape of this error object comes from the [JSON API spec](https://jsonapi.org/format/#errors) :)

client/app/components/packages/pas-form/edit.js
```diff
@@ -7,6 +7,10 @@ import SaveablePasFormValidations from '../../../validations/saveable-pas-form';
 import SubmittablePasFormValidations from '../../../validations/submittable-pas-form';
 
 export default class PasFormComponent extends Component {
+  @tracked saveError;
+
+  @tracked submitError;
+
```
allthesignals, line 13: My worry here is that this is more state that we have to clean up. What happens when they fix the error? Does the message persist?

client/app/components/packages/pas-form/edit.hbs
trbmcginnis: Chatted with @andycochran and we didn't think it needed to be separated out into a separate component since there were only two instances where we used it, but I agree that edit.hbs is getting pretty long! Separating out into a component shouldn't be too hard if we want to shorten this file.

In response to "Shouldn't we also display the error message" -- how will the error message help our users? If you think it's important for them to know whether it's a 404, a 500, or some other error then yes we can add that, but I feel like that could get messy/might not be super relevant to our users? Let me know if you think differently

client/app/components/packages/pas-form/edit.js
trbmcginnis: that's a good point. I'm going to look into that API for dealing with errors on ember data and try to refactor :)

client/app/components/packages/pas-form/edit.hbs
allthesignals: Yeah, but this component is too long already, don't you think? We don't want to put ourselves in a position where it's so hard to maintain because it's over 1000 lines long.

As for showing the error message — including the status code — it will help our users provide better feedback to us so that we can fix whatever is causing the error in the first place. Usually they provide a screenshot. I think this is a cheaper way of getting bugs fixed faster, based on our experience with the LUP portal (where we didn't provide that feedback).

Great questions!

trbmcginnis: Good to know, will do!

client/app/components/packages/pas-form-error.hbs
```diff
@@ -0,0 +1,18 @@
...
+  <div class="callout alert" data-test-error-saving>
+    <h5 class="text-red-dark">
+      <FaIcon
+      @icon="exclamation-circle"
+      @size="2x"
+      class="text-red float-left medium-margin-right tiny-margin-top"
+      />
+        There was a problem saving your information.
+    </h5>
+      {{#each @pasForm.adapterError.errors as |error|}}
```
allthesignals, line 11: Great! What'd you find out about this? Why can't we use `@pasForm.errors`?

PR #227, opened 2020-05-12
prevent form elements having z-index > site-header

andycochran:
### Before: 

image

### After: 

image


allthesignals:
Sweet! @andycochran also reminds me, is there any way to make the styling for the typeahread results a littel better? It's very small and awk, and doesn't fit with the rest of the nice design!

PR #228, opened 2020-05-12
View files

allthesignals:
This PR provides handling for the get method on the /documents resource, which takes a sharepoint relative server URL, retrieves the file, and pipes it to the user.

This is not a shareable URL and may be too memory intensive but probably isn't an issue unless we expect hundreds of simultaneous downloads.

This also adds some helpers that generate deep links to CRM entities from models

### DIFF
--- a/server/src/document/document.controller.ts
+++ b/server/src/document/document.controller.ts
@@ -9,6 +9,9 @@ import {
   Delete,
   Query,
   HttpCode,
+  Param,
+  Get,
+  Res,
 } from '@nestjs/common';
 import { ConfigService } from '../config/config.service';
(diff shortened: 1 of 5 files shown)


server/src/document/document.controller.ts
```diff
@@ -68,4 +71,12 @@ export class DocumentController {
   async destroy(@Query('serverRelativeUrl') serverRelativeUrl) {
     return this.crmService.deleteSharepointFile(serverRelativeUrl);
   }
+
```
godfreyyeung, line 74: Would prefer some comments here documenting the `path` parameter, with an example path! 
Otherwise, looks good

Issue #235, opened 2020-05-13
For the Proposed Land Use Actions field. when the user selects an change an option, the question and answer text box is displayed twice


Hemachekuri:
## Browser(s) 
IE Browser and chrome

This issue is happening in all section,when user selects on another similar field is displayed.

## What should happen:
Proposed Land Use Actions, Proposed Land Use Actions. when the user selects Add a Proposed Action: change in city map, When user selects start entering character in input box, then another citymap should not be displayed,
## What happened instead:
Proposed Land Use Actions, Proposed Land Use Actions. when the user selects Add a Proposed Action: change in city map, When user selects start entering character in input box, then another citymap is displayed,
## How to reproduce this bug:
Log Into Applicant Portal
Click on the Edit Pre-Application Statement button
Navigate to the Proposed Land Use Actions, Proposed Land Use Actions. when the user selects Add a Proposed Action: change in city map, When user selects start entering character in input box, then another citymap is displayed,

Proposed Land Use Actions. when the user selects Add a Proposed Action: change in city map, 
![image](https://user-images.githubusercontent.com/63460627/81833494-ca567a80-950d-11ea-8d88-beffdde46830.png)
When user selects start entering character in input box, then another citymap is displayed,
![image](https://user-images.githubusercontent.com/63460627/81833595-e9eda300-950d-11ea-95aa-feec9a2f2e35.png)

trbmcginnis:
okay I see what's going on, looks like the new filtering methods check for if an action already exists on the pasForm model and if it does, it adds it to an "existing" array. We're adding newly added actions to a "new" array and then combining the "new" and "existing" arrays. Because inputting a number in that input box causes a save directly to the model, the app is double counting that action and adding it to both the "new" and "existing" array -- meaning it's listed twice. Setting up a fix after the component meeting! and writing an associated test

hannahkates:
@trbmcginnis I don't think it changes implementation for this bug, but want to link this issue with #245 for reference

trbmcginnis:
@hannahkates Thanks! Going to implement this bug fix and then I'll tackle this one

ksinghprod:
@hannahkates @Hemachekuri 
This issue is now fixed 

![image](https://user-images.githubusercontent.com/57362229/82091079-c0c34300-96c4-11ea-89a0-6aaa0116a691.png)

PR #242, opened 2020-05-13
235: update new and existing actions arrays to assure actions are not double counted

trbmcginnis:
**Bug fix**:

In the landUseAction component --> previously, when a user would enter a value into one of the action inputs (and therefore the field was set on the model), the action was added to the "existing" actions array, meanwhile it still also existed on the "new" actions array, meaning it showed up twice in the UI.

This PR filters the "new" actions array so that actions that have fields set on the model will be filtered out of that array, avoiding these actions being double counted.

**This PR also:**
- updated the property names in `allLandUseActionOptions`
- moved `allLandUseActionOptions` to be an exported constant

Addresses #235

### DIFF
--- a/client/app/components/land-use-action.js
+++ b/client/app/components/land-use-action.js
@@ -2,6 +2,64 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 
+// lookup to match action titles to their associated attributes
+export const allLandUseActionOptions = [
+  {
+    name: 'Acquisition of Real Property', countField: 'dcpPfacquisitionofrealproperty', attr1: '', attr2: '',
+  },
+  {
+    name: 'Change in CityMap', countField: 'dcpPfchangeincitymap', attr1: '', attr2: '',
+  },
+  {
+    name: 'Concession', countField: 'dcpPfconcession', attr1: '', attr2: '',
+  },
+  {
+    name: 'Disposition of Real Property', countField: 'dcpPfdispositionofrealproperty', attr1: '', attr2: '',
+  },
+  {
+    name: 'Franchise', countField: 'dcpPffranchise', attr1: '', attr2: '',
+  },
+  {
+    name: 'Housing Plan & Project', countField: 'dcpPfhousingplanandproject', attr1: '', attr2: '',
+  },
+  {
+    name: 'Landfill', countField: 'dcpPflandfill', attr1: '', attr2: '',
+  },
+  {
+    name: 'Modification', countField: 'dcpPfmodification', attr1: 'dcpPreviousulurpnumbers1', attr2: '',
+  },
+  {
+    name: 'Renewal', countField: 'dcpPfrenewal', attr1: 'dcpPreviousulurpnumbers2', attr2: '',
+  },
+  {
+    name: 'Revocable Consent', countField: 'dcpPfrevocableconsent', attr1: '', attr2: '',
+  },
+  {
+    name: 'Site Selection - Public Facility', countField: 'dcpPfsiteselectionpublicfacility', attr1: '', attr2: '',
+  },
+  {
+    name: 'UDAAP', countField: 'dcpPfudaap', attr1: '', attr2: '',
+  },
+  {
+    name: 'URA', countField: 'dcpPfura', attr1: '', attr2: '',
+  },
+  {
+    name: 'Zoning Authorization', countField: 'dcpPfzoningauthorization', attr1: 'dcpZoningauthorizationpursuantto', attr2: 'dcpZoningauthorizationtomodify',
+  },
+  {
+    name: 'Zoning Certification', countField: 'dcpPfzoningcertification', attr1: 'dcpZoningpursuantto', attr2: 'dcpZoningtomodify',
+  },
+  {
+    name: 'Zoning Map Amendment', countField: 'dcpPfzoningmapamendment', attr1: 'dcpExistingmapamend', attr2: 'dcpProposedmapamend',
+  },
+  {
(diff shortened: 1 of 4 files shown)


client/app/components/land-use-action.js
```diff
@@ -74,12 +74,16 @@ export default class LandUseActionComponent extends Component {
...
 
-  // Actions selected by user in dropdown, or already
-  // existing in CRM
+  // Actions selected by user in dropdown ("new"), or already
+  // existing in CRM ("existing")
   get landUseActionSelections() {
+    const { pasForm } = this.args;
+    // Filter the "new" actions array so that the action does not exist on BOTH the "exisiting" and "new" arrays,
+    // when user enters a value in the action inputs (and therefore updates the model)
+    const filteredNewActions = this.newLandUseActionSelections.filter((action) => !pasForm[action.field] && !pasForm[action.attr1] && !pasForm[action.attr2]);
```
bfreeds, line 83: @trbmcginnis could we just filter based on the `action.field`?  Just trying to understand why we need to explicitly look for `attr1` and `attr2` as well.
```js
    const filteredNewActions = this.newLandUseActionSelections.filter(
      action => !pasForm[action.field]
    );
```

client/app/components/land-use-action.js
```diff
@@ -92,29 +96,29 @@ export default class LandUseActionComponent extends Component {
...
   // The selected option is added to the newLandUseActionSelections array
   // (and therefore displays in the list of selected actions).
   @action
   addNewLandUseActionSelection() {
-    // use unshiftObject to place new action at top of list
     const selection = this.selectedLandUseAction.field;
 
     if (selection) {
-      this.newLandUseActionSelections.unshiftObject(this.selectedLandUseAction);
+      this.newLandUseActionSelections.pushObject(this.selectedLandUseAction);
```
bfreeds, line 108: Interesting, was initially confused what `pushObject` was because I couldn't find it in JS docs, before learning that it's an Ember array method.

Why do we need to switch to using `pushObject` from `unshiftObject`?

client/app/components/land-use-action.js
trbmcginnis: If you look at how "exisiting" actions are defined, they are defined as having at least ONE of their attributes (including `field`, `attr1`, `attr2`) saved to the model (I might actually rename these to make them more clear). So if an action has ANY of their fields stored on the model/db then they will be put into the existing actions. This is to account for the case when a user does a partial save after having filled out only some inputs. 

existing = `this.allLandUseActionOptions.filter((option) => pasForm[option.field] || pasForm[option.attr1] || pasForm[option.attr2]);` 

So if we're trying to include only "new" actions in the `filteredNewActions` array (and make sure they don't exist in "existing") we need to make sure that the action has NO attributes saved at all on the model. They're basically just local variables at that point.

client/app/components/land-use-action.js
trbmcginnis: Originally, when we only had one array that made up the selected actions list, we wanted new actions to show up at the top of the list -- so `unshiftObject` was used rather than `pushObject` (which adds to the end of an array). Now that we are combining two arrays to create that selected actions list, pushing the object to the front of the "new" actions array won't actually put it at the top of the actions list in the UI. Going to make a new issue for making sure most recently added actions show up at the top of the list.

client/app/components/land-use-action.js
bfreeds: From the action where users add a new selection, it looks like `field` is the the unique key for identifying / adding the new land use action:
```js
 @action
  addNewLandUseActionSelection() {
    const selection = this.selectedLandUseAction.field;

    if (selection) {
      this.newLandUseActionSelections.pushObject(this.selectedLandUseAction);
      this.selectedLandUseAction = {};
    }
  }
```

But if I'm understanding, it is possible for us to get existing land use actions from CRM where the field is empty but the `attrs` are not?  In other words, `attrs` don't require the `field`?

client/app/components/land-use-action.js
bfreeds: That's super helpful, thanks for clarifying!

client/app/components/land-use-action.js
trbmcginnis: I think the `const selection = this.selectedLandUseAction.field` is there because if you just did `this.selectedLandUseAction`, whether it was an empty object or a filled object, it would still come back as truthy. 

There are technically "requirements" for what you can submit to CRM but that's handled on the Submit button's validations (if an action has extra attributes and the count field is filled in, then the extra attributes are required to be filled in to Submit). But because a user can hit `save` at any point in the process, there might be times with any of those fields (`field`, `attr1`, or `attr2`) are filled and none of the others are.

client/app/components/land-use-action.js
```diff
@@ -2,120 +2,124 @@ import Component from '@glimmer/component';
...
 
-  // Actions selected by user in dropdown, or already
-  // existing in CRM
+  // Actions selected by user in dropdown ("new"), or already
+  // existing in CRM ("existing")
   get landUseActionSelections() {
+    const { pasForm } = this.args;
+    // Filter the "new" actions array so that the action does not exist on BOTH the "exisiting" and "new" arrays,
+    // when user enters a value in the action inputs (and therefore updates the model)
+    const filteredNewActions = this.newLandUseActionSelections.filter((action) => !pasForm[action.countField] && !pasForm[action.attr1] && !pasForm[action.attr2]);
```
godfreyyeung, line 83: Ohh 💡  I see. So now filteredNewActions will hold the actions selected from the dropdown but not yet given any text inputs. A user might select multiple actions from the dropdown before scrolling through the action fields and filling them in.

client/app/components/land-use-action.js
allthesignals: Yeah, this is a case where Ember's template observers need to be notified through this method.

It's akin to needing to call ".setState" in React.

I'm not sure if it's still true, though, but we use it out of habit. Would be worth researching.

PR #244, opened 2020-05-13
Save-button component

andycochran:
This PR adds a `<Packages::save-button />` component for the save and submit buttons, which now visually indicate to the user that the save action was successful. 

image

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -693,14 +693,14 @@
     <div class="cell large-4">
 
       <div class="sticky-sidebar">
-        <Packages::save-button
+        <Packages::SaveButton
           @clickEvent={{perform this.save}}
           @isEnabled={{this.isSaveable}}
           @isRunning={{this.save.isRunning}}
           @isSaved={{and (not this.isSaveable) (not this.save.isRunning)}}
           data-test-save-button
         />
-        <Packages::save-button
+        <Packages::SaveButton
           @clickEvent={{this.toggleModal}}
           @isEnabled={{this.isSubmittable}}
           @isRunning={{this.submit.isRunning}}
@@ -727,7 +727,7 @@
                   </button>
                 </div>
                 <div class="cell large-6 text-right">
-                  <Packages::save-button
+                  <Packages::SaveButton
                     @clickEvent={{perform this.submit}}
                     @isEnabled={{not this.submit.isRunning}}
                     @isRunning={{this.submit.isRunning}}
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -693,35 +693,21 @@
     <div class="cell large-4">
 
       <div class="sticky-sidebar">
-        <p>
-          <button
-            type="button"
-            class="button expanded large"
-            disabled={{not this.isSaveable}}
-            {{on "click" (perform this.save)}}
-            data-test-save-button
-          >
-            {{!-- TODO: Make this button work with the action passed in @save--}}
-            {{#if this.save.isRunning}}
-              <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
-            {{else}}
-              Save
-            {{/if}}
-          </button>
-          <button
-            type="button"
-            class="button expanded large secondary"
-            disabled={{not this.isSubmittable}}
-            {{on "click" this.toggleModal}}
-            data-test-submit-button
-          >
-            {{#if this.submit.isRunning}}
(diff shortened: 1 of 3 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -741,19 +727,14 @@
                   </button>
                 </div>
                 <div class="cell large-6 text-right">
-                  <button
-                    type="button"
-                    class="button expanded large no-margin"
-                    disabled={{this.submit.isRunning}}
-                    {{on "click" (perform this.submit)}}
+                  <Packages::save-button
```
allthesignals, line 730: I think the naming should be camelCase `Packages::SaveButton`

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -693,35 +693,21 @@
...
-          >
-            {{#if this.submit.isRunning}}
-              <FaIcon @icon="spinner" @spin="true" size="5x" class="light-gray" />
-            {{else}}
-              Submit
-            {{/if}}
-          </button>
-        </p>
+        <Packages::save-button
+          @clickEvent={{perform this.save}}
```
allthesignals, line 697: can we name this the same convention we have? `onClick`?

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -693,35 +693,21 @@
...
+          @isEnabled={{this.isSaveable}}
+          @isRunning={{this.save.isRunning}}
+          @isSaved={{and (not this.isSaveable) (not this.save.isRunning)}}
```
allthesignals, line 700: this seems like a lot of state to be passing down. I wonder how much of these conditionals can be extracted from pas-form/edit.js itself? Probably can't do it yet.

To reduce the public api surface of this component, can we just pass the `save` task? A requirement of the button is that it expects a "task" type object. They all have the same API.

When we have that inside this component, then we can derive all that state:

```handlebars
<Packages::SaveButton
  @onClick={{perform this.save}}
  @isEnabled={{this.isSaveable}}
  @task={{this.save}}
/>
```

```javascript
class SaveButton {
  get isSaved() {
    return (!this.args.isEnabled && !this.save.isRunning)
  }
}
```

client/app/components/packages/save-button.hbs
```diff
@@ -0,0 +1,17 @@
...
+>
+  {{#if @isRunning}}
+    <FaIcon @icon="spinner" @spin={{true}} @pulse={{true}} class="small-margin-right" />
+    {{if @isSubmit 'Submitting' 'Saving'}}
+  {{else if @isSaved}}
+    <FaIcon @icon="check" class="small-margin-right" />
+    {{if @isSubmit 'Submitted' 'Saved'}}
+  {{else}}
+    {{if @isSubmit 'Submit' 'Save'}}
+  {{/if}}
```
allthesignals, line 16: Perfect.

Only way I can think to make this better is with named blocks — then the consumer would have more control over the naming of the state, and then this component wouldn't care about the specifics of the language:

```handlebars
<SaveButton as |state|>
  <state.running>
    Saving
  </state.running>
  <state.isSaved>
    Saved
  </state.isSaved>
  <state.isIdle>
     Save
  </state.isIdle>
</SaveButton>
```

The RFC has been accepted for a while: https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md.  Jut a matter of time!

Anyway, this is perf.

client/app/components/packages/pas-form/edit.hbs
andycochran: Oops. Fixed in 22c8602.

client/app/components/packages/pas-form/edit.hbs
andycochran: 👍 Addressed in 5042596.

client/app/components/packages/pas-form/edit.hbs
andycochran: I tried it that way and it got… complicated. I discovered that `@isRunning` and `@isSaved` args made everything simpler. 

I'm passing states explicitly because there are different conditions for Save and Submit — e.g. we don't want "Submitted" to turn green when disabled and finished running, so if I don't pass an `@isSaved` arg it can't turn green. If the states are computed, it makes JS necessary (the component is template-only rn). 

I thought it better to just pass the state as args because computing the states in the component JS would need to cover multiple edge cases — e.g. the two "Submit" buttons have different conditions for `@isEnabled`, one of which can't be computed off of the `@onClick`. 

Maybe we can refactor this later. Once we're dealing with multiple packages, there could be other factors that determine the state of these buttons.

PR #248, opened 2020-05-14
187 Enable Save when files marked for upload or deletion. Upload+delete files on Save. 

godfreyyeung:
This PR ...
1) Enable the Save button when files marked for upload or deletion
2) Calls fileManager.save() when the PAS Form Save button is clicked. 
fileManager.save() will execute file upload and deletion for files appropriately marked.

### DIFF
--- a/client/app/components/packages/attachments.js
+++ b/client/app/components/packages/attachments.js
@@ -2,9 +2,7 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 
 /**
-  * This component creates a new File Manager specifically for
-  * the passed in package. The File Manager is among one of many
-  * on the the application File Manager service.
+  * This component wires a fileManager to the attachments UI.
   * @param      {Package Model}  package
   */
 export default class PackagesAttachmentsComponent extends Component {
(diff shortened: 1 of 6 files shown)


client/app/services/file-manager.js
```diff
@@ -20,6 +21,26 @@ export default class FileManager {
     this.filesToUpload = filesToUpload; // EmberFileUpload QUEUE Object
   }
 
+  @tracked existingFiles;
```
bfreeds, line 24: From conversation - double check to see if we need this (might not).

client/app/components/packages/pas-form/edit.js
```diff
@@ -57,8 +57,16 @@ export default class PasFormComponent extends Component {
       isBblsDirty,
       isApplicantsDirty,
     } = this.pasForm;
-
-    return !this.isUpdatingDb && ((isPasFormDirty && isPasFormValid) || isBblsDirty || isApplicantsDirty);
+    const {
+      isDirty: isAttachmentsDirty,
+    } = this.package.fileManager;
+
+    return !this.isUpdatingDb && ((isPasFormDirty
```
bfreeds, line 64: Nice, this is more readable.

client/app/components/packages/attachments.js
```diff
@@ -22,6 +20,11 @@ export default class PackagesAttachmentsComponent extends Component {
     this.fileManager.unMarkFileForDeletion(file);
   }
 
+  @action
```
bfreeds, line 23: I'd just add a comment something along the lines of this so future developers can understand why this action isn't being passed any files as parameters (because it doesn't need to know, just incremenets the variable by one to trigger `isDirty`).
```js
 // we need this to track is the attachments isDirty, action just increments a variable by one so we can respond in UI
  @action
  trackFileForUpload() {
    this.fileManager.trackFileForUpload();
  }
```

allthesignals:
@godfreyyeung I wonder if we could leverage ember-changeset for this?

Issue #240, opened 2020-05-13
When the Applicant clicks on and adds other team member, the data saves as an applicant type in CRM, not in the Authorized Applicant Representative grid


ksinghprod:
## Browser(s) and Device(s) observed on:
Chrome, IE, Firefox

## What should happen:

When the Applicant clicks on and adds other team member, the data should save as (type) Authorized Applicant Representative in CRM, not (type) Applicant 

## What happened instead:
the data saved as an applicant type in CRM, not Authorized Applicant Representative (type) 

## How to reproduce this bug:

1. Log into applicant Portal
2. Click on the Edit Pre-application statement button
3. navigate to the applicant team section
4. click on the add other team member button
5. enter information and click save
6. log into UAT2 CRM and go to the same PAS form/package
7. Navigate to the the Prospective Applicant/Representative Contact section (error is displayed here)

![image](https://user-images.githubusercontent.com/57362229/81846169-ac921100-951f-11ea-9706-7c0f01281df5.png)

![image](https://user-images.githubusercontent.com/57362229/81846244-c2073b00-951f-11ea-9149-b0cf509a526d.png)

bfreeds:
When a user clicks `Add Other Team Member` and saves, the `Team Member` should not appear in the second table on the Applicant team on the PAS Form "Applicant Representative Information."

Need to investigate why the records being added through `Add Applicant Team Member` is being posted to the incorrect table.  
![image](https://user-images.githubusercontent.com/5316367/81855704-b02c9480-952d-11ea-8b02-9d1dac019d1b.png)

bfreeds:
Likely same root cause as #238

ksinghprod:
Hi @hannahkates  retested this bug with new modified "Add other team member button", the added other team member is still displaying in the applicant grid/table on CRM 

cc: @Hemachekuri 

![image](https://user-images.githubusercontent.com/57362229/82240356-4d117800-9908-11ea-9aca-d2f2d54013c7.png)

![image](https://user-images.githubusercontent.com/57362229/82240548-9a8de500-9908-11ea-9bb8-916d880c5f8a.png)

bfreeds:
@ksinghprod @hannahkates I opened a draft pr (#260) that may have moved this to the wrong zenhub pipeline -- the fix to this and #238 are being addressed in that PR (which is still work in progress).

bean75:
[...]

Issue #238, opened 2020-05-13
When the Applicant enters the Organization/Authorized Representative information in the Applicant Team section, the information entered displays in the Applicant Information grid on CRM


ksinghprod:
## Browser(s) and Device(s) observed on:
Chrome, Firefox, IE

## What should happen:
When the Applicant enters the Organization/Authorized Representative information in the Applicant Team section from the Applicant Portal, the information entered should be displayed in the Authorized Representative Information grid on CRM

## What happened instead:
 the information entered by the applicant displays in the Applicant Information grid on CRM not the Authorized Representative Information

## How to reproduce this bug:
1 Log into the Applicant Portal
2 Click on the Edit Pre-Application Statement button
3 navigate to the Applicant team section
4 select the organization radio button
5 enter data for all fields in the applicant team section and save
6 log into UAT2 CRM, open the same project, PAS form
7 navigate to the Prospective Applicant/Representative Contact section (error displayed here)

![image](https://user-images.githubusercontent.com/57362229/81842431-01cb2400-951a-11ea-9c4a-a2d8d1194f2d.png)

![image](https://user-images.githubusercontent.com/57362229/81842472-10194000-951a-11ea-93a7-6ab2b687fc84.png)

![image](https://user-images.githubusercontent.com/57362229/81842565-37700d00-951a-11ea-9324-2bfda577f4cb.png)

bean75:
This is not a bug, it is functioning as designed - I believe we discussed the clarification needed is to call "Authorized Applicant Representative" an "Authorized Organization Representative."

ksinghprod:
Closing this bug as per Aline's Comment 
@Hemachekuri

PR #260, opened 2020-05-15
240:  Route applicants to the correct CRM entity/table

bfreeds:
This PR addresses #238 and addresses #240, by adding conditional logic to our api server for handling applicants.  

Previously all applicants being added in applicant portal would go to the `Applicant Information` grid in CRM, and would not bring in existing `Applicant Representative Information` records.  

This fixes by:
- Including the representative entity in the packages controller (for getting existing applicant representatives)
- Gives the frontend an attribute `target_entity` so the app knows what kind of applicant to render
- Conditionally Posts/Patches/Deletes the appropriate CRM entity through conditional logic in the applicant controller.

### DIFF
--- a/server/src/packages/packages.service.ts
+++ b/server/src/packages/packages.service.ts
@@ -46,6 +46,7 @@ export class PackagesService {
         dcp_pasformid eq ${_dcp_pasform_value}
       &$expand=
         dcp_dcp_applicantinformation_dcp_pasform,
+        dcp_dcp_applicanrepresentativetinformation_dcp_pasform,
         dcp_package,
         dcp_dcp_projectbbl_dcp_pasform($filter=statecode eq 0)
     `);
--- a/server/src/packages/packages.service.ts
+++ b/server/src/packages/packages.service.ts
@@ -46,7 +46,7 @@ export class PackagesService {
         dcp_pasformid eq ${_dcp_pasform_value}
       &$expand=
         dcp_dcp_applicantinformation_dcp_pasform,
-        dcp_dcp_applicanrepresentativetinformation_dcp_pasform,
+        dcp_dcp_applicantrepinformation_dcp_pasform,
         dcp_package,
         dcp_dcp_projectbbl_dcp_pasform($filter=statecode eq 0)
     `);
--- a/server/src/packages/pas-form/applicants/applicants.controller.ts
+++ b/server/src/packages/pas-form/applicants/applicants.controller.ts
@@ -15,11 +15,11 @@ import { JsonApiSerializeInterceptor } from '../../../json-api-serialize.interce
 import { AuthenticateGuard } from '../../../authenticate.guard';
 import { JsonApiDeserializePipe } from '../../../json-api-deserialize.pipe';
 
-// CRM has two entities that encompass the applicant team, 
+// CRM has two entities that encompass the applicant team,
 // dcp_applicantinformation and dcp_applicantrepinformation
 // TODO: This controller exposes these two entites as one "applicant" resource
 // and handles routes applicant resources from and to the appropriate CRM entity
-// the frontend provides (and needs) a custom property "targetEntity"
+// the frontend uses a custom attribute, "targetEntity",
 // to track the ultimate CRM entity destination (and source when getting existing)
 
 export const APPLICANT_REPRESENTATIVE_ATTRIBUTES = [
@@ -38,7 +38,6 @@ export const APPLICANT_ATTRIBUTES = [
   ...APPLICANT_REPRESENTATIVE_ATTRIBUTES,
   'dcp_type',
 ];
-
 
 @UseInterceptors(
   new JsonApiSerializeInterceptor('applicants', {
@@ -67,31 +66,44 @@ export class ApplicantsController {
   @Post('/')
   create(@Body() body) {
     const { target_entity } = body;
-  
+
     let allowedAttrs;
 
     // determine the appropraite attributes based on the entity type
     if (target_entity === 'dcp_applicantinformation') {
       allowedAttrs = pick(body, APPLICANT_ATTRIBUTES);
-    } else if (target_entity === 'dcp_applicantrepinformation') {
-      allowedAttrs = pick(body, APPLICANT_REPRESENTATIVE_ATTRIBUTES);
-    }
-
-    if (body.pas_form) {
-      let applicantBody = {
-        ...allowedAttrs,
+      if (body.pas_form) {
+        return this.crmService.create('dcp_applicantinformations', {
+          ...allowedAttrs,
+          // Dy365 syntax for associating a newly-created record
+          // with an existing record.
+          // see: https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/create-entity-web-api#associate-entity-records-on-create
+          'dcp_dcp_applicantinformation_dcp_pasform@odata.bind': [
+            `/dcp_pasforms(${body.pas_form})`,
+          ],
+        });
+      } else {
+        return this.crmService.create(`dcp_applicantinformations`, allowedAttrs);
       }
-      const objKey = `dcp_${target_entity}_dcp_pasform@odata.bind` 
-      // Dy365 syntax for associating a newly-created record
-      // with an existing record.
-      // see: https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/create-entity-web-api#associate-entity-records-on-create
-      applicantBody[objKey] = [`/dcp_pasforms(${body.pas_form})`]
-
-      console.log(applicantBody);
-      
-      return this.crmService.create(`${target_entity}s`, applicantBody);
-    } else {
-      return this.crmService.create(`${target_entity}s`, allowedAttrs);
+    } else if (target_entity === 'dcp_applicantrepresentativeinformation') {
+      allowedAttrs = pick(body, APPLICANT_REPRESENTATIVE_ATTRIBUTES);
+      if (body.pas_form) {
+        return this.crmService.create(
+          'dcp_applicantrepresentativeinformations',
+          {
+            ...allowedAttrs,
+            // Dy365 syntax for associating a newly-created record
+            // with an existing record.
+            // see: https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/create-entity-web-api#associate-entity-records-on-create
+            'dcp_dcp_applicantrepinformation_dcp_pasform@odata.bind': [`/dcp_pasforms(${body.pas_form})`],
+          },
+        );
+      } else {
+        return this.crmService.create(
+          `dcp_applicantrepinformations`,
+          allowedAttrs,
+        );
+      }
     }
   }
 
@@ -104,5 +116,3 @@ export class ApplicantsController {
     };
   }
 }
-
-
--- a/server/src/packages/pas-form/applicants/applicants.controller.ts
+++ b/server/src/packages/pas-form/applicants/applicants.controller.ts
@@ -36,7 +36,6 @@ export const APPLICANT_REPRESENTATIVE_ATTRIBUTES = [
   'dcp_state',
   'dcp_zipcode',
   'dcp_phone',
-  'target_entity',
 ];
 
 export const APPLICANT_ATTRIBUTES = [
@@ -63,8 +62,7 @@ export class ApplicantsController {
     let allowedAttrs;
 
     if (target_entity === 'dcp_applicantinformation') {
-      // we can't send target_entity to CRM because it doesn't exist there
-      allowedAttrs = pick(body, APPLICANT_ATTRIBUTES.filter(attribute => attribute !== 'target_entity'));
+      allowedAttrs = pick(body, APPLICANT_ATTRIBUTES);
 
       await this.crmService.update(
         'dcp_applicantinformations',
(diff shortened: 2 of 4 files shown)


server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -15,15 +32,16 @@ export const APPLICANT_ATTRIBUTES = [
...
 ];
 
-@UseInterceptors(new JsonApiSerializeInterceptor('applicants', {
-  id: 'dcp_applicantinformationid',
-  attributes: [
-    ...APPLICANT_ATTRIBUTES,
-  ],
-}))
+@UseInterceptors(
+  new JsonApiSerializeInterceptor('applicants', {
```
bfreeds, line 40: Readability fix from
`prettier --write applicants.controller.ts`

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -1,10 +1,27 @@
...
+  Controller,
+  Patch,
+  Body,
+  Param,
+  Post,
+  UseInterceptors,
+  UseGuards,
+  UsePipes,
+  Delete,
+} from '@nestjs/common';
```
bfreeds, line 11: Readability fix from
prettier --write applicants.controller.ts

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -53,7 +71,9 @@ export class ApplicantsController {
         // Dy365 syntax for associating a newly-created record
         // with an existing record.
         // see: https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/create-entity-web-api#associate-entity-records-on-create
-        'dcp_dcp_applicantinformation_dcp_pasform@odata.bind': [`/dcp_pasforms(${body.pas_form})`],
+        'dcp_dcp_applicantinformation_dcp_pasform@odata.bind': [
```
bfreeds, line 74: Readability fix from
prettier --write applicants.controller.ts

server/src/packages/pas-form/applicants/applicants.controller.ts
allthesignals: I would just find a way to get this working as a linting requirement so it's consistent across.

Now the other controllers will look different.

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -15,15 +32,16 @@ export const APPLICANT_ATTRIBUTES = [
   'dcp_state',
   'dcp_zipcode',
   'dcp_phone',
-  'dcp_type',
+  'dcp_type', // only on dcp_applicantinformation entity, not applicantrepresentative
+  'targetEntity', // custom attribute to route to/from correct crm entity
```
allthesignals, line 36: Originally, the `APPLICANT_ATTRIBUTES` were supposed to be a list of valid CRM attributes names _including our internal naming for relationships (`dcp_project` is `projects`)_.

Now that `targetEntity` is included, it will be allowed through the `PATCH` and `POST` methods, and CRM will reject it because it's not a real key. We might want to keep this "one-off" handling in the controller methods themselves until we're comfortable handling this sort of situation in a more generic way.

server/src/packages/pas-form/applicants/applicants.controller.ts
allthesignals: Eh, I think it's nice, but I'd rather it be applied everywhere instead of just one place. I'd prefer all-or-nothing.

server/src/packages/pas-form/applicants/applicants.controller.ts
bfreeds: I see, instead am going to organize it like this, adding a second ATTRIBUTES variable to show the different attributes for the applicant vs. appicant rep entities.   Trying to use `targetEntity` in the http methods instead:
```js
export const APPLICANT_REPRESENTATIVE_ATTRIBUTES = [
  'dcp_firstname',
  'dcp_lastname',
  'dcp_organization',
  'dcp_email',
  'dcp_address',
  'dcp_city',
  'dcp_state',
  'dcp_zipcode',
  'dcp_phone',
];

export const APPLICANT_ATTRIBUTES = [
  ...APPLICANT_REPRESENTATIVE_ATTRIBUTES,
  'dcp_type',
];
```

server/src/packages/packages.service.ts
```diff
@@ -46,6 +46,7 @@ export class PackagesService {
         dcp_pasformid eq ${_dcp_pasform_value}
       &$expand=
         dcp_dcp_applicantinformation_dcp_pasform,
+        dcp_dcp_applicanrepresentativetinformation_dcp_pasform,
```
bfreeds, line 49: Looking at the [crm metadata](https://dcppfsuat2.crm9.dynamics.com/api/data/v9.1/$metadata), is looks like this needs to be `dcp_dcp_applicantreptinformation_dcp_pasform`
image

bfreeds:
Need to figure out when/how to use `dcp_applicantrepinformation` vs. `dcp_applicantrepresentativeinformation`.  From the CRM power pane, the entity appears to be `dcp_applicantrepresentativeinformation`.

image

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -39,24 +61,37 @@ export class ApplicantsController {
...
+      applicantBody[objKey] = [`/dcp_pasforms(${body.pas_form})`]
 
-        // Dy365 syntax for associating a newly-created record
-        // with an existing record.
-        // see: https://docs.microsoft.com/en-us/powerapps/developer/common-data-service/webapi/create-entity-web-api#associate-entity-records-on-create
-        'dcp_dcp_applicantinformation_dcp_pasform@odata.bind': [`/dcp_pasforms(${body.pas_form})`],
-      });
+      console.log(applicantBody);
+      
+      return this.crmService.create(`${target_entity}s`, applicantBody);
```
bfreeds, line 92: Currently getting this error when trying to `Post`:
image

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -39,24 +61,37 @@ export class ApplicantsController {
...
     } else {
-      return this.crmService.create('dcp_applicantinformations', allowedAttrs);
```
bfreeds, line 59: @allthesignals is pluralizing the entity part of the web crm api?  Getting an error when trying to parametrize this `create` call to accomodate the `dcp_applicantrepinformation` entity.

server/src/packages/pas-form/applicants/applicants.controller.ts
bfreeds: Tried to `post` without the pluralization, received same error
image

bfreeds: This is the error when using `dcp_applicantrepresentativeinformation` (passed in through frontend action), which is what shows in the CRM power pane, instead of the shortened `dcp_applicantrepinformaiton` found in the crm metadata.
image

server/src/packages/pas-form/applicants/applicants.controller.ts
bfreeds: @allthesignals thanks for flagging this -- super helpful.

Currently how I addressed this in POST and PATCH is filter to remove `target_entity` from the `allowedAttrs,`
```js
      allowedAttrs = pick(body, APPLICANT_REPRESENTATIVE_ATTRIBUTES.filter(attribute => attribute !== 'target_entity'));
```
This is a little duplicative, but it works for now.  Happy to adjust to make it more ergonomic - let me know if you prefer it another way.

bfreeds: The reason I kept it in this array, is that I needed it in packages controller to be able to pass that information along to the frontend `applicants` array so the frontend can know to render as `Other Team Member`

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -15,15 +36,20 @@ export const APPLICANT_ATTRIBUTES = [
   'dcp_state',
   'dcp_zipcode',
   'dcp_phone',
+  'target_entity',
```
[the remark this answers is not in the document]
allthesignals: Ah, the way I've had to deal with this is to add it into packages controller. `documents`, for example, are a kind of derived or virtual property that are added mid-query: 

image

I would just add this to packages.controller:
image

With a good comment, I think we can get away with leaving it there — keeping in mind that these kinds of middle layer virtual properties should be handled (or refactored away later). It's also just more consistent with how we're already dealing with this problem. I like to keep those "smells" close together at least so it's a little easier to deal with. 

You can also get rid of those 4 filters you have to do :) ~`.filter(attribute => attribute !== 'target_entity')`~

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -32,38 +58,107 @@ export class ApplicantsController {
...
+            `/dcp_pasforms(${body.pas_form})`,
+          ],
+        });
+      } else {
+        return this.crmService.create(
+          `dcp_applicantinformations`,
+          allowedAttrs,
+        );
+      }
+    } else if (target_entity === 'dcp_applicantrepresentativeinformation') {
```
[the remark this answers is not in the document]
bfreeds: I was writing this thinking about future code maintainers, and was thinking that being explicit about the two branches would reinforce / make more transparent what's happening (vs. implicitly using a default target_entity with an `if` that handles one of the scenarios. 

@allthesignals do you think it would be clearer if the `if` statement was only for applicant representative, and the applicant entity was treated as default / outside of any conditional?

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -15,15 +36,20 @@ export const APPLICANT_ATTRIBUTES = [
...
-  attributes: [
-    ...APPLICANT_ATTRIBUTES,
-  ],
-}))
+@UseInterceptors(
+  new JsonApiSerializeInterceptor('applicants', {
+    id: 'dcp_applicantinformationid',
+    attributes: [...APPLICANT_ATTRIBUTES],
+  }),
+)
```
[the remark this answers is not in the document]
bfreeds: I hear you - I ran prettier to make it easier for me to read and understand as I was working, should have not committed.  In our next tech tooling conversation I'd like to assess code formatters as part of our workflow/tooling.

server/src/packages/pas-form/applicants/applicants.controller.ts
```diff
@@ -32,38 +58,107 @@ export class ApplicantsController {
...
+    if (target_entity === 'dcp_applicantinformation') {
+      // we can't send target_entity to CRM because it doesn't exist there
+      allowedAttrs = pick(body, APPLICANT_ATTRIBUTES.filter(attribute => attribute !== 'target_entity'));
+
+      await this.crmService.update(
+        'dcp_applicantinformations',
+        id,
+        allowedAttrs,
+      );
+    } else if (target_entity === 'dcp_applicantrepresentativeinformation') {
```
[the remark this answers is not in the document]
bfreeds: Same as above - was being a little more verbose to be more explicit for future code readers/maintainers.

server/src/packages/pas-form/applicants/applicants.controller.ts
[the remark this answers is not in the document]
bfreeds: @allthesignals great feedback and agreed this makes a lot more sense.  Moved it into the packages controller, removed the filters.

server/src/packages/pas-form/applicants/applicants.controller.ts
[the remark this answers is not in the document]
allthesignals: The alternative I had in mind was this:

```
if (value === 1) { }
if (value === 2) { }
```

server/src/packages/pas-form/applicants/applicants.controller.ts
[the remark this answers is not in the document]
allthesignals: let's do it! i'd rather it be all one way than halvsies. thanks :)

bfreeds:
> What ideas do you have for future refactorings? 

This PR is somewhat verbose with some repetition in the `if...else` blocks.  One way to refactor would be to parametrize using `target_entity` to supply some of the function arguments.   

The main priority would be looking at the workaround for using a consistent `ref` in the services controller, and how we are mapping the applicant representative id to a common field for maintaining Ember data relationships (in the json api spec).  

I think the way to get around these workarounds and conditional logic in the controller would be to have two ember data models, one for applicants, and one for applicant representatives, and have a base controller for applicants and a representative controller that extends the applicants controller to share the logic that's the same (the data models are nearly identical).  In hoping to treat these two entities as one data model in the frontend, was hoping it would be simpler, but really because CRM treats them as two entities I think it adds a layer of complexity because we kind of obscure that fact.

> Could we get away with this tech debt or will it grow?
If this is the only place/way we use applicants, then I think this will work and focusing on some of the smaller refactorings to make this implementation as clear/readable as possible could keep this maintainable. 

Two tricky things here are taking the two unique ids from CRM entities, and merging them into one unique ID to provide to JSON API Spec, and handling methods that don't provide us with a `body.target_entity` attribute where we have to derive which CRM entity through the custom pre-pended string to the ID.

server/src/packages/pas-form/applicants/applicants.controller.ts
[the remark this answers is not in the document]
bfreeds: @allthesignals if I see two `if`s like that, reading I would think that there would be a scenario where both conditionals are true and get executed.

Vs. in this case, the two conditionals are mutually exclusive so communicating that with the `if... else if` makes it (I think) more declarative.

Issue #210, opened 2020-05-11
Applicant Information Geographic State codes


allthesignals:
These are numeric state codes: 

![image.png](https://images.zenhubusercontent.com/5ca23922d9fcf5278d715977/b84f2541-8649-47d5-9a3e-ba99f6646d90)

We need to correctly serialize these or the request will fail

hannahkates:
Good catch! This must be the cause of #183

hannahkates:
Will be solved by #262

Issue #183, opened 2020-05-06
The geographic State (NY, etc) entered by the applicant disappears after saving the PAS Form 


ksinghprod:
## Browser(s) and Device(s) observed on: 
Chrome, Firefox, IE

## What should happen:
The state entered in the state field by the applicant, should be saved when the applicant clicks on the save button on the PAS form

## What happened instead:

The state entered by the applicant disappears after saving the PAS Form 

## How to reproduce this bug:

1. Log Into Applicant Portal
2. Click on the Edit Pre-Application Statement button
3. Navigate to the Applicant team 
4. Fill out all fields for the applicant team member, including State
5. Click on the Save button (error happens here)

## ADO Test Case
 R10744-AC1-1, step 5 
https://dcp-paperless.visualstudio.com/ZAP%20Portals/_testPlans/define?view=_TestManagement&planId=10799&suiteId=10912

![image](https://user-images.githubusercontent.com/57362229/81218385-1e141180-8fac-11ea-923e-5a680a18c9f9.png)

![image](https://user-images.githubusercontent.com/57362229/81301750-73e9c780-9047-11ea-90c3-06509a44e807.png)

hannahkates:
The cause of this is #210. We had thought it was text, not an option set.

godfreyyeung:
I'm going to look at this

ksinghprod:
This issue is now fixed 
@Hemachekuri 
![image](https://user-images.githubusercontent.com/57362229/82242910-aaa7c380-990c-11ea-9bf4-0f23c7140c0f.png)

PR #262, opened 2020-05-15
183 Patch applicant geographic state

godfreyyeung:
Since the applicant dcpState field is an option set, we
introduce a dropdown for that field so for users can search and
select for a specific state. In the frontend business logic,
we translate the selected state into its respective option set
code and assign the code to the applicant.dcpState field.

Addresses #183 and #210

### DIFF
--- a/client/app/components/packages/applicant-fieldset.js
+++ b/client/app/components/packages/applicant-fieldset.js
@@ -14,7 +14,8 @@ export default class ApplicantFieldset extends Component {
       (accumulator, option) => ({
         [option.code]: option,
         ...accumulator,
-      }), { });
+      }), { },
+    );
   }
 
   // TODO: Refactor this to engage with Changeset, when changesets are
--- a/client/app/components/packages/applicant-fieldset.js
+++ b/client/app/components/packages/applicant-fieldset.js
@@ -1,7 +1,29 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
+import { STATE_OPTION_SET } from '../../models/applicant';
 
 export default class ApplicantFieldset extends Component {
+  get stateOptions() {
+    return Object.values(STATE_OPTION_SET);
+  }
+
+  // TODO: Use a global helper instead after writing
+  // unified optionset handling
+  get stateLabelLookup() {
+    return this.stateOptions.reduce(
+      (accumulator, option) => ({
+        [option.code]: option,
+        ...accumulator,
+      }), { });
+  }
+
+  // TODO: Refactor this to engage with Changeset, when changesets are
+  // introduced for applicant, BBls.
+  @action
+  updateAttrWithOption(currentObject, attr, { code: optionCode }) {
+    this.updateAttr(currentObject, attr, optionCode);
+  }
+
   // syncs radio input toggle to Ember model
   @action
   updateAttr(currentObject, attr, newVal) {
--- a/client/app/components/packages/applicant-fieldset.hbs
+++ b/client/app/components/packages/applicant-fieldset.hbs
@@ -74,7 +74,20 @@
       <label>City <Input type="text" @value={{@applicant.dcpCity}}/></label>
     </div>
     <div class="cell auto medium-2">
-      <label>State <Input type="text" @value={{@applicant.dcpState}}/></label>
+      <label data-test-applicant-state-dropdown>
+        State
+        <PowerSelect
+          supportsDataTestProperties={{true}}
+          @selected={{get this.stateLabelLookup @applicant.dcpState}}
+          @placeholder="state"
+          @searchEnabled={{true}}
+          @searchField="label"
+          @options={{this.stateOptions}}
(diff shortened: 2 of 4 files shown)


client/app/components/packages/applicant-fieldset.js
```diff
@@ -1,7 +1,32 @@
...
+    return lookup;
+  }
+
+  // TODO: Refactor this to engage with Changeset, when changesets are
+  // introduced for applicant, BBls.
+  @action
+  setApplicantState({ code: stateCode }) {
+    this.args.applicant.dcpState = stateCode;
+  }
+
```
allthesignals, line 29: there's already an `updateAttr` method below. Why make a new one?

client/app/components/packages/applicant-fieldset.js
```diff
@@ -1,7 +1,32 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
+import { STATE_OPTION_SET } from '../../models/applicant';
 
 export default class ApplicantFieldset extends Component {
+  get stateOptions() {
+    return Object.values(STATE_OPTION_SET);
+  }
+
+  // TODO: Use a global helper instead after writing
```
allthesignals, line 10: Why not just do it now? It could be a helper called "group-by". Here's how D3 describes it: https://github.com/d3/d3-array#group.

```handlebars
  {{group-by 'code' this.stateOptions}}
```

client/app/components/packages/applicant-fieldset.js
```diff
@@ -1,7 +1,32 @@
...
+  get stateLabelLookup() {
+    const lookup = {};
+    const { stateOptions } = this;
+
+    for (let i = 0; i < stateOptions.length; i += 1) {
+      lookup[stateOptions[i].code] = stateOptions[i];
+    }
+
+    return lookup;
+  }
```
allthesignals, line 21: You might add a comment "key the lookup by its code name"

Are there any other ways to do this? 

Alternative to consider: 
```javascript
  get stateLabelLookup() {
    return this.stateOptions
      .reduce((accumulator, option) => ({
        [option.code]: option,
        ...accumulator,
    }), { });
  }
```

```javascript
  get stateLabelLookup() {
    return this.stateOptions
      .reduce((accumulator, option) => {

        return {
          [option.code]: option,
          ...accumulator,
        }
      }, {});
  }
```

client/app/components/packages/applicant-fieldset.js
allthesignals: Perhaps we could have another helper that looks up optionsets for us:

```handlebars
{{group-by 'code' (lookup-options 'applicants' 'state')}}
```

allthesignals: https://github.com/DockYard/ember-composable-helpers#group-by ?

client/app/components/packages/applicant-fieldset.js
godfreyyeung: Is there a way to destructure the selected object before passing it into `newVal` parameter of `updateAttr`?

allthesignals: I'm not sure — have you looked at the docs? Could also just call `updateAttr` for here.

Not sure `setApplicantState` is the best name. "State" means a lot of different things in different contexts. If you need to pull an attribute from the selected state object, maybe that's what this method could be responsible for?

client/app/components/packages/applicant-fieldset.js
godfreyyeung: Made some changes!

client/app/components/packages/applicant-fieldset.js
godfreyyeung: Made some changes!

client/app/components/packages/applicant-fieldset.js
godfreyyeung: Thinking of a helper with this API, what do you think? 
@trbmcginnis @bfreeds 

**{{optionset 'applicant' 'state'}}** --> returns the optionset object
```js
    { OR: { label: 'OR', code: 12345}, { label: 'MN', code: 123456 } ... } 
```

**{{optionset 'applicant' 'state' 'list'}}** --> returns an optionset options as an array
```js
    [ { label: 'OR', code: 12345} ... }, { label: 'MN', code: 123456 } ... ] 
```

**{{optionset  'applicant' 'state' 'code' 'OR'}}** --> returns code for given label
```
12345
```

**{{optionset  'applicant' 'state' 'label' 12345}}** --> returns label for given code
```
'OR'
```

allthesignals: That looks awesome!

allthesignals:
Merging after tests

client/app/components/packages/applicant-fieldset.js
bfreeds: @godfreyyeung I love your proposed interface for working with changesets -- that would be really pleasant & ergonomic to work with.

Issue #254, opened 2020-05-14
Handle sorting of new actions added to Land Use Actions list


trbmcginnis:
When a user adds a new action, this action should appear at the TOP of the actions list. 

Right now, there are two arrays that are combined to make up this list, so using 'unshiftObject` to push the object to the top of the array would no longer work. When a user inputs a value, "new" actions are shifted to the "existing" actions array, which means they're resorted in the array and therefore on the UI. To avoid this, we should have better sorting methods.

Issue #245, opened 2020-05-14
Revise PAS land use action component


hannahkates:
## User context
From applicant feedback in a recent demo, it was determined that we can reduce confusion for users by eliminating an unnecessary question in the land use action component.

Currently, all actions require a question asking how many of that type of action there are. It turns out it's only possible to have one of most types of actions, so we can safely assume that the answer is one in those cases.

Only three land use actions require the "how many" question:
- Zoning Special Permit, dcp_pfzoningspecialpermit 
- Zoning Authorization, dcp_pfzoningauthorization 
- Zoning Certification, dcp_pfzoningcertification 

### Display and Model Logic
Only Zoning Special Permits, Authorizations, and Certifications require the "how many" question. 

For the rest of the zoning actions, that question should be hidden. When the user adds one of the other actions, the number 1 should be saved to the model for that action.

### Updated wireframe for actions with no questions
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/469df3bc-0b47-499a-9566-d414c2f6b5bb)

hannahkates:
@trbmcginnis I'm assigning you to this since you're already assigned to other land use action issues. I want to make sure you're not fixing validations for some fields that we don't need user validation for anymore.

hannahkates:
@ksinghprod @Hemachekuri Tagging so you're aware of this planned change

ksinghprod:
For the listed 3 selection in the issue, all are working fine 
![image](https://user-images.githubusercontent.com/57362229/82242213-7bdd1d80-990b-11ea-8a0a-79233747fcfc.png)

![image](https://user-images.githubusercontent.com/57362229/82242269-91eade00-990b-11ea-9938-f062cc3e2a26.png)

hannahkates:
Thanks @ksinghprod  should this issue be closed then?

Hemachekuri:
Closing the issue as it is working good.


![image](https://user-images.githubusercontent.com/63460627/82604462-5e65b900-9b82-11ea-82b6-61c42b680183.png)

PR #264, opened 2020-05-15
245 revise land use action component

trbmcginnis:
**Feature Update:**
- Only "Zoning Authorization", "Zoning Certification", and "Zoning Special Permit" now have a count field input. When an action is added from the dropdown, its count field is automatically set to 1. Addresses #245 

**Refactor:**
- Create a new array `actionsAddedByUser` that acts as a reference array for distinguishing new actions that the user has selected from the dropdown.
- Redefine `selectedActions` by filtering by actions that have values set on the model OR exist in `actionsAddedByUser` array. By avoiding combining two different arrays (one "new" and one "existing"), and instead just filtering ONE array we can simplify this computed property and avoid double counting actions that might meet both criteria.

**Sorting:**
- Add sorting logic. Actions should be sorted by: (1) if they're from the db they should be on the bottom of the list and sorted alphabetically. (2) if they have been added by the user from the dropdown they should be sorted on top with most recently added first. Addresses #254

### DIFF
--- a/client/app/components/land-use-action.js
+++ b/client/app/components/land-use-action.js
@@ -61,9 +61,6 @@ export const allLandUseActionOptions = [
 ];
 
 export default class LandUseActionComponent extends Component {
-  // local variable that records singular action result of dropdown selection
-  @tracked selectedAction = {};
-
   // actions added by user in current session (pushed each time user selects from dropdown)
   @tracked actionsAddedByUser = [];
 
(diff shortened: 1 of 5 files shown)


client/app/components/land-use-action.js
```diff
@@ -61,64 +61,65 @@ export const allLandUseActionOptions = [
...
+      const includesA = newActions.includes(a);
+      const includesB = newActions.includes(b);
 
-  get availableLandUseActionOptions() {
-    return allLandUseActionOptions.filter((option) => !this.landUseActionSelections.includes(option));
+      // if actions are both new -- sort most recently added on top
+      if (includesA && includesB) {
+        return newActions.indexOf(a) - newActions.indexOf(b);
+      // if actions are both from db -- sort alphabetically
+      } else if (!includesA && !includesB) { // eslint-disable-line
```
allthesignals, line 91: eslint-disable-line?

client/app/components/land-use-action.js
```diff
@@ -61,64 +61,65 @@ export const allLandUseActionOptions = [
...
+      } return 0;
+    });
   }
 
-  get sortedAvailableLandUseActionOptions() {
-    return this.availableLandUseActionOptions.sort((a, b) => ((a.name > b.name) ? 1 : -1));
+  // actions that appear in dropdown sorted alphabetically
+  get availableActions() {
+    const availableLandUseActions = allLandUseActionOptions.filter((option) => !this.selectedActions.includes(option));
+    return availableLandUseActions.sort((a, b) => ((a.name > b.name) ? 1 : -1));
```
allthesignals, line 106: Just a preference but I think this is a little easier to read: 

```javascript
    return allLandUseActionOptions
      .filter((option) => !this.selectedActions.includes(option))
      .sort((a, b) => ((a.name > b.name) ? 1 : -1));
```

client/app/components/land-use-action.js
```diff
@@ -61,64 +61,65 @@ export const allLandUseActionOptions = [
...
-    const selection = this.selectedLandUseAction.countField;
+  addSelectedAction() {
+    this.args.pasForm[this.selectedAction.countField] = 1;
+    this.actionsAddedByUser.unshiftObject(this.selectedAction);
 
-    if (selection) {
-      this.newLandUseActionSelections.pushObject(this.selectedLandUseAction);
-      this.selectedLandUseAction = {};
-    }
+    this.selectedAction = {};
```
allthesignals, line 114: What's another way of handling selectedAction? What would the code look like if you had the dropdown call `addSelectedAction` directly? I'm wondering — how can we make it to where we don't need to store a reference to the "selectedAction", any state at all.

Any time you store a reference on the class (setting `selectedAction` or something), you have to manage it through setting it back, etc.

PR #268, opened 2020-05-15
203 bread crumbs

allthesignals:
This PR shows the project name in the bread crumbs.

This also shifts the route model logic a little by fetching the package data in the packages route rather than having that happen in the sub routes.


andycochran:
This looks great, @allthesignals! We'll need to refactor later so the segments of the breadcrumbs.hbs are dynamic, based on the route structure, so we can reuse this component on other views (e.g. the project dashboard, non-PAS packages). You concerned with the tiny decrease in codecov?

allthesignals:
> This looks great, @allthesignals! We'll need to refactor later so the segments of the breadcrumbs.hbs are dynamic, based on the route structure, so we can reuse this component on other views (e.g. the project dashboard, non-PAS packages). You concerned with the tiny decrease in codecov?

Yeah, so I think we'll want to do that eventually but not yet sure how that would work. I think this is probably good for now until we need to refactor and build in more behavior.

PR #276, opened 2020-05-15
improve PASForm markup

andycochran:
This PR improves the PAS form, fixing some invalid syntax (paragraphs within labels), using `.help-text` and `.text-small` more semantically, moves the Tax Lot Selector note/link to a more appropriate place in the layout (not inside the `fieldset-adder`, and makes the language more consistent in the Team section (using "Other" instead of "Representative").

image


andycochran:
Atom does that when you save a file. :)

Issue #271, opened 2020-05-15
Save button doesn't detect changes to PAS land use actions


hannahkates:
## What should happen:
Adding, deleting, or modifying a land use action should enable the save button

## What happened instead:
Only making changes to land use action doesn't enable the save button. If you make changes to another field outside of the land use component, the save button will then become enabled.

## How to reproduce this bug:

1. Open a PAS form
2. Only make changes to the land use action section. Add a new action and fill out the required question. Don't touch any other form sections.
3. See that the save button is disabled.
4. Modify a nearby field, like Description of Geography
5. See that save button now becomes enabled.

allthesignals:
@trbmcginnis noticing some weird behavior with this component

trbmcginnis:
Passing in `this.saveableChanges` instead of `this.pasForm` when you render in `edit.hbs` should trigger the save button to detect the changes.

ksinghprod:
This issue is now fixed 
@Hemachekuri 

![image](https://user-images.githubusercontent.com/57362229/82244468-4d614180-990f-11ea-9dbf-55bf34940aaf.png)

After Proposed Land Use Acton Selection 

![image](https://user-images.githubusercontent.com/57362229/82244826-d5dfe200-990f-11ea-8b83-0c4da6e11c9e.png)



![image](https://user-images.githubusercontent.com/57362229/82244871-ebeda280-990f-11ea-89e8-98effd807a13.png)

PR #281, opened 2020-05-18
271: save button on PAS form triggers enabling of save button

trbmcginnis:
Pass in `this.saveableChanges` instead of `this.pasForm` when render landUseAction component in edit.hbs

addresses #271


allthesignals:
oooooooooooooooooooooooh

Issue #275, opened 2020-05-15
Add SUBMIT validations for land use actions


trbmcginnis:
If a count field is present for a land use action then it's associated extra questions MUST be answered to submit. Look in land-use-action.js for references to extra questions.

Issue #257, opened 2020-05-14
In Proposed Land Use Actions, all drop down fields in ADD Proposed Action, such as Change in CityMap should not allow for alphabets, are currently allowing "e"


Hemachekuri:
Browser(s) IE Browser and chrome

What should happen:
In Proposed Land Use Actions, all drop down fields in ADD Proposed Action,  select Change in CityMap user not be able to enter alphabets except "e" , as per DD it is a text feild. It should allow all alphabets and numbers. Number is allowed which is correct, but unable to enter enter alphabets except "e"
What happened instead:
In Proposed Land Use Actions, all drop down fields in ADD Proposed Action,  select Change in CityMap user not be able to enter alphabets except "e" , as per DD it is a text feild. Number is allowed which is correct, but unable to enter enter alphabets except "e"Number is allowed which is correct, but unable to enter enter alphabets except "e"

How to reproduce this bug:
Log Into Applicant Portal
Click on the Edit Pre-Application Statement button
Navigate In Proposed Land Use Actions, all drop down fields in ADD Proposed Action,  select Change in CityMap user not be able to enter alphabets except "e" Number is allowed which is correct, but unable to enter enter alphabets except "e"

![image](https://user-images.githubusercontent.com/63460627/81985458-4af7a280-9604-11ea-8085-e960e370464c.png)

trbmcginnis:
`e` stands for exponent which is why it's allowed in an input type of "number". Will investigate if we can disallow that

allthesignals:
@trbmcginnis can we look at this component? I'm still seeing some issues on develop.

ksinghprod:
This issue is now fixed @Hemachekuri 

![image](https://user-images.githubusercontent.com/57362229/82606578-4d6a7700-9b85-11ea-82c9-0ec999aff6b4.png)
![image](https://user-images.githubusercontent.com/57362229/82606627-6410ce00-9b85-11ea-8b15-de454d3a10ac.png)

Issue #251, opened 2020-05-14
In Proposed Land Use Actions, all drop down fields  in ADD Proposed Action, fields validations with respect to length are not correct as per DD and have validations if extra characters are entered


Hemachekuri:
Browser(s) IE Browser and chrome

What should happen:
In Proposed Land Use Actions, all drop down fields  in ADD Proposed Action, fields validations with respect to length should work as per DD and have validations if extra characters are entered
What happened instead:
In Proposed Land Use Actions, all drop down fields  in ADD Proposed Action, fields validations with respect to length are not correct as per DD and have validations if extra characters are entered
What happened instead:
How to reproduce this bug:
Log Into Applicant Portal
Click on the Edit Pre-Application Statement button
Navigate In Proposed Land Use Actions, all drop down fields  in ADD Proposed Action, fields validations with respect to length are not correct as per DD and have validations if extra characters are entered,

![image](https://user-images.githubusercontent.com/63460627/81976241-3f9d7a80-95f6-11ea-87f9-32c5f8b90153.png)
![image](https://user-images.githubusercontent.com/63460627/81976367-6fe51900-95f6-11ea-8968-51bb9e039749.png)

Proposed   Land Use Actions | Expected Character limit | Actual Limit
-- | -- | --
How many   Change in CityMap actions? | 10 | >10
How many   Site Selection - Public Facility actions? | 10 | >10
How many   Acquisition of Real Property actions? | 10 | >10
How many   Disposition of Real Property actions? | 10 | >10
How many   Revocable Consent actions? | 10 | >10
How many   Landfill actions? | 10 | >10
How many   UDAAP actions? | 10 | >10
How many   URA actions? | 10 | >10
How many   Franchise actions? | 10 | >10
How many   Concession actions? | 10 | >10
  |   | >10
How many   Zoning Special Permit actions? | 10 | >10
Where in   the Zoning Resolution can this action be found? | 250 | >250
Which   sections of the Zoning Resolution does this modify? | 250 | >250
How many   Zoning Authorization actions? | 10 | >10
Where in   the Zoning Resolution can this action be found? | 250 | >250
Which   sections of the Zoning Resolution does this modify? | 250 | >250
How many   Zoning Certification actions? | 10 | >10
Where in   the Zoning Resolution can this action be found? | 250 | >250
Which   sections of the Zoning Resolution does this modify? | 250 | >250
How many   Modification actions? | 10 | >10
[...]

Issue #197, opened 2020-05-07
In Proposed Land Use Actions in zoning Text amendment, in How many Zoning Text amendment actions, user is able to add negative values


Hemachekuri:
Browser(s) IE Browser and chrome

What should happen:
In Proposed Land Use Actions in zoning Text amendment, in How many Zoning Text amendment actions, user should not be able to add negative values
What happened instead:
In Proposed Land Use Actions in zoning Text amendment, in How many Zoning Text amendment actions, user is able to add negative values
How to reproduce this bug:
Log Into Applicant Portal
Click on the Edit Pre-Application Statement button
Navigate to the Proposed Land Use action, and select any item like change in city map.
How many changes in City map is displayed with input text box, and in that input box, In Proposed Land Use Actions in zoning Text amendment, in How many Zoning Text amendment actions, user should not be able to add negative values

Click on the Save button (error happens here)

![image](https://user-images.githubusercontent.com/63460627/81325127-ecf81780-9065-11ea-89c1-78154356da00.png)

![image](https://user-images.githubusercontent.com/63460627/81324904-a3a7c800-9065-11ea-8c2d-a9163d6c39d9.png)

hannahkates:
The Dynamics CRM system allows negative numbers, but we can discourage users from submitting them in the front-end app. @godfreyyeung could this be included in #174?

hannahkates:
This should be considered with #245

trbmcginnis:
We could add a validation for `save` where users are unable to save if this value is below 1. Only relevant for "Zoning Certification", "Zoning Special Permit", and "Zoning Authorization".

Hemachekuri:
Working good. Hence closed
![image](https://user-images.githubusercontent.com/63460627/82350988-008e7100-99ca-11ea-8cea-e484d1a5ff73.png)

PR #282, opened 2020-05-18
197: validate land use action fields for save and submit

trbmcginnis:
Validate land use action fields for `save`:
- length of fields for land use actions

Validate land use action fields for `submit`:
- Zoning Special Permit, Zoning Certification, and Zoning Authorization count field should be greater than or equal to 1 (creation of a new `validateNumberIf` function for Zoning Authorization, Zoning Certification, and Zoning Special Permit)
- if count field for action exists, assure extra questions are answered
- update `validatePresenceIf` to allow for `!==`. This was necessary for checking whether one of the three zoning fields above were not `null` or `undefined` (if a user has selected one of these options from the dropdown menu, the count field will NOT be null or undefined, but a number or an empty string. We don't want validation to happen if a user has not selected the action from the dropdown menu)

Addresses #197, #251, #257, #275 

Will need to rebase/update once changes in #295 are merged in

### DIFF
--- a/client/app/validations/submittable-pas-form.js
+++ b/client/app/validations/submittable-pas-form.js
@@ -1,5 +1,6 @@
 import SaveablePasForm from './saveable-pas-form';
 import validatePresenceIf from '../validators/required-if-selected';
+import validateNumberIf from '../validators/validate-number-if';
 
 export default {
   ...SaveablePasForm,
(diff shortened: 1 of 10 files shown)


client/app/validations/submittable-pas-form.js
```diff
@@ -62,4 +63,130 @@ export default {
       message: 'This field is required',
     }),
   ],
+  dcpZoningauthorizationpursuantto: [
+    validatePresenceIf({
+      presence: true,
+      comparison: 'not',
+      on: 'dcpPfzoningauthorization',
+      withValue: null || undefined,
```
allthesignals, line 71: I think I see what you're doing but this will evaluate immediately when it's defined: 
image

So I don't think it does what you think exactly.

The trick is it'd have to be a function:

```
withValue(something) {
  return something === null || something === undefined;
}
```

The validatePresenceIf code would have to be updated to do this accordingly.

client/app/validators/required-if-selected.js
```diff
@@ -3,12 +3,12 @@ import {
...
 
 export default function validatePresenceIf(options) {
-  const { withValue, on } = options;
+  const { withValue, on, comparison } = options;
 
   return (...args) => {
     const [,,, changes, content] = args;
     const target = Object.prototype.hasOwnProperty.call(changes, on) ? changes[on] : content[on];
-    const hasMatchingWith = target === withValue;
+    const hasMatchingWith = comparison === 'not' ? target !== withValue : target === withValue;
```
allthesignals, line 11: I think if we want to have this behavior we can be a little more creative with the API. What i've seen with some APIs is this:

by default, accept a primitive:
```javascript
{ withValue: 123 }
```

or, accept a function, that's provided some context:

```javascript
{
  withValue: (target) => {
    return target === null || target === undefined;
  }
}
```

In your code here, you would do something like this:

```javascript
let hasMatchingWith;

if (typeof withValue === 'function') {
  hasMatchingWith = withValue(target); // now call it with the derived context
} else {
  hasMatchingWith = target === withValue;
}
```

untested code but this could work.


then get rid of your "comparison" from the API space since this will give you the flexibility you need.

client/app/validators/validate-number-if.js
```diff
@@ -0,0 +1,19 @@
...
+    const target = Object.prototype.hasOwnProperty.call(changes, on) ? changes[on] : content[on];
+    const hasMatchingWith = target !== withoutValue;
+
+    if (hasMatchingWith) {
+      return validateNumber(options)(...args);
+    }
+
+    return true;
+  };
+}
```
allthesignals, line 19: Ah, cool! I wonder if there's a way we could "curry" some generic "hasMatchingWith" function with a given validator:

```javascript
conditionallyValidate(validateNumber, {
  // ...normal options
})
```

allthesignals:
@trbmcginnis you'll need to work with @hannahkates to resolve the merge conflict. I tried and have no idea how to reconcile.

Happy to suggest ways to avoid these sorts of conflicts.

allthesignals:
> Almost there! Have you tried to enter text recently? This component seems responsible:
> 
> https://app.zenhub.com/workspaces/labs-sprint-5e56ebd113bedb2575e694da/issues/nycplanning/labs-applicant-portal/302
> 
> I have a partial fix, but it breaks deletion. I'm looking right now, but want to make you aware.

Let's do this ASAP in a follow up :) Thanks

PR #287, opened 2020-05-18
Create SaveableForm component

allthesignals:
Extracts some general behavior from pas-form/edit so we can re-use that functionality in other types of forms.

```handlebars
<SaveableForm
  @model={{this.dummyModel}}
  @submit={{this.doSubmit}}
  as |saveable-form|
>
  <saveable-form.radio
    @groupValue={{this.saveableChanges.someAttribute}}
  />
  <saveable-form.submitButton
    @isEnabled={{not this.model.isDirty}}
  />
</SaveableForm>
```

### DIFF
--- a/client/app/components/saveable-form.hbs
+++ b/client/app/components/saveable-form.hbs
@@ -1,0 +1,18 @@
+{{yield (hash
+  saveableChanges=this.saveableChanges
+  submittableChanges=this.submittableChanges
+  radio=(component 'radio-button'
+    changed=this.validate
+  )
+
+  saveButton=(component 'packages/save-button'
+    onClick=this.save
+    isEnabled=true
+  )
+
+  submitButton=(component 'packages/save-button'
+    onClick=this.submit
+    isEnabled=true
+    isSubmit=true
+  )
+)}}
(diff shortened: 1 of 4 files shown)


client/app/components/saveable-form.hbs
```diff
@@ -0,0 +1,22 @@
+{{!-- template-lint-disable no-log --}}
```
bfreeds, line 1: Is this supposed to be logging or should it be removed?

client/app/components/saveable-form.js
```diff
@@ -0,0 +1,68 @@
...
+      yield changeset.rollback();
+    } catch (error) {
+      console.log('Save error:', error);
+    }
+  }).withTestWaiter()
+  saveChangeset;
+
+  @action
+  validate() {
+    console.log('asdf');
```
bfreeds, line 65: remove

client/app/components/saveable-form.hbs
allthesignals: I kinda want it to, for helping with any debugging going on.

client/app/components/saveable-form.js
allthesignals: woops. thanks

Issue #231, opened 2020-05-13
In PAS Form,  In Project Description, validation of 2000 characters in the fields are not consistent.


Hemachekuri:
Browser(s) IE Browser and chrome

## What should happen:
validation should be consistent across all feilds in Project Description input box for all feilds.
## What happened instead:
 In Project Description, 

**validation for 2000 characters is working good and displayed and save is disabled**
- Description of proposed CEQR scope

**validation for 2000 characters is not working good and displayed and save is not disabled**
- Description of the proposed development being facilitated by the land use actions, 
- Description of existing land uses and structures in the proposed Project Area and Development Site, 
- Description of land uses and built context in the surrounding area (within 1000 ft)

**user not able to enter more than 2000 characters and no validation is displayed**
- Why is this application being proposed? What is the legal, environmental, or land use background?, 
- What is the land use rationale for all the proposed actions?

## How to reproduce this bug:
Log Into Applicant Portal
Click on the Edit Pre-Application Statement button
Navigate to the In Project Description, 
Description of proposed CEQR scope - validation for 2000 characters is working good and displayed and save is disabled
Description of the proposed development being facilitated by the land use actions, Description of existing land uses and structures in the proposed Project Area and Development Site, Description of land uses and built context in surrounding area (within 1000 ft) - validation for 2000 characters is not working good and displayed and save is not disabled
Why is this application being proposed? What is the legal, environmental, or land use background?, What is the land use rationale for all the proposed actions? - user not able to enter more than 2000 characteres and no validation is displayed
![image](https://user-images.githubusercontent.com/63460627/81830045-924d3880-9509-11ea-8fc7-d5218c7d6997.png)
![image](https://user-images.githubusercontent.com/63460627/81830261-d93b2e00-9509-11ea-81e2-351d2545dcd8.png)

hannahkates:
@Hemachekuri Thanks, I reformatted your description so it's easier to parse out as a checklist. GitHub issues can be formatted using markdown language: https://www.markdownguide.org/cheat-sheet/

**First set of bullet points:**
It sounds like the "Description of proposed CEQR scope" question has no unexpected behavior, correct?

**Second set of bullet points:**
The character limit for these three questions is 3000, not 2000, so I don't think this is a problem. Was 2000 a typo?

**Third set of bullet points:**
I think this is the only thing that needs to be updated. When the user reaches 2000 characters, we need to make sure they get a red "text is too long" message.

[...]

PR #289, opened 2020-05-18
231 Display Text Input validation messages

godfreyyeung:
Fixes validation messages not showing up for Text Input and TextArea fields.
Does so by removing maxlength attributes. This addresses Issue #231.
Later, we can add character count display to improve user experience.

This commit also:
- adds a 'min' attribute to validateLength validators so
their error messages do not show an undefined variable.
- Displays validation messages in the UI from both
saveableChangeset and submittableChangeset
next to respective fields.

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -30,6 +30,9 @@
             @type="text"
             @value={{this.saveableChanges.dcpRevisedprojectname}} data-test-dcprevisedprojectname
           />
+          <ValidationMessage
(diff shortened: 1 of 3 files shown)


godfreyyeung:
It may be slightly unclear to future programmers at first why we mutate saveableChangesets but display validation errors from submittableChangesets in some cases:
image

allthesignals:
> It may be slightly unclear at first to future programmers why we mutate saveableChangesets but display validation errors from submittableChangesets in some cases:
> image

My thinking for this is submittable validations enforces a stricter standard and we'll want the user to know those while they're editing the form.

Those validations also extend validations for saveability validations. 

Thoughts?

Issue #290, opened 2020-05-18
Applicant is able to upload but not save a document on the PAS Form


ksinghprod:
[AB#10532](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10532)

## Browser(s) and Device(s) observed on:
Chrome, IE, Firefox

## What should happen:
User should be able to upload and save documents on the PAS Form, in the Attached Documents Section

## What happened instead:
Applicant is able to upload but not save a document on the PAS Form
## How to reproduce this bug:

1. Log into the Applicant Portal
2. Click on the Edit Pre-Application Statement Button
3. Navigate to the Attached Documents Section 
4. click on the choose files
5. select a file and save (after saving document disappears)

![image](https://user-images.githubusercontent.com/57362229/82243747-1a6a7e00-990e-11ea-9f50-b3b6f5bf93a5.png)

After Saving 

![image](https://user-images.githubusercontent.com/57362229/82243791-2c4c2100-990e-11ea-9eee-e0a0a5f7a455.png)

allthesignals:
it does upload, it just disappears. this is documented in an issue somewhere. can you find it and make sure it's not a dupe?

hannahkates:
It's documented in #229. I'll put this in the duplicates column.

ksinghprod:
After uploading and saving the user needs to refresh the page, still will keep this bug open

Hemachekuri:
working good.
![image](https://user-images.githubusercontent.com/63460627/84941122-e5894c80-b0ae-11ea-85f2-6b358b71b974.png)

Issue #229, opened 2020-05-13
Documents don't appear when navigating from the projects list


allthesignals:
Go to /projects then click a package with documents. If there are already docs on that package, they won't show up. A hard refresh gets them to appear.

I think the attrs being provided in the request to /projects (which includes packages) does not include those documents. This is probably why, but I'm surprised that the forced reload on route transition didn't see those new docs attr...

allthesignals:
@godfreyyeung 

I think this is because we use the deprecated "reload" hook. We got bit because we need to call "reload", but ready doesn't get re-run.

I think the current "transition path" would be to instantiate the file managers in the packages route, probably after awaiting the model in the model hook.

Hemachekuri:
working fine upon refresh.
![image.png](https://images.zenhubusercontent.com/5ebc48df610bc025777641a4/7b941080-d6ba-4a9f-b9b7-225105a7e738)

PR #301, opened 2020-05-19
Refactor how filemanager is invoked; add test case

allthesignals:
Changes how filemanger is instantiated on the package model. This is called manually now instead from the deprecated ready hook.

Includes a test cause simulating the server response.


allthesignals:
Gonna merge but need @godfreyyeung input later :)

hannahkates:
@allthesignals This addresses #290 too right? In addition to #229

PR #303, opened 2020-05-19
Refactor memory use in Land Use Action component

allthesignals:
Move initialization of selectedActions into constructor so it is

### DIFF
--- a/client/app/components/land-use-action.js
+++ b/client/app/components/land-use-action.js
@@ -5,58 +5,112 @@ import { tracked } from '@glimmer/tracking';
 // lookup to match action titles to their associated attributes
 export const All_LAND_USE_ACTION_OPTIONS = [
   {
-    name: 'Acquisition of Real Property', countField: 'dcpPfacquisitionofrealproperty', attr1: '', attr2: '',
+    name: 'Acquisition of Real Property',
+    countField: 'dcpPfacquisitionofrealproperty',
+    attr1: '',
+    attr2: '',
   },
   {
-    name: 'Change in CityMap', countField: 'dcpPfchangeincitymap', attr1: '', attr2: '',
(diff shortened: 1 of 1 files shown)


client/app/components/land-use-action.js
```diff
@@ -99,11 +153,6 @@ export default class LandUseActionComponent extends Component {
       if (includesA && !includesB) {
         return -1;
       }
-
-      // if top action is from db and bottom action is new -- reorder
-      if (!includesA && includesB) {
```
trbmcginnis, line 104: why the removal of this?

allthesignals: i'm trying to understand why codecov is saying this line isn't hit.

i pushed this change to see if codecov was less mad, it seems it is...

just trying to understand why codecov is red, not gonna merge.

allthesignals: @trbmcginnis any theories on why removing these lines actually increases coverage? i don't think they're logically ever hit? is there a case in which they would be hit?

allthesignals: @trbmcginnis i think because we use `unshiftObject` the new item is never actually at the bottom of the list?

trbmcginnis: Oh interesting -- I'm thinking this would be hit in the case of an action that is alphabetically situated higher (e.g. `Change in City Map`) and exists in the original db being compared to an action that is alphabetically situated lower (e.g. `Landfill`) but added by the user and therefore should be sorted ABOVE `Change in City Map`. I'm wondering though if these are sorted correctly by the other conditions before it hits this condition and therefore it's not necessary? idk, I can look into it and try to update tests to include more cases

allthesignals:
oh lolol yes

PR #309, opened 2020-05-19
Update minor styling changes on land-use-action component

trbmcginnis:
Re-add styling changes that got lost in a past merge conflict for `landUseAction` component.

### DIFF
--- a/client/app/components/land-use-action.hbs
+++ b/client/app/components/land-use-action.hbs
@@ -29,17 +29,17 @@
           <div class="shrink cell">
             <label for="middle-label" class="middle">How many {{landUseAction.name}} actions?</label>
           </div>
-        </div>
-        <div class="auto cell">
-          <ActionFormInput 
-            @pasForm={{@pasForm}}
-            @fieldType="number"
-            @length="10"
-            @landUseActionField={{landUseAction.countField}}
-          />
-          <ValidationMessage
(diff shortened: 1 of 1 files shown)


client/app/components/land-use-action.hbs
```diff
@@ -29,17 +29,17 @@
...
+            <ActionFormInput 
+              @pasForm={{@pasForm}}
+              @fieldType="number"
+              @length="10"
+              @landUseActionField={{landUseAction.countField}}
+            />
+            <ValidationMessage
+              @changesetError={{get @submittableChanges.error landUseAction.countField}}
+            />
+          </div>
```
allthesignals, line 42: Hmm do you see the spacing issues in the diff here? It doesn't even look like anything was added. How come there are new spaces added?

trbmcginnis: The ActionFormInput was moved inside another div that was previously above it

trbmcginnis: It's hard to see on the git diff :/ 

Screen Shot 2020-05-19 at 12 15 41 PM

allthesignals: Oh sorry I think my Sourcegraph plugin is going haywire 
image

bfreeds: @allthesignals you can hide whitespace changes here:
image

PR #310, opened 2020-05-19
Improve PAS form show template

andycochran:
This PR fixes up the PAS show template.

### DIFF
--- a/client/app/components/packages/pas-form/show.hbs
+++ b/client/app/components/packages/pas-form/show.hbs
@@ -233,27 +233,9 @@
           Does the proposed Project, or portion thereof, require a NYC DEP Storm Water Construction Permit?
         </strong>
         <br />
-        {{
-          if
-          (eq
-            @package.pasForm.dcpProposedprojectorportionconstruction "717170000"
-          )
-          "Yes"
-        }}
-        {{
-          if
-          (eq
-            @package.pasForm.dcpProposedprojectorportionconstruction "717170001"
-          )
-          "No"
-        }}
-        {{
-          if
-          (eq
-            @package.pasForm.dcpProposedprojectorportionconstruction "717170002"
-          )
-          "Unsure at this time"
-        }}
+        {{if (eq @package.pasForm.dcpProposedprojectorportionconstruction 717170000) "Yes"}}
+        {{if (eq @package.pasForm.dcpProposedprojectorportionconstruction 717170001) "No" }}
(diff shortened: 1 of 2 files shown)


client/app/components/packages/pas-form/show.hbs
```diff
@@ -1,715 +1,551 @@
...
-          if
-          (eq @package.pasForm.dcpLegalstreetfrontage "717170004")
-          "Mapped but not Build"
-        }}
+        <span class="cell large-3">
+          {{if (eq @package.pasForm.dcpLegalstreetfrontage 717170000) "Mapped and Build"}}
+          {{if (eq @package.pasForm.dcpLegalstreetfrontage 717170001) "Private Road"}}
+          {{if (eq @package.pasForm.dcpLegalstreetfrontage 717170002) "Record Street"}}
+          {{if (eq @package.pasForm.dcpLegalstreetfrontage 717170003) "Corporation Counsel Opinion"}}
+          {{if (eq @package.pasForm.dcpLegalstreetfrontage 717170004) "Mapped but not Build"}}
```
allthesignals, line 271: We neeeeeed that code lookup thing @godfreyyeung is designing :)

client/app/components/packages/pas-form/show.hbs
```diff
@@ -1,715 +1,551 @@
...
-          (eq @package.pasForm.dcpProjectareaindustrialbusinesszone false)
-          "No"
-        }}
+        <span class="cell large-3">
+          {{if (eq @package.pasForm.dcpProjectareaindustrialbusinesszone true) "Yes"}}
+          {{if (eq @package.pasForm.dcpProjectareaindustrialbusinesszone false) "No"}}
+        </span>
       </p>
 
       {{#if (eq @package.pasForm.dcpProjectareaindustrialbusinesszone true)}}
```
allthesignals, line 316: I know this is not part of the commit but IDK why we need to see if it's `true`, why not just:

```
{{#if @package.pasForm.dcpProjectareaindustrialbusinesszone}}
{{/if}}
```

allthesignals: ii'd understand if we need to check if something is `false` as opposed to `null`

andycochran: I imagine this was just a result of copy/paste from one that's evaluating codes for "true."

PR #312, opened 2020-05-19
Sets dcp_revisedprojectname to projectname if empty

allthesignals:
This makes changes to the backend to set the pas form 'revised project name' to the associated _project's_ name if it's empty.

### DIFF
--- a/client/app/models/pas-form.js
+++ b/client/app/models/pas-form.js
@@ -45,12 +45,7 @@ export default class PasFormModel extends Model {
 
   @attr('number') dcpProjectblock;
 
-  @attr({
-    defaultValue(pasForm) {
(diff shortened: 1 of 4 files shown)


client/app/models/pas-form.js
```diff
@@ -45,11 +45,7 @@ export default class PasFormModel extends Model {
 
   @attr('number') dcpProjectblock;
 
-  @attr({
-    defaultValue(pasForm) {
-      return pasForm.package.project.dcpProjectname;
-    },
-  })
+  @attr('string')
```
godfreyyeung, line 48: perhaps place it on the same line, like other properties?

allthesignals: Woops, yes. Thx

PR #315, opened 2020-05-19
Display route localName

allthesignals:
Displays route name in breadcrumbs

### DIFF
--- a/client/app/components/breadcrumbs.hbs
+++ b/client/app/components/breadcrumbs.hbs
@@ -12,7 +12,7 @@
       Pre-Application Statement
     </li>
     <li>
-      <span class="show-for-sr">Current: </span> Edit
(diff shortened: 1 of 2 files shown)


allthesignals:
> Nice. This'll fix PAS show/edit.
> 
> Soon we'll still need to figure out how to make the "Pre-Application Statement" segment of the breadcrumb dynamic (and make the component work for other views/routes). Would it'd be simpler if the breadcrumbs were explicitly coded in each template?

Hmm. I think we'd use the parent route's naming for that. I linked this documentation earlier, but basically you can specify route metadata: https://api.emberjs.com/ember/3.18/classes/Route/methods/buildRouteInfoMetadata?anchor=buildRouteInfoMetadata. In there, we could deal with wordsmithing a route's name "Pre-Application Statement" for a route called `pas` for example.

PR #318, opened 2020-05-19
Make file manager “upsert” new documents

allthesignals:
This method is triggered multiple times because it needs to update the reference to documents inside the class.

This change updates-or-creates (“inserts”) a new file manager, depending on if one was already created.

We should get rid of ember-file-upload later so there’s less of a dependency on the service.

@godfreyyeung this is more-or-less a version of the getter you had before! Except. This is in a method.


client/app/models/package.js
```diff
@@ -72,14 +72,18 @@ export const PACKAGE_STATE_CODES = {
...
-    const fileQueue = this.fileQueue.create(this.id);
-
-    this.fileManager = new FileManager(
-      this.id,
-      this.documents,
-      [],
-      fileQueue,
-    );
+    if (this.fileManager) {
+      this.fileManager.existingFiles = this.documents;
```
bfreeds, line 76: @allthesignals I understand we're adding the conditional to check whether a package instance already has a fileManager.  What is the documents attribute doing for us in this model?  I see that it's being used in `pas-form/show.hbs` 

```hbs

    <section class="form-section" id="attached-documents">
      <h3>
        Attached Documents
      </h3>
      {{#each @package.documents as |document|}}
        <p>
          {{document.name}}
        </p>
      {{/each}}
    </section>
```
But I guess I'm not understanding why we need to explicitly set `this.fileManager.existingFiles` to the value of `this.documents`.  Could we just use `this.fileManager.existinFiles` in the `show` template above?

bfreeds: In `components/packages/attachments.hbs`, we do use the `fileManager.existingFiles`:
```hbs
      <ul class="no-bullet">
        {{#each @fileManager.existingFiles as |file idx|}}
          <li
          ...
```
Trying to gain an understanding of how these are used.  Also noting that when we think about refactoring the file management to possibly drop the ember FileQueue add-on, I think we should be careful about the semantic usage of `documents`, `attachments`, and `file`, and just make sure that if we're using different language has meaningful and consistent use.

client/app/services/file-manager.js
```diff
@@ -21,6 +21,8 @@ export default class FileManager {
     this.filesToUpload = filesToUpload; // EmberFileUpload QUEUE Object
   }
 
+  @tracked existingFiles;
```
bfreeds, line 24: Makes sense, to make sure templates/components are reacting to changes in existingFiles (if I'm understanding this change correctly).

client/app/models/package.js
godfreyyeung: > @allthesignals I understand we're adding the conditional to check whether a package instance already has a fileManager. What is the documents attribute doing for us in this model? I see that it's being used in `pas-form/show.hbs`
> 
> ```
>     <section class="form-section" id="attached-documents">
>       <h3>
>         Attached Documents
>       </h3>
>       {{#each @package.documents as |document|}}
>         <p>
>           {{document.name}}
>         </p>
>       {{/each}}
>     </section>
> ```
> 
> But I guess I'm not understanding why we need to explicitly set `this.fileManager.existingFiles` to the value of `this.documents`. Could we just use `this.fileManager.existinFiles` in the `show` template above?

We currently need to insert the Package documents into the existingFiles array, because the Attachments component engages with the existingFiles array in order to track files to marked for deletion

godfreyyeung: A

> In `components/packages/attachments.hbs`, we do use the `fileManager.existingFiles`:
> 
> ```
>       <ul class="no-bullet">
>         {{#each @fileManager.existingFiles as |file idx|}}
>           <li
>           ...
> ```
> 
> Trying to gain an understanding of how these are used. Also noting that when we think about refactoring the file management to possibly drop the ember FileQueue add-on, I think we should be careful about the semantic usage of `documents`, `attachments`, and `file`, and just make sure that if we're using different language has meaningful and consistent use.

Agree, if we drop ember-file-upload, we can refer to everything in all code as "Documents", since that's the CRM term for Documents. If in the UI copy we want to use "Attachments", that's fine.

hannahkates:
@allthesignals which issues will this fix? #290?

client/app/models/package.js
trbmcginnis: Researching more into `ember-file-upload` in issue #325

Issue #258, opened 2020-05-14
Answers to Conditional fields questions on the submitted PAS form, Proposed Land Use Actions field/section are not displaying properly 


ksinghprod:
## Browser(s) and Device(s) observed on:
Chrome, IE, Firefox

## What should happen:
 User should be able to see the question and answers to conditional questions on the PAS Form for all sections on the Proposed Land Use Action section

## What happened instead:
For Example, when the user selects Zoning Certification from the drop down, and answers the conditional questions and submits, the answers to those question is displayed in one line, the conditional question disappears 

## How to reproduce this bug:

1. log into the applicant portal
2. click on the Edt-PAS Statement button
3. Navigate to the Proposed Land Use Actions field/drop down
4. Select Zoning Certification
5. save and submit
6. Navigate back to the Proposed Land Use Actions Field (Error will be seen here) 

NOTE: this is happening for all conditional/follow up questions on the Proposed Land Use Actions field/dropdown

![image](https://user-images.githubusercontent.com/57362229/81989121-a88eed80-960a-11ea-8b22-4a53df3b2a3f.png)

trbmcginnis:
@andycochran Would you want to work on this together?

hannahkates:
@trbmcginnis @andycochran I think a design scan is needed for the entire view-only page.

When you select multiple development type options, similar to the land use action answers, they are displayed in an unformatted list without delimiters which isn't very user friendly.

![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/13967925/82098259-97a9af00-96d2-11ea-914a-09e5e3f34149.png)

hannahkates:
Might make sense to tackle this with #256 since they're in the same view-only template.

andycochran:
Addressed by b4ad95e5b2fe182b31c49ea10433a1b81dfb70f5 and #334.

hannahkates:
@andycochran The development types still don't have a delimiter
![image](https://user-images.githubusercontent.com/13967925/82495556-45dd9c00-9ab9-11ea-94ae-0de6d8416f1f.png)

andycochran:
Thanks for catching this. Working on a PR…

ksinghprod:
This issue is now fixed 
@Hemachekuri 
![image](https://user-images.githubusercontent.com/57362229/82575036-eda8a780-9b55-11ea-9a09-a61d58a99d31.png)

PR #334, opened 2020-05-20
improve logic and layout of PAS show

andycochran:
This PR addresses #258 (changing the conditions for when sub-questions show) and improves the design of the PAS show template.

### DIFF
--- a/client/app/components/packages/pas-form/show.hbs
+++ b/client/app/components/packages/pas-form/show.hbs
@@ -82,142 +82,172 @@
         Proposed Land Use Actions
       </h2>
 
-      <ul>
+      <ul class="no-bullet">
         {{#if @package.pasForm.dcpPfacquisitionofrealproperty}}
-          <li class="medium-margin-bottom">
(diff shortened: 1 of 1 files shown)


client/app/components/packages/pas-form/show.hbs
```diff
@@ -288,7 +318,7 @@
         </span>
       </p>
 
-      {{#if (eq @package.pasForm.dcpLanduseactiontype2 717170000)}}
+      {{#if @package.pasForm.dcpPleaseexplaintypeiienvreview}}
```
allthesignals, line 321: fab! reads much better imo

andycochran: It's different logic. Instead of checking that the parent question's answer makes the input available in the edit view (which is potentially irrelevant here), we just check if the data we wanna show exists.

allthesignals: okay. if it's a boolean and it's false it won't show. btw.

Issue #317, opened 2020-05-19
In Project description, in Description of proposed CEQR scope, number of characters validation is not consistent with other feilds in the section.


Hemachekuri:
## What should happen:
In Description of proposed CEQR scope, when user enters, when you enter only 2000 characters, Text is too long is displayed for but 2000/2000 is not displayed right side bottom text box, but when user enters 3000 characters it is displayed. when user enter 2000 character then 2000/2000 should be displayed like other feilds.
## What happened instead:
In Description of proposed CEQR scope, when user enters, when you enter only 2000 characters, Text is too long is displayed for but 2000/2000 is not displayed right side bottom text box, but when user enters 3000 characters it is displayed. 
## How to reproduce this bug:
Log into the applicant Portal
click on the Edit Pre-Application Statement Button
Navigate to the Project Description
In Description of proposed CEQR scope, when user enters, when you enter only 2000 characters, Text is too long is displayed for but 2000/2000 is not displayed right side bottom text box, but when user enters 30000 characters it is displayed. 

![image](https://user-images.githubusercontent.com/63460627/82382912-d5704580-99fa-11ea-9e39-d81c91e1cb11.png)

![image](https://user-images.githubusercontent.com/63460627/82382757-8d512300-99fa-11ea-9d50-1c6aa8f222c0.png)
hear when you enter only 2000 characters, Text is too long is displayed for but 2000/2000 is not displayed right side bottom text box, but when user enters 30000 characters it is displayed. 
![image](https://user-images.githubusercontent.com/63460627/82382837-b4a7f000-99fa-11ea-97fc-20befe94df41.png)

Issue #252, opened 2020-05-14
The fields lengths on the the PAS Form are too long


ksinghprod:
[AB#10528](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10528)

## Browser(s) and Device(s) observed on:
Chrome, IE, Firefox

## What should happen:

Fields lengths on the PAS form should match the CRM PAS Form and the PAS form Data Dictionary  

## What happened instead:
The Listed field names in the attached spreadsheet are greater than the expected field limits 

## How to reproduce this bug:

1. Log into Applicant Portal
2. Click on the Edit Pre-Application Statement Button
3. Navigate to listed sections, field names on the PAS Form and enter data that is greater than the listed field character limit (error happens here)

[PAS Form field lengths.xlsx](https://github.com/NYCPlanning/labs-applicant-portal/files/4630446/PAS.Form.field.lengths.xlsx)

trbmcginnis:
Fields that have length validations:

- [ ] `dcpRevisedprojectname`

hannahkates:
It looks like this was already resolved in #336. Ready for retest.

PR #333, opened 2020-05-20
326 consistent test selectors

godfreyyeung:
Update all (and add a couple) test selectors that conform to the pattern

```
data-test-[element or form input type]="<dcpModelpropertyname>"
```

Radio inputs follow this pattern:
```
data-test-radio="<dcpModelpropertyname>"
data-test-radio-option="<model property value>"
```

Note that `<dcpModelpropertyname>` represents the CRM-defined property name. We keep CRM's capitalization pattern.

Examples:
```
<Input
  @type="text"
  @value={{saveable-form.saveableChanges.dcpProposeddevelopmentsiteotherexplanation}}
  data-test-input="dcpProposeddevelopmentsiteotherexplanation"
/>
```

```
<Input
  @type="checkbox"
   ....
  data-test-checkbox="dcpProposeddevelopmentsitenewconstruction"
/>
```

```
<saveable-form.radio
  @value={{radio.code}}
  @groupValue={{saveable-form.saveableChanges.dcpHousingunittype}}
  data-test-radio="dcpHousingunittype"
  data-test-radio-option={{radio.label}}
>
```

Issue #313, opened 2020-05-19
PAS: Validations for Applicant Team Editor


trbmcginnis:
[AB#10528](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10528)

Need to create changeset validations and include validation messages in the template

Validations should occur on:

- [x] `dcpFirstname`-----------max length: 50------------type: text
- [x] `dcpLastname`-----------max length: 50------------type: text
- [x] `dcpEmail`---------------max length: 50------------type: text
- [x] `dcpZipcode`-------------max length: 5------------type: number
- [x] `dcpCity`-----------------max length: 80------------type: text
- [x] `dcpPhone`---------------max length: 10------------type: number
- [x] `dcpAddress`-------------max length: 250------------type: text

trbmcginnis:
Completing this issue should also resolve Issue #298

allthesignals:
consider reorganizing validations into pas-specific folders:

validations/pas/saveable-form
validations/pas/submittable-form
validations/pas/applicants

hannahkates:
The data type enforcements aren't done / aren't working on develop.

PR #336, opened 2020-05-20
252: Standardize and add test for PAS form validation

trbmcginnis:
This PR:
- Adds acceptance test to account for ALL validations that occur on the PAS form (although land use actions have their own acceptance test) so that we have them all in one place. Because checking the required validations can be tedious, this test is meant to capture all validations that QA has been checking for and to make sure we have generic tests for when we conduct future refactors of the form (e.g. separating out components). This file can also function as extra documentation. 
- Standardize validation messages on text length
- Adds requirement for `dcpRevisedprojectname` to exist before submitting

Addresses #252 

Also addresses #317 (`dcpProjectattachmentsotherinformation` character counter is now consistent with its maximum length of `2000`)

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -41,7 +41,8 @@
             data-test-input="dcpRevisedprojectname"
           />
           <ValidationMessage
-            @changesetError={{saveable-form.saveableChanges.error.dcpRevisedprojectname}}
+            @changesetError={{saveable-form.submittableChanges.error.dcpRevisedprojectname}}
+            data-test-validation-message="dcpRevisedprojectname"
           />
         </label>
       </section>
@@ -99,6 +100,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpDescriptionofprojectareageography}}
+              data-test-validation-message="dcpDescriptionofprojectgeography"
             />
           </label>
         </section>
@@ -424,6 +426,7 @@
                 />
                 <ValidationMessage
                   @changesetError={{saveable-form.saveableChanges.error.dcpCityregisterfilenumber}}
+                  data-test-validation-message="dcpCityregisterfilenumber"
                 />
               </label>
             {{/if}}
@@ -469,6 +472,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.saveableChanges.error.dcpEstimatedcompletiondate}}
+              data-test-validation-message="dcpEstimatedcompletiondate"
             />
             {{!-- QUESTION: Should this be a date picker? --}}
           </label>
@@ -632,6 +636,7 @@
               </fieldset>
               <ValidationMessage
                 @changesetError={{saveable-form.submittableChanges.error.dcpHousingunittype}}
+                data-test-validation-message="dcpHousingunittype"
               />
             {{/if}}
           </fieldset>
@@ -660,6 +665,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpProjectdescriptionproposeddevelopment}}
+              data-test-validation-message="dcpProjectdescriptionproposeddevelopment"
             />
           </label>
 
@@ -676,6 +682,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpProjectdescriptionbackground}}
+              data-test-validation-message="dcpProjectdescriptionbackground"
             />
           </label>
 
@@ -692,6 +699,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpProjectdescriptionproposedactions}}
+              data-test-validation-message="dcpProjectdescriptionproposedactions"
             />
           </label>
 
@@ -708,6 +716,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpProjectdescriptionproposedarea}}
+              data-test-validation-message="dcpProjectdescriptionproposedarea"
             />
           </label>
 
@@ -724,6 +733,7 @@
             />
             <ValidationMessage
               @changesetError={{saveable-form.submittableChanges.error.dcpProjectdescriptionsurroundingarea}}
+              data-test-validation-message="dcpProjectdescriptionsurroundingarea"
             />
           </label>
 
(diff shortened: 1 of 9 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -89,6 +90,7 @@
             </small>
             <Textarea
               @value={{saveable-form.saveableChanges.dcpDescriptionofprojectareageography}}
+              data-test-dcpDescriptionofprojectgeography
```
godfreyyeung, line 93: If #333  get merged, this will become `data-test-textarea="dcpDescriptionofprojectareageography"`

hannahkates:
@trbmcginnis does this cover #298 too?

trbmcginnis:
> @trbmcginnis does this cover #298 too?

@hannahkates No, it doesn't. #298 will be covered with the #313 implementations since they're on the applicant form (and therefore a different model)

client/tests/acceptance/land-use-action-fields-validate-test.js
```diff
@@ -28,23 +28,23 @@ module('Acceptance | land use action fields validate', function(hooks) {
 
     await selectChoose('[data-test-land-use-action-picker]', 'Zoning Authorization');
 
-    assert.dom('[data-test-validation-message="Number of actions must be greater than 0."]').doesNotExist();
+    assert.dom('[data-test-validation-message="dcpPfzoningauthorization"]').doesNotExist();
 
     await fillIn('[data-test-input="dcpPfzoningauthorization"]', '-1');
 
-    assert.dom('[data-test-validation-message="Number of actions must be greater than 0."]').exists();
+    assert.dom('[data-test-validation-message="dcpPfzoningauthorization"]').hasText('Number of actions must be greater than 0');
```
bfreeds, line 35: This way of writing these validation tests is so clear and declarative.

client/tests/acceptance/land-use-action-fields-validate-test.js
```diff
@@ -53,24 +53,41 @@ module('Acceptance | land use action fields validate', function(hooks) {
       pasForm: this.server.create('pas-form'),
     });
 
+    // representation of input that is 1 over the maximum character limit
+    const maximum = (maxLength, singleCharacter) => singleCharacter.repeat(maxLength) + singleCharacter;
```
bfreeds, line 57: Looking at how you're using this, it seems like the parameters `singleCharacter` is giving you the flexibility of either `string` of `number` inputs.  This might be more declarative in its usage if the interface was, for example:

```js
await fillIn('[data-test-input="dcpZoningauthorizationpursuantto"]', maximum(250, String));
// or for number
await fillIn('[data-test-input="dcpZoningauthorizationpursuantto"]', maximum(250, Number));
```

trbmcginnis: How are you imagining defining `String` and `Number`?

client/tests/acceptance/pas-form-validation-messages-display-test.js
```diff
@@ -0,0 +1,163 @@
...
+  });
+
+  test('Certain fields display both Saveable and Submittable validation errors', async function(assert) {
+    this.server.create('package', {
+      project: this.server.create('project'),
+      pasForm: this.server.create('pas-form'),
+    });
+
+    // representation of input that is 1 over the maximum character limit
+    const maximum = (maxLength, singleCharacter) => singleCharacter.repeat(maxLength) + singleCharacter;
```
bfreeds, line 28: Wonder if there's an easy place to store helper functions used in tests?  This seems like a useful utility function.

client/tests/acceptance/land-use-action-fields-validate-test.js
bfreeds: Was thinking something like this? 
```js
    const maximum = (maxLength, inputType) => {
       if (inputType === String) {
         'a'.repeat(maxLength + 1) 
        } else if ( inputType === Number ) { 
           '5'.repeat(maxLength + 1)
        }
    }
```

trbmcginnis: Okay looks good, will make some changes :)

client/tests/acceptance/pas-form-validation-messages-display-test.js
trbmcginnis: Good point -- I know that we have custom test helpers, will look into if it's worth it to make this one of them!

client/app/validations/submittable-pas-form.js
```diff
@@ -6,6 +9,7 @@ export default {
   ...SaveablePasForm,
   dcpRevisedprojectname: [
     ...SaveablePasForm.dcpRevisedprojectname,
+    validatePresence({ presence: true, message: 'This field is required' }),
```
godfreyyeung, line 12: super tiny nit: line break

trbmcginnis:
> Software development philosophical question: I wonder if testing validations on all fields makes tests _too_ brittle and introduces too much maintenance burden.
> 
> I remember @allthesignals also commenting on this at some point
> 
> Great work Taylor!

That's a really good point and something I've thought about often when writing acceptance tests that might cover many cases. QA was sending in so may validation error issues that I figured it'd be helpful to have a test that covered all of their manual tests they were doing anyway, but tests requiring maintenance/being too brittle is a good argument against it. Maybe we can chat about that philosophical question at standup? Would be curious to hear everyone's thoughts.

client/app/validations/submittable-pas-form.js
bfreeds: @godfreyyeung @allthesignals @trbmcginnis was going to bring this up during our tooling discussion, but what are your thoughts on trying an auto code formatter (prettier) so that we can standardize things like this and not think about them?

bfreeds:
> Software development philosophical question: I wonder if testing validations on all fields makes tests too brittle and introduces too much maintenance burden.

> QA was sending in so may validation error issues that I figured it'd be helpful to have a test that covered all of their manual tests they were doing anyway, but tests requiring maintenance/being too brittle is a good argument against it. 

Agree with both of these, and think that the brittle maintenance burden is something that we should be careful about, but in this case I agree with @trbmcginnis's intuition that this is an observed problem, so addressing it here I think adds resilience and will ultimately save time.

allthesignals:
Yeah it's a good question, I think it comes down to your own experiences updating, fixing, and debugging tests.

Some common pitfalls I run into with addressing testing is not quite knowing what the assertion is trying to check. I think assertion comments can help with that (you can add specific language around a failed assertion and the test runner will display it for you).

Another thing is repeated actions. Are there a lot of repeated clicks going on? Maybe as we build out our pas-form component, we also have test helpers that can more easily fill out, click, and so on.

I do think we'll wanna at least re-organize our application (acceptance) tests into nicer folder names/file names. 

All that said, having this in acceptance testing is most appropriate because it's so _specific_ to the domain problem we're trying to solve. As you drill down into integration/unit, I think we moved across that spectrum of specific-to-general towards general.

PR #337, opened 2020-05-20
add proposedDevelopmentTypes array for PAS show

andycochran:
This PR addresses #258 by adding comma delimiters to the "Proposed development type" list in the PAS show route.


client/app/components/packages/pas-form/show.js
```diff
@@ -0,0 +1,53 @@
+import Component from '@glimmer/component';
+
+export default class PackagesPasFormShowComponent extends Component {
+  get proposedDevelopmentTypes() {
+    const { pasForm } = this.args.package;
+    const arr = [];
+
+    if (pasForm.dcpProposeddevelopmentsitenewconstruction) {
```
godfreyyeung, line 8: Hmm, this style seems a more verbose, hybrid of the way you had previously and the way we discussed during engie standup. My personal preference is one of the more concise versions.
Was there something about one of those styles that didn't work?

allthesignals:
```handlebars
{{#let (array 
  (hash label='Newly constructed buildings' value=package.@pasForm.dcpProposeddevelopmentsitenewconstruction)
  (hash label='Demolition' value=package.@pasForm.dcpProposeddevelopmentsitedemolition)
  (hash label='Alteration' value=package.@pasForm.dcpProposeddevelopmentsiteinfoalteration)
  (hash label='Addition' value=package.@pasForm.dcpProposeddevelopmentsiteinfoaddition)
  (hash label='Change of use' value=package.@pasForm.dcpProposeddevelopmentsitechnageofuse)
  (hash label='Enlargement' value=package.@pasForm.dcpProposeddevelopmentsiteenlargement)
) as |labels|}}
  {{#each (filter-by 'value' labels true) as |key index|}}
    {{get lookup key}} {{if (eq keys.length index) '' ','}}
  {{/each}}

  {{#if @package.pasForm.dcpProposeddevelopmentsiteinfoother}}
  , Other: {{@package.pasForm.dcpProposeddevelopmentsiteotherexplanation}}
  {{/if}}
{{/let}}

```

allthesignals:
Eh, this is the only solution I could come up with. IF we need to start putting other things like this in the component class, we'll want to start encapsulating this stuff inside their own components.

@godfreyyeung @trbmcginnis @bfreeds i don't feel very strongly about this. Thoughts? 

```handlebars
{{#let (array 
  (hash label='Newly constructed buildings' value=@package.pasForm.dcpProposeddevelopmentsitenewconstruction)
  (hash label='Demolition' value=@package.pasForm.dcpProposeddevelopmentsitedemolition)
  (hash label='Alteration' value=@package.pasForm.dcpProposeddevelopmentsiteinfoalteration)
  (hash label='Addition' value=@package.pasForm.dcpProposeddevelopmentsiteinfoaddition)
  (hash label='Change of use' value=@package.pasForm.dcpProposeddevelopmentsitechnageofuse)
  (hash label='Enlargement' value=@package.pasForm.dcpProposeddevelopmentsiteenlargement)
  (hash label=(concat 'Other: ' @package.pasForm.dcpProposeddevelopmentsiteotherexplanation) value=@package.pasForm.dcpProposeddevelopmentsiteinfoother)
) as |labels|}}
  {{join ', ' (map-by 'label' (filter-by 'value' labels true))}}
{{/let}}
```

godfreyyeung:
> Eh, this is the only solution I could come up with. IF we need to start putting other things like this in the component class, we'll want to start encapsulating this stuff inside their own components.
> 
> @godfreyyeung @trbmcginnis @bfreeds i don't feel very strongly about this. Thoughts?
> 
> ```
> {{#let (array 
>   (hash label='Newly constructed buildings' value=@package.pasForm.dcpProposeddevelopmentsitenewconstruction)
>   (hash label='Demolition' value=@package.pasForm.dcpProposeddevelopmentsitedemolition)
>   (hash label='Alteration' value=@package.pasForm.dcpProposeddevelopmentsiteinfoalteration)
>   (hash label='Addition' value=@package.pasForm.dcpProposeddevelopmentsiteinfoaddition)
>   (hash label='Change of use' value=@package.pasForm.dcpProposeddevelopmentsitechnageofuse)
>   (hash label='Enlargement' value=@package.pasForm.dcpProposeddevelopmentsiteenlargement)
>   (hash label=(concat 'Other: ' @package.pasForm.dcpProposeddevelopmentsiteotherexplanation) value=@package.pasForm.dcpProposeddevelopmentsiteinfoother)
> ) as |labels|}}
>   {{join ', ' (map-by 'label' (filter-by 'value' labels true))}}
> {{/let}}
> ```

It looks like a too chunky and bloated piece of business logic to put in the templates.  I'd prefer something like the algorithm you had in https://github.com/NYCPlanning/labs-applicant-portal/pull/337#pullrequestreview-416398069

Or, here's my take (pseudocode):

```models/pas-form.js
get joinedProposedDevelopments() {
  const developmentTypes = [
    { 
        prop: 'dcpProposeddevelopmentsitenewconstruction',
        label: 'Newconstruction',
      },
      { 
        prop: 'dcpProposeddevelopmentsitedemolition',
        label: 'Demolition',
      },
      { 
        prop: 'dcpProposeddevelopmentsiteinfoalteration',
        label: 'Info alteration',
      },
      { 
        prop: 'dcpProposeddevelopmentsiteinfoaddition',
        label: 'Info addition',
      },
      { 
        prop: 'dcpProposeddevelopmentsitechnageofuse',
        label: 'Change of use',
      },
      { 
        prop: 'dcpProposeddevelopmentsiteenlargement',
        label: 'Enlargement',
      },
  ]

  return developmentTypes.map((type) => {
    if(pasForm[type.prop]) {
      return prop.label;
    }
  }).join(',')
}
```

allthesignals:
@godfreyyeung that looks great. Can we split this off into its own component please? Then it's good to go

client/tests/integration/components/packages/pas-form/show-test.js
```diff
@@ -0,0 +1,42 @@
...
+        dcpProposeddevelopmentsitenewconstruction: true,
+        dcpProposeddevelopmentsitedemolition: true,
+        dcpProposeddevelopmentsiteinfoalteration: true,
+        dcpProposeddevelopmentsiteinfoaddition: true,
+        dcpProposeddevelopmentsitechnageofuse: true,
+        dcpProposeddevelopmentsiteenlargement: true,
+        dcpProposeddevelopmentsiteinfoother: true,
+        dcpProposeddevelopmentsiteotherexplanation: 'Esta bien',
+      },
+    });
```
allthesignals, line 21: Why should we duplicate the tests for both components?

client/app/components/packages/pas-form/show.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@glimmer/component';
+
+export default class PackagesPasFormShowComponent extends Component {
+}
```
allthesignals, line 4: Why do we need this file?

client/app/components/proposed-development-types-list.hbs
```diff
@@ -0,0 +1 @@
+{{this.proposedDevelopmentTypes}}
```
allthesignals, line 1: image
It's telling me there's no newline — should we follow this rule?

client/app/components/proposed-development-types-list.js
```diff
@@ -0,0 +1,47 @@
...
+    ];
+
+    const arr = developmentTypes.reduce((proposedTypes, type) => {
+      if (pasForm[type.prop]) {
+        proposedTypes.push(type.label);
+      }
+      return proposedTypes;
+    }, []);
+
+    return arr.join(', ');
```
allthesignals, line 45: Style thing but do you think this is easier to read than just returning `arr`?:

```javascript
    return developmentTypes
      .reduce((proposedTypes, type) => {
          if (pasForm[type.prop]) {
            proposedTypes.push(type.label);
          }
          return proposedTypes;
        }, [])
      .join(', ');
```

godfreyyeung: You're right, I broke it out for debugging but should revert it.

client/app/components/proposed-development-types-list.hbs
godfreyyeung: yes 👍

client/app/components/packages/pas-form/show.js
godfreyyeung: 😬  will delete

client/tests/integration/components/packages/pas-form/show-test.js
godfreyyeung: I was actually really conflicted where to put these tests...

Playing devil's advocate: Would repeating the tests make the components more modular, though? 

Say one day I removed the ProposedDevelopmentTypesList component, however there was still the requirement that the show.hbs page shows a list of proposed development types. Removing the component and its related tests won't mean we have a hole in our tests.

allthesignals: my instinct is that the application test is more important than the rendering test.

allthesignals: ur call just pick 1 :)

godfreyyeung: :) okay 👍

Issue #307, opened 2020-05-19
When a user removes applicants on the PAS Form they are immediately removed in CRM, not queued for deletion upon "Save"


bfreeds:
## What should happen:
When a user clicks to remove an applicant or applicant team member, this record should be queued for deletion, triggering a dirty state on the PAS form.  Then, clicking "Save" should fire the api request to remove the applicant in CRM.

## What happened instead:
As soon as a user clicks the remove button, the applicant gets removed in CRM.
![2020-05-19 11 44 25](https://user-images.githubusercontent.com/5316367/82348124-7264bb80-99c6-11ea-8888-4514ce2e72ae.gif)

## How to reproduce this bug:

1. Open a non-dirty PAS Form with applicants on it
2. Click delete
3. Look at network logs or check CRM to see the applicant removed

allthesignals:
Use store.deleteRecord instead of called destroyRecord directly. there might be other options.

godfreyyeung:
I can take a look at this

PR #340, opened 2020-05-21
307:  Update add and remove applicant state management

bfreeds:
This PR addresses #307.

Previously, if the applicants array was empty we were adding a blank applicant record so we could render an empty applicant fieldset for the user.   This commit removes that to simplify UX and DX.

When a user clicked “Remove Applicant” button, we were using model.destroyRecord(), which would immediately send a DELETE request to the api server.

In this commit, we use model.deleteRecord() instead, which sets model.isDeleted = true.  This queues the record for deletion when the user clicks “Save” (which ultimately will trigger model.save(), where Ember Data makes the api reqeust to DELETE.

### DIFF
--- a/client/tests/acceptance/pas-form-validation-messages-display-test.js
+++ b/client/tests/acceptance/pas-form-validation-messages-display-test.js
@@ -28,8 +28,6 @@ module('Acceptance | pas form validation messages display', function(hooks) {
     await visit('/packages/1/edit');
 
     assert.dom('[data-test-submit-button]').hasAttribute('disabled');
-    assert.dom('[data-test-save-button]').hasNoAttribute('disabled');
-    await click('[data-test-save-button]');
     assert.dom('[data-test-save-button]').hasAttribute('disabled');
 
     // dcpRevisedprojectname (length limits and presence required)
--- a/client/tests/acceptance/user-can-click-package-edit-test.js
+++ b/client/tests/acceptance/user-can-click-package-edit-test.js
@@ -63,15 +63,6 @@ module('Acceptance | user can click package edit', function(hooks) {
 
     await visit('/packages/2/edit');
 
-    // ! We remove the pregenerated applicant here to disable the "save" button.
-    // Clicking "Save" to disable the Save button causes a race condition:
-    // The automated clicks act so fast that the file is marked for deletion
-    // BEFORE the full Save operation completes,
-    // causing the file marked for deletion to be immediately cleared.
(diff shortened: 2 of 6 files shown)


trbmcginnis:
Loving this PR description, super clear!

client/tests/acceptance/pas-form-validation-messages-display-test.js
```diff
@@ -28,8 +28,6 @@ module('Acceptance | pas form validation messages display', function(hooks) {
     await visit('/packages/1/edit');
 
     assert.dom('[data-test-submit-button]').hasAttribute('disabled');
-    assert.dom('[data-test-save-button]').hasNoAttribute('disabled');
```
bfreeds, line 31: @trbmcginnis this test started failing - I think because this ^ was in here to accomodate the beginning dirty state that was caused by having the `addApplicants()` trigggered when applicants was empty.  

If that's not the case, then I'll dig into why this assertion started failing.

client/tests/acceptance/user-can-click-package-edit-test.js
```diff
@@ -63,15 +63,6 @@ module('Acceptance | user can click package edit', function(hooks) {
 
     await visit('/packages/2/edit');
 
-    // ! We remove the pregenerated applicant here to disable the "save" button.
```
bfreeds, line 66: @godfreyyeung I removed this entire comment because we no longer have to deal with this, but let me know if there are parts of this that you think should still be present.

client/tests/acceptance/pas-form-validation-messages-display-test.js
trbmcginnis: Yep I incorporated that "dirty-when-land-on-page-so-need-to-press-save-button" situation to my test! So feel free to edit based on your changes

bfreeds: Sweet - glad this PR lets us remove this weird unintended behavior from our tests :-)

bfreeds:
> were you able to get around the problem you had with the "changed" state?
@allthesignals 
It was seemingly a browser issue that I can no longer replicate.  When I ran it from a private window I didn't see it happening, and after my computer crashed and I started it up again neither my private or regular browser demonstrated the behavior, and tests were/are passing as expected.

When I was seeing it, I added debuggers in the `applicant-team-editor.js` actions to see if any were being invoked that may have changed the state (nothing was invoked from the component).

PR #343, opened 2020-05-21
add CSS overrides so Ember Power Select matches our styles

andycochran:
### Before

(too short, rounded corners, wrong border color)

 

### After

(matches input height, square corners, matching border color)


allthesignals:
Any thoughts on addressing the SassCompiler slowness?

Issue #293, opened 2020-05-18
Clear conditional answer if user changes condition to false


Hemachekuri:
[AB#10528](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10528)

## What should happen:
In Proposed Development Site, when user switched between Yes to NO to Yes, in Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?, previously input data should not be displayed.
## What happened instead:
In Proposed Development Site, when user switched between Yes to NO to Yes, in Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?, previously input data is displayed.
## How to reproduce this bug:

Log in to Applicant Portal
Click on the Edit-Pre Application Statement button
navigate to the Proposed Development site
Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area? input data and save. 
Changed to No  and saved - Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?
switched to Yes   Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?

![image](https://user-images.githubusercontent.com/63460627/82249222-6968e100-9917-11ea-85c2-96bbccbe1cfc.png)

Changed to No  and saved - Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?
![image](https://user-images.githubusercontent.com/63460627/82249271-7ede0b00-9917-11ea-8ee3-bcbb4248c735.png)

switched to Yes   Is the Development Site in an Inclusionary Housing Designated Area/Mandatory Inclusionary Housing Area?
Previously input data Test for Save is displayed.
![image](https://user-images.githubusercontent.com/63460627/82249424-b947a800-9917-11ea-995d-48f72435f8de.png)

allthesignals:
Why?

andycochran:
I think the UI/UX is good as-is. If the user enters text in a conditional input, then accidentally toggles the condition for that question, they don't lose the text they've already entered. 

However, we need to **make sure that the data for the conditional input is not sent to CRM**. We don't want the client sending bad data. It could be misleading to have answers to conditional questions in the data, if the condition for them is false. 

[...]

PR #349, opened 2020-05-22
293: reset conditional input value when radio button set to no

trbmcginnis:
When a user enters text into a conditional input (which only displays when the associated radio button is set to Yes or checkbox is `checked`), then clicks the No radio button or the checkbox again -- the text value is reset to empty string or null, so that text is not saved to the model.

Setting a conditional `@value` on the input (through an inline `if` statement) did not work. So instead, an action was defined on the radio buttons (`@checked`) and on the checkboxes (`click`) to set the targeted input value to an empty string or null anytime that the radio button/checkbox was selected.

Addresses #293

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -167,6 +167,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpUrbanrenewalarea}}
+                @changed={{fn (mut saveable-form.saveableChanges.dcpUrbanareaname) ""}}
                  data-test-radio="dcpUrbanrenewalarea"
                  data-test-radio-option={{radio.label}}
               >
@@ -238,6 +239,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpLanduseactiontype2}}
+                @changed={{fn (mut saveable-form.saveableChanges.dcpPleaseexplaintypeiienvreview) ""}}
                 data-test-radio="dcpLanduseactiontype2"
                 data-test-radio-option={{radio.label}}
               >
@@ -288,6 +290,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpProjectareaindustrialbusinesszone}}
+                @changed={{fn (mut saveable-form.saveableChanges.dcpProjectareaindutrialzonename) ""}}
                 data-test-radio="dcpProjectareaindustrialbusinesszone"
                 data-test-radio-option={{radio.label}}
               >
@@ -325,6 +328,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpIsprojectarealandmark}}
+                @changed={{fn (mut saveable-form.saveableChanges.dcpProjectarealandmarkname) ""}}
                 data-test-radio="dcpIsprojectarealandmark"
                 data-test-radio-option={{radio.label}}
               >
@@ -402,6 +406,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpRestrictivedeclaration}}
+                @changed={{fn (mut saveable-form.saveableChanges.dcpCityregisterfilenumber) null}}
                 data-test-radio="dcpRestrictivedeclaration"
                 data-test-radio-option={{radio.label}}
               >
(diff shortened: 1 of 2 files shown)


client/tests/acceptance/user-can-click-package-edit-test.js
```diff
@@ -381,4 +381,52 @@ module('Acceptance | user can click package edit', function(hooks) {
 
     assert.dom('[data-test-validation-message="dcpProjectarealandmarkname"]').hasText('Name is too long (max 250 characters)');
   });
+
+  test('For input dependent on radio button/checkbox -- when user fills out input, then clicks radio button/checkbox again, text is not saved to model', async function(assert) {
```
godfreyyeung, line 385: nit: "input, then selects radio option that hides input..."

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -286,6 +288,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpProjectareaindustrialbusinesszone}}
+                @changed={{fn this.updateAttr saveable-form.saveableChanges "dcpProjectareaindutrialzonename" ""}}
```
godfreyyeung, line 291: This is a clever solution to use the `changed` property and set the value to null on change!

godfreyyeung:
Another consideration: From a UX perspective, it might be nice for the user to not lose their input if they switch between radio options. We could hold their answer in memory instead of wiping it, and only submit it to the backend if 

One possibility is to write a function that iterates through the pasForm properties _right before_ save, and null out text input values based on the true/false values of the radio answers.

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -532,6 +537,7 @@
               <li>
                 <Input
                   @type="checkbox"
+                  {{on "click" (fn this.updateAttr saveable-form.saveableChanges "dcpProposeddevelopmentsiteotherexplanation" "")}}
```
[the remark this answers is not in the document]
trbmcginnis: It does -- that was a design question that was brought up when we were talking about the issue. Whether that input should persist when a user clicks on the checkbox/radio button and then clicks back. Looks like QA wanted the text to disappear BUT if we want it to remain I'd be open to chatting about that as a design issue.

trbmcginnis:
> Another consideration: From a UX perspective, it might be nice for the user to not lose their input if they switch between radio options. We could hold their answer in memory instead of wiping it, and only submit it to the backend if
> 
> One possibility is to write a function that iterates through the pasForm properties _right before_ save, and null out text input values based on the true/false values of the radio answers.

That's a good point! @andycochran from a design perspective how do you feel about this issue? I built this change under the assumption that we wanted the input text to disappear when a user re-clicked the "Yes" radio button BUT if we want that text to persist, then we can consider this route instead.

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -606,6 +613,7 @@
               <saveable-form.radio
                 @value={{radio.code}}
                 @groupValue={{saveable-form.saveableChanges.dcpDiscressionaryfundingforffordablehousing}}
+                @changed={{fn this.updateAttr saveable-form.saveableChanges "dcpHousingunittype" ""}}
```
allthesignals, line 616: why not use `mut`? i really want to avoid this style of using string references to key names.

allthesignals: have you considered other options, too? why is this the best path? what about:

making a new component
changing the existing component
using built-in helpers for managing state

do you think this approach will harm future maintainability?

trbmcginnis: @allthesignals I updated the radio buttons in `edit.hbs` to use the `mut` helper thanks for the comment!

As for creating editing existing components (either the `radio-button` or `saveable-form` components), after looking over it a bit I don't think that making that change would simplify this implementation. We would still be required to pass in extra parameters (e.g. `@attributeToReset`) which might complicate the rendering of the radio button -- by either changing what is essentially source code in the `radio-button` component or needing to pass in an extra parameter when we render the `saveable-form.radio` component. The way that the `saveable-form.radio` button is rendered now in `edit.hbs`, there is no input attribute passed in, so that would have to be something new we'd add (essentially adding an extra line of code like we're already doing with the @changed parameter). 

As for creating a contextual component and using it to wrap the radio button and then creating a function that helps to reset that input value --> I think this would present the same scenario as editing existing components. We would still have to pass in an extra parameter to the component, and I think creating an entirely knew contextual component just to handle what could be handled in the template by a mut helper might over-complicate things / cause confusion for other developers trying to decipher the code.

I know that we were trying to pull things out of `edit.hbs` and really simplify that template but I think using the mut helper with the `@changed` parameter is the simplest way to go about it. 

Let me know if you agree/disagree with any of this! ^^ Happy to chat :)

allthesignals: > As for creating a contextual component and using it to wrap the radio button and then creating a function that helps to reset that input value --> I think this would present the same scenario as editing existing components. We would still have to pass in an extra parameter to the component, and I think creating an entirely knew contextual component just to handle what could be handled in the template by a mut helper might over-complicate things / cause confusion for other developers trying to decipher the code.

Great! This sounds like very sound reasoning. Thanks for looking deeper :)

trbmcginnis: Thanks for challenging me to think about all the options!

PR #353, opened 2020-05-22
Ac custom transitions

andycochran:
This PR adds CSS-only animations that make it clear to the user that an element has been inserted into the DOM — specifically, adding Team Members, BBLs, Land Use Actions, and Attachments.

### DIFF
--- a/client/app/components/land-use-action.hbs
+++ b/client/app/components/land-use-action.hbs
@@ -16,10 +16,12 @@
   </label>
 </div>
 {{#if this.sortedSelectedActions}}
-  <h5>Land Use Actions Included in This Project</h5>
+  <h5 {{animate-with 'slide-in-top' 'slide-in-top-active'}}>
+    Land Use Actions Included in This Project
+  </h5>
 {{/if}}
 {{#each this.sortedSelectedActions as |landUseAction|}}
-  <fieldset class="fieldset relative">
+  <fieldset class="fieldset relative" {{animate-with 'scale-fade-in' 'scale-fade-in-active'}}>
     <legend>
       <strong data-test-action-name="{{landUseAction.name}}">{{landUseAction.name}}</strong>
     </legend>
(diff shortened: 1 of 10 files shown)


client/app/components/land-use-action.js
```diff
@@ -181,4 +181,20 @@ export default class LandUseActionComponent extends Component {
...
+    }, 10);
+  }
+
+  @action
+  slideInTop(el) {
+    el.classList.add('slide-in-top');
+    setTimeout(() => {
+      el.classList.add('slide-in-top-active');
+    }, 10);
+  }
```
allthesignals, line 199: can these be helpers instead of having to change the component itself? is this the best way to do this? does ember-animated have anything for us that might be useful: https://ember-animation.github.io/ember-animated/

andycochran: Exactly what I was thinking. I tried to make helpers from the actions but failed at that. Would love to pair on it. It'd be nice to keep the code DRY and just import the specific transitions you need for the component. 

I explained at Engie Standup this AM how I looked at both `ember-css-transitions` and `ember-animated` and that they both have caveats which I think outweigh any risk maintenance with this custom method. Glad to recap that for you, but to summarize: 

- `ember-css-transitions` is very flexible with the CSS transitions, as you write them yourself. However, it interferes with the way that elements are inserted into the DOM, which makes testing difficult. 
- `ember-animated` doesn't break the tests, but it is pretty opinionated about its styles. You can't combine its OOTB styles (e.g. fade in & slide from top). It allows for custom styles, but the API for creating them is complicated, messy JS.

allthesignals: yeah, e-animated uses generator functions and seems scary.

i wonder if we could look into custom modifiers? https://github.com/ember-modifier/ember-modifier#functional-modifiers

thanks for the thoroughness :) i think you're right that helpers would not work because I dn't believe they get a reference to the actual element. custom modifiers do.

PR #372, opened 2020-05-29
290: assure that uploaded files show up in existing files list after save

### LINKED ISSUE #290: Applicant is able to upload but not save a document on the PAS Form

trbmcginnis:
Previously, because `existingFiles` is only loaded once when the fileManager is instantiated, the `existingFiles` array was not being updated (and therefore not displaying) the recently uploaded files after the user had clicked save. They would only display when a user refreshed. This PR assures that the uploaded file objects are pushed to the `existingFiles` array so that they show in the UI. 

Screen Shot 2020-05-29 at 12 49 38 AM

`pas-form-error` component was also updated to account for package errors, since `documents` exist as an attribute on the package model rather than the PAS Form model.

Closes #290

### DIFF
--- a/client/app/components/packages/pas-form-error.hbs
+++ b/client/app/components/packages/pas-form-error.hbs
@@ -1,4 +1,4 @@
-{{#if @pasForm.adapterError}}
+{{#if (or @package.pasForm.adapterError @package.adapterError)}}
   <div class="callout alert" data-test-error-saving>
     <h5 class="text-red-dark">
       <FaIcon
@@ -8,8 +8,11 @@
       />
         There was a problem saving your information.
     </h5>
-      {{#each @pasForm.adapterError.errors as |error|}}
-        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.details}}</p>
+      {{#each @package.pasForm.adapterError.errors as |error|}}
+        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.detail}}</p>
+      {{/each}}
+      {{#each @package.adapterError.errors as |error|}}
+        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.detail}}</p>
       {{/each}}
       <p class="text-red-dark text-small">
         Please try again. If the problem persists, please contact City Planning at <a href="mailto:ZAP_feedback_DL@planning.nyc.gov">ZAP_feedback_DL@planning.nyc.gov</a> or <a href="tel:212-720-3300" class="nowrap">212-720-3300</a>.
--- a/client/app/components/packages/pas-form/edit.js
+++ b/client/app/components/packages/pas-form/edit.js
@@ -30,7 +30,14 @@ export default class PasFormComponent extends Component {
 
   @action
   async savePackage() {
-    await this.args.package.save();
+    try {
+      await this.args.package.save();
+      await this.args.package.reload();
+    } catch (error) {
+      console.log('Save package error:', error);
+    }
+
+    this.args.package.refreshExistingDocuments();
   }
 
   @action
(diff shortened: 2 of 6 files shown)


client/app/services/file-manager.js
```diff
@@ -59,6 +54,7 @@ export default class FileManager {
   }
 
   uploadFiles() {
+    this.filesToUpload.files.forEach((file) => this.existingFiles.pushObject(file));
```
allthesignals, line 57: Nice!

What if the file uploads fail? What if we `await` the Promse.all on the line below, wrapped in a try / catch, then say:

```
this.existingFiles.pushObjects(this.filesToUpload.files)
```

godfreyyeung, line 57: The files won't have an actual link here, though. They don't look like these will be actual "uploaded files"

godfreyyeung: Jotting pseudocode down from our engineering standup:

models/package.js
```
  refreshFileQueue() {
    this.fileManager.existingFiles = this.documents;
  }
```

components/packages/pas-form/edit.js
```
  @action
  async savePackage() {
    await this.args.package.save();
    await this.args.package.reload();
    this.args.package.refreshFileQueue();
  }
```

trbmcginnis: Thanks, I'll try that!

godfreyyeung: The `refreshFileQueue` name was totally wrong... since it's refreshing existing documents. 🙈  I updated it to another possibility. This is all just draft. You have the final call!

client/app/components/packages/pas-form-error.hbs
```diff
@@ -8,8 +8,11 @@
       />
         There was a problem saving your information.
     </h5>
-      {{#each @pasForm.adapterError.errors as |error|}}
-        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.details}}</p>
+      {{#each @package.pasForm.adapterError.errors as |error|}}
+        <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>{{error.detail}}</p>
+      {{/each}}
```
allthesignals, line 13: Good

client/app/components/packages/pas-form/edit.js
```diff
@@ -30,7 +30,14 @@ export default class PasFormComponent extends Component {
...
   async savePackage() {
-    await this.args.package.save();
+    try {
+      await this.args.package.save();
+      await this.args.package.reload();
+    } catch (error) {
+      console.log('Save package error:', error);
+    }
+
+    this.args.package.refreshExistingDocuments();
```
allthesignals, line 40: I think this works for now, but I'm beginning to think that these documents should be their own model now. This line here is a sign that we're sort of having to manually manage the state of documents. Right? For example, when we have docs queued up for upload, those are "dirty" in a way, we can show those as such. When they save, their "clean", they've saved, we can show those accordingly.

I think a refactor task should be to seriously re-think doc upload and embrace Ember Data for them.

trbmcginnis: That makes a lot of sense -- I'm going to go ahead and merge this for the sake of the end of the sprint and then I'll be sure to make an issue addressing this, thanks!

PR #375, opened 2020-05-29
311 Use revision sharepoint folder for documents

godfreyyeung:
This commit updates the document upload and retrieval
sections of the backend (document controller and package
service, respectively) to reference Sharepoint document
folders created automatically by the package revision loop.
Those folders use a slightly different naming convention than
the default folders used previously.

Package Revision folders:
```
<formatted package name>_<unformatted package id>
```

Previous/default Package folders:
```
<unformatted package name>_<formatted package id>
```

As long as the client+server uploads and retrieves documents to
packages starting from the very first revision,
any given package revision folder will hold all documents uploaded
to the revision itself and all previous revisions. This is because
CRM automatically carries over documents from previous revisions
into the folder for the latest revision.
_____

I've done manual testing on this for both packages already with multiple revisions, and packages on the very first revision.

### DIFF
--- a/server/src/document/document.controller.ts
+++ b/server/src/document/document.controller.ts
@@ -54,7 +54,16 @@ export class DocumentController {
 
     const { records: [ { dcp_name: packageName }]} = packageRecord;
 
-    const folderName = `${packageName}_${instanceId.replace(/\-/g,'').toUpperCase()}`;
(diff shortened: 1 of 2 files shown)


godfreyyeung:
> I think it'd be good to comment up these parts of the code to explain what's going on

will do!

PR #388, opened 2020-06-04
Change transitions to animations

andycochran:
This PR refactors the CSS transitions to use animations instead. Transitions are meant to be used to show the change from one state to another, where animations do not rely on state changing. This means we don't need the `animate-with` modifier, since we don't need timeouts that add classes after a delay. Animating an element into the DOM is as simple as giving it a class.


andycochran:
I think the codecov fail is only because there are fewer total lines being hit because I removed JS. But there aren't more misses, so this shouldn't be a problem.

PR #390, opened 2020-06-08
386 load rwcds or pas based on packageType

godfreyyeung:


andycochran:
Separate endpoints for each type of package might make sense. Various packages may conceptually be the same thing, but each one is an edge case rife with conditionals. Would it be possible to isolate the commonalities among all packages in one file and have each individual package extend that base class with the specifics of the package type?

godfreyyeung:
> Separate endpoints for each type of package might make sense. Various packages may conceptually be the same thing, but each one is an edge case rife with conditionals. Would it be possible to isolate the commonalities among all packages in one file and have each individual package extend that base class with the specifics of the package type?

Interesting idea. This might be possible with the controller. Will have to check if it conflicts with Nest paradigms somehow. Extending and then customizing services and JSON API Serializer for each form will be tricky.

server/src/packages/packages.controller.ts
```diff
@@ -92,30 +134,35 @@ export const PASFORM_PROJECTADDRESS_ATTRS = [
...
 
   // Transform here should only be used for remapping
   // navigation links into cleaner names as well as
   // handling special virtual properties that do not
   // come from CRM 
   transform(projectPackage) {
-    const { dcp_pasform: pasForm } = projectPackage;
+    const {
+      dcp_pasform: pasForm,
+      dcp_rwcdsform: rwcdsForm
```
godfreyyeung, line 162: Add code comments about making this code more sustainable

server/src/packages/packages.service.ts
```diff
@@ -35,21 +98,40 @@ export class PackagesService {
...
       dcp_packageid,
       dcp_name,
     } = firstPackage;
 
-    const { records: [projectPackageForm] } = await this.crmService.get('dcp_pasforms', `
+    let targetForm = '';
+
+    if (dcp_packagetype === PACKAGE_TYPE_CODES['PAS_PACKAGE'].code) {
+      targetForm = 'pas';
+    } else if (dcp_packagetype === PACKAGE_TYPE_CODES['RWCDS'].code) {
```
godfreyyeung, line 111: Duplicating code here is fine for better readability. Avoid using targetForm

server/src/packages/packages.controller.ts
```diff
@@ -54,6 +54,47 @@ export const PASFORM_PROJECTADDRESS_ATTRS = [
   'statecode',
 ]
 
+export const RWCDSFORM_ATTRS = [
```
allthesignals, line 57: Should probably be a space `RWCDS_FORM_ATTRS` because that's how `PAS_FORM` is named

PR #392, opened 2020-06-08
373 Restructure routes and create RWCDS route and model

godfreyyeung:
## TL;DR
This PR
1) Changes route structure so that each form has its own route.
2) Introduces the RWCDS Form model
---
## Details
Previously, the route structure was
```
/packages/package_id
```
Now, the route structure is:
```
/pas-form/package_id
/rwcds-form/package_id
```

We make this change with the assumption that each package is only associated with one form.
Assigning each form its own route allows each form route to customize what it sideloads with the `include` property during DataStore queries. 

Technically, the `include` property doesn't matter since our backend API ignores it anyway, and uses the package's `packageType` to determine what to sideload. However, this makes the frontend code cleaner. 

The RWCDS Form model has a 1:1 relationship with the Package model.


andycochran:
I'd suggest that the word "form" is unnecessary. 

Instead of: 
```
/pas-form/package_id
/rwcds-form/package_id
```

… the new route structure could just be: 
```
/pas/package_id
/rwcds/package_id
```

andycochran:
Looks like the path for rwcds might already be as I suggested: 

image

godfreyyeung:
> Looks like the path for rwcds might already be as I suggested:
> 
> image

Oh, good catch, I forgot to update the path to `rwcds-form`. 
Hmm, I think I added `-form` to the routes to match the model naming.
We could get rid of `-form`, but we should be consistent with the model names.
It would mean also getting rid of `-form` throughout the backend, which i think would be too much overhead

client/app/routes/pas-form.js
```diff
@@ -1,18 +1,18 @@
 import Route from '@ember/routing/route';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default class PackagesRoute extends Route.extend(AuthenticatedRouteMixin) {
+export default class PasFormRoute extends Route.extend(AuthenticatedRouteMixin) {
   authenticationRoute = '/';
 
   async model(params) {
-    const projectPackage = await this.store.findRecord('package', params.id, {
+    const pasFormPackage = await this.store.findRecord('package', params.id, {
```
trbmcginnis, line 8: Love this naming, makes it much clearer what's going on

PR #397, opened 2020-06-09
Curry click save/submit handlers inside task

allthesignals:
Save button is now responsible for dealing with the tasks. To make it clearer how the form works, the onClick callback is a first-class API. Underneath, it handles triggering of the saveableForm `saveChangeset` async function (no longer a task).

### DIFF
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -1,8 +1,6 @@
 <SaveableForm
   @model={{this.pasForm}}
   @validators={{array this.saveablePasFormValidations this.submittablePasFormValidations}}
-  @onSave={{this.savePackage}}
-  @onSubmit={{this.submitPackage}}
   as |saveable-form|
 >
 
@@ -762,6 +760,7 @@
       <div class="sticky-sidebar">
         <saveable-form.saveButton
           @isEnabled={{or @package.isDirty (and saveable-form.saveableChanges.isDirty saveable-form.saveableChanges.isValid)}}
+          @onClick={{this.savePackage}}
           data-test-save-button
         />
         {{!-- Error Handling --}}
@@ -793,6 +792,7 @@
                 </div>
                 <div class="cell large-6 small-padding-left">
                   <saveable-form.submitButton
+                    @onClick={{this.submitPackage}}
                     @isEnabled={{saveable-form.submittableChanges.isValid}}
                     class="no-margin"
                     data-test-confirm-submit-button
--- a/client/app/components/packages/save-button.js
+++ b/client/app/components/packages/save-button.js
@@ -19,7 +19,7 @@ export default class SaveButtonComponent extends Component {
 
   // triggered when the button is clicked
   @task(function* () {
-    yield this.args.onClick();
-  })
+    yield this.args._saveChangeset(this.args.onClick);
+  }).withTestWaiter()
   clickTask;
 }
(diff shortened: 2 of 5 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -762,6 +760,7 @@
       <div class="sticky-sidebar">
         <saveable-form.saveButton
           @isEnabled={{or @package.isDirty (and saveable-form.saveableChanges.isDirty saveable-form.saveableChanges.isValid)}}
+          @onClick={{this.savePackage}}
```
trbmcginnis, line 763: Much clearer having this here than having it at the top of the file

client/app/components/saveable-form.js
```diff
@@ -39,26 +38,15 @@ export default class SaveableFormComponent extends Component {
...
-    await this.saveChangeset.perform(this.args.onSave, this.saveableChanges);
-  }
-
-  @action
-  async submit() {
-    await this.saveChangeset.perform(this.args.onSubmit, this.submittableChanges);
-  }
-
-  @task(function* (callback, changeset) {
+  async saveChangeset(changeset, callback) {
```
trbmcginnis, line 41: Looks like we pass in `this.saveableChanges` to the callback parameter. Can you explain why the parameter is called callback?

client/app/components/packages/save-button.js
```diff
@@ -19,7 +19,7 @@ export default class SaveButtonComponent extends Component {
 
   // triggered when the button is clicked
   @task(function* () {
-    yield this.args.onClick();
-  })
+    yield this.args._saveChangeset(this.args.onClick);
```
trbmcginnis, line 22: Does this assume that if we're reusing the `save-button` component in different forms that every attribute that's saveable is also associated with a changeset? Just thinking about attributes that don't have any validation requirements on them for saving or submitting. I know then we'd probably just pass in the model rather than `saveableForm.saveableChanges.attribute_name`, but if this action is stored in the `save-button` component are we always expecting a changeset?

client/app/components/saveable-form.js
allthesignals: dashingd3js.com/lessons/javascript-callback-functions#:~:text=A%20JavaScript%20Callback%20Function%20is,be%20used%20synchronously%20or%20asynchronously

allthesignals: https://en.wikipedia.org/wiki/Callback_(computer_programming)

allthesignals: should it be some other name?

client/app/components/saveable-form.hbs
```diff
@@ -6,12 +6,12 @@
   )
 
   saveButton=(component 'packages/save-button'
-    onClick=this.save
+    _saveChangeset=(fn this.saveChangeset this.saveableChanges)
```
godfreyyeung, line 9: That's odd that the underscore indicate `_saveChangeset` is an internal property, but it's being referenced as a public parameter.

client/app/components/packages/save-button.js
allthesignals: Yes you're right, but here I'm making save-button very much a dependency on saveable-form. So that's the tradeoff: save-button because something more specific to saveable-form.

I'll try to think of a way to "curry" this in a more agnostic way... 

Great question. It may just be that the naming is wrong, and should not refer to changeset. It behaves as an optional wrapper function around the public onClick. 

Going to think about it some more...

client/app/components/saveable-form.hbs
allthesignals: It's private to saveable-form

PR #399, opened 2020-06-09
Ac breadcrumbs

andycochran:
This PR refactors the breadcrumbs so that they'll work with any routes or model. 

Previously, we had the breadcrumbs hardcoded in the projects template, and used a component that only worked for the PAS form. 

The proposed changes allow the `<Breadcrumbs />` component to be call from any template with a contextual component used for its list items. This lets us nix `breadcrumbs.js`, as we no longer rely on the router service for comparing the current route. This also adds breadcrumbs to the appropriate templates, and adds tests to make sure we render breadcrumbs correctly. 

Other methods explored (brainstormed w/ @godfreyyeung): 
- We attempted to simply pass the model to `<Breadcrumbs />` and let the component create the list items. However, this involved complex logic to dynamically show/hide parts of the breadcrumb trail and would be further complicated with each form or route that needs breadcrumbs. That logic also made assumptions about the data being passed to the component — e.g. the projects/project models will have a different schema than the packages (which also might have differing schemas). 
- Just hardcoding the breadcrumbs in each template, since they're rather simple. However, the `breadcrumbs-item.js` slims the markup significantly.

### DIFF
--- a/client/app/components/breadcrumbs-item.hbs
+++ b/client/app/components/breadcrumbs-item.hbs
@@ -1,0 +1,12 @@
+<li class="{{if @disabled 'disabled'}}">
+  {{#if @current}}
+    <span class="show-for-sr">Current:</span>
+  {{/if}}
+  {{#if @route}}
+    <LinkTo @route={{@route}}>
+      {{@text}}
+    </LinkTo>
+  {{else}}
+    {{@text}}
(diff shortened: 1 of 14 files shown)


andycochran:
Just fixed the linting that was causing CI to fail. I think this is ready to go now.

PR #402, opened 2020-06-10
Refactor applicant team: extract state mgmt into wrapper form

allthesignals:
Individual applicant fieldsets need to be wrapped in a changeset. This PR moves some behavior buried inside the applicant team editor up a level of abstraction so that it’s “closer” to the domain-specific data store methods.


godfreyyeung:
Oh, so is this PR only setting things up for a future PR that will introduce applicant changesets at the pas-form/edit level?

allthesignals:
> Oh, so is this PR only setting things up for a future PR that will introduce applicant changesets at the pas-form/edit level?

Yeah, I'm doing some refactoring in small pieces. Can I get an approve?

PR #412, opened 2020-06-12
410 Save RWCDS Form 

godfreyyeung:
This PR sets up end-to-end saving of the RWCDS Form. It

- Introduces the RWCDS Controller in the backend. This strictly follows the same pattern as the PAS Form Controller.
- Sets up the frontend Save button, and modifies the Package.save() method to save either a child PAS Form or RWCDS Form depending on dcpPackagetype.
  - Meanwhile, it also resolves a conflict between two backend endpoints over whether the optionset property, package.dcpPackagetype, should be translated before delivering a package to the frontend.
- Introduces the lightweight StickySidebar component, for re-use across forms.
- Fixes and boosts tests


godfreyyeung:
I can't figure out why Codecov is freaking out over this line:
image

client/app/models/project.js
```diff
@@ -21,7 +21,7 @@ export default class ProjectModel extends Model {
 
   get pasPackages() {
     const [firstPackage] = this.packages
-      .filter((projectPackage) => projectPackage.dcpPackagetype === 'PAS Package')
+      .filter((projectPackage) => projectPackage.dcpPackagetype === 717170000)
```
allthesignals, line 24: No lookup hash for this?

client/app/components/breadcrumbs.hbs
```diff
@@ -9,10 +9,10 @@
       {{@package.project.dcpProjectname}}
     </li>
     <li class="disabled">
-      {{#if (eq @package.dcpPackagetype '717170000')}}
+      {{#if (eq @package.dcpPackagetype (optionset 'package' 'type' 'code' 'PAS_PACKAGE'))}}
         Pre-Application Statement
       {{/if}}
-      {{#if (eq @package.dcpPackagetype '717170004')}}
+      {{#if (eq @package.dcpPackagetype (optionset 'package' 'type' 'code' 'RWCDS'))}}
```
allthesignals, line 15: thank you

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -770,108 +770,72 @@
...
+          (hash anchorId="project-area"
+            label="Project Area")
+          (hash anchorId="proposed-development-site"
+            label="Proposed Development Site")
+          (hash anchorId="project-description"
+            label="Project Description")
+          (hash anchorId="attached-documents"
+            label="Attached Documents")
+          }}
+      >
```
allthesignals, line 792: Cool!

client/app/models/project.js
godfreyyeung: Hmm, oh yeah I can maybe import the helper here...

godfreyyeung:
> This PR is doing quite a lot... it refactors PAS edit form to add StickySidebar. It fixes a lookup issue in breadcrumbs. Distinguishes between pas/rwcds error messages in testing. Adds in the actual rwcds save functionality. Adds rwcds error handling. Adds in forking behavior for the kind of package's form (pas/rwcds) save method...
> 
> Does this all really need to be in the same PR? I'm fine with it for now, but this touches _a lot_ of the code, and that can make this difficult for other people who are also trying to push code.
> 
> On top of that, we've got a new design issue: packages and "types" of forms. You have forking logic for handling these things, all throughout the code. What is a possible design solution for that? When we keep adding more types of forms, do you think this is going to continue to be maintainable? Will it be more confusing?

I agree, sorry that's my bad. I started off with a big goal and then realized I needed all these intermediate steps. 

The forking dilemma is tricky because in CRM, Packages is an entity that essentially morphs into different entities based on a property (dcpPackagetype). 

Ideally we have separate endpoints for different package types -- pasPackage, rwcdsPackage. It's just awkward because they actually both refer to a Package entity.

allthesignals:
@godfreyyeung cool, let's resolve these conflicts and start thinking about a long-term design fix for this! :)

Issue #298, opened 2020-05-18
PAS: In Applicant information, ZIP and Phone number should be numeric but it allowing alphabets


Hemachekuri:
[AB#10528](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10528)

## What should happen:
In Applicant information, ZIP and Phone number should allow only numbers
## What happened instead:
In Applicant information, ZIP and Phone number should be numeric but it allowing alphabets
## How to reproduce this bug:

Log into Applicant Portal
Click on the Edit-Pre Application Statement button
Enter any details in alphabets on the Form in the applicant section, in zip code and phone numbers

## Browser(s) 
![image](https://user-images.githubusercontent.com/63460627/82264456-f2414600-9932-11ea-87a0-7ae315383f6e.png)
![image](https://user-images.githubusercontent.com/63460627/82264547-18ff7c80-9933-11ea-9a92-5c1dfcb24d7f.png)

trbmcginnis:
Will be solved by completing Issue #313

Hemachekuri:
feild level validations for Applicnat information is as per PAS Form DD
![image](https://user-images.githubusercontent.com/63460627/82366589-68e84d00-99e0-11ea-8cd4-50bfba082109.png)
![image](https://user-images.githubusercontent.com/63460627/82366626-7f8ea400-99e0-11ea-9177-19c169f5ce1c.png)

ksinghprod:
This issue is now fixed

![image](https://user-images.githubusercontent.com/57362229/88095620-68487000-cb63-11ea-8e38-ebf2b9c131ad.png)

PR #414, opened 2020-06-16
Yield child SaveableForms; Recurse-to-save

allthesignals:
This adds behavior to SaveableForm so that it yields a version of itself that registers with the parent form, and validates state accordingly.

Many of the changes here are "consumer side" —  they address #298 by actually adding validation messages.

A miscellaneous change that also increases the size of the diff is changing the PAS form block variable name from `saveable-form` to `pasForm`.


godfreyyeung:
If I'm understanding this correctly, it seems that this new design ensures that clicking the save button of the root-level SaveableForm will trigger the validations of nested SaveableForms. However, it doesn't necessarily mean that the _models_  of nested SaveableForms themselves will save. 

There's a separate stream of logic for causing  _models_  of nested SaveableForm to save -- through the package/pasForm/rwcdsForm save() functions. 

What happens if we leave never define the saveButton/submitButton for nested saveableForms? What happens if we do?

It seems like for each nested SaveableForm, if we don't hook up the saveButton/submitButton, we get the scenario above (separate streams for saving the _models_)
If we _do_ hook up the saveButton/submitButton, then there might be _double_ saves to the models.

allthesignals:
> However, it doesn't necessarily mean that the _models_ of nested SaveableForms themselves will save.

This isn't new to the design. SaveableForm hasn't been responsible for calling model#save for a while, and I think we should keep it that way. We should keep these components unaware of the particulars of the model (specifically, the saving strategy, as that's complicated by our need for saving a bunch of other models in a single click). Thoughts?

> What happens if we leave never define the saveButton/submitButton for nested saveableForms? What happens if we do?

Hmm not sure I follow, but we don't define these buttons for nested forms as is. SaveableForm is designed to be able to _persist_ changeset changes from the root down to children. This doesn't involve calling a model save method — that's what the callback is for. If we _do_ invoke a child form's save button, it would simply trigger those children's changeset persistence methods — as well as the provided save button callback — but ignore the parent's changesets (as children don't know about their parents).

> It seems like for each nested SaveableForm, if we don't hook up the saveButton/submitButton, we get the scenario above (separate streams for saving the _models_)

I'm not sure I understand this scenario. The stream for saving a model is different from handling for persisting changesets. It's up to the consumer end to decide how saving models works. The design here assumes that click save on the parent form will persist changes for the parent form as well as any of its descendants.

As for how the model saves in our implementation, we simply have it save _everything_, including descendants.


> If we _do_ hook up the saveButton/submitButtons of each nested saveableForm, then there might be _double_ saves to the models.

That's only true if on the consumer end we pass the same callback for the parent and child forms. But as far as SaveableForm, it never even concerns itself with calling the save method on models.

Parent ->
  Child
  Child
  Child

If I call Parent.persistSaveableChanges, it's going to call _changeset_ methods that persist changes, including the consumer-side callback, and perform the same operations for child SaveableForms. If _those_ child forms have saveButtons invoked (with callbacks), those get ignored — the parent component doesn't know about child saveButton callbacks.

Am I missing something? Lemme know!

godfreyyeung:
> . We should keep these components unaware of the particulars of the model (specifically, the saving strategy, as that's complicated by our need for saving a bunch of other models in a single click). 

Okay, if this is the intentional design to decouple the model and changeset saving, I think that's fine! We should add documentation to clarify and make this very intentional. It's maybe not immediately clear that this is more of a recurse-to-validate because changeset.execute() will actually move changes into the models throughout nested SaveableForms, but it just won't send the PATCH requests. I don't know a better solution yet though, this seems like a pretty good solution for now.

If _those_ child forms have saveButtons invoked (with callbacks), those get ignored — the parent component doesn't know about child saveButton callbacks.
> 

Ah okay, I just wanted to make sure that if there are save buttons defined for nested SaveableForms, their `onClick` callbacks won't get triggered from a root-level SaveableForm save. 
Now I get why there is a `noop` default argument on `persistSaveableChangeset` passed to `_saveChangeset()`. Maybe add a comment to clarify this -- that in child saveableForms the callback is always a `noop`?

allthesignals:
> Okay, if this is the intentional design to decouple the model and changeset saving, I think that's fine! We should add documentation to clarify and make this very intentional. It's maybe not immediately clear that this is more of a recurse-to-validate because changeset.execute() will actually move changes into the models throughout nested SaveableForms, but it just won't send the PATCH requests. I don't know a better solution yet though, this seems like a pretty good solution for now.

Yeah, we really ought to think about this component as totally decoupled from "patch requests" or how the model gets saved. The only thing it does it is calls a function that gets passed to it. This is a way [to decouple things](https://sourcemaking.com/design_patterns/bridge). Perhaps the name of the component is wrong? More of a "TieredValidator" or something, and very closely tied to Changeset. 

> Now I get why there is a `noop` default argument on `persistSaveableChangeset` passed to `_saveChangeset()`. Maybe add a comment to clarify this -- that in child saveableForms the callback is always a `noop`?

Well, a child saveableForm's saveButton callback is not necessarily always a noop. For example, if it's defined:

```handlebars
<SaveableForm as |form|>
  <form.SaveableForm as |child|>
    <child.saveButton
      @onClick={{this.handleChildSave}}
    />
  </form.SaveableForm>
</SaveableForm>
```

In this case, it will call the `persistSaveableChanges` _for that child form_, _then_ trigger the passed onClick. If there isn't `onClick`, it will definitely use the `noop`.

PR #415, opened 2020-06-16
Use mut helper; extract optionsets metadata;

allthesignals:
Removes applicant-fieldset component class, uses mut helper and optionset helper for fieldsets.


client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -94,8 +94,8 @@
           @placeholder="state"
           @searchEnabled={{true}}
           @searchField="label"
-          @options={{this.stateOptions}}
-          @onchange={{fn this.updateAttr @applicant 'dcpState'}}
+          @options={{map-by 'code' (optionset 'applicant' 'state' 'list')}}
```
godfreyyeung, line 97: yay! more of the optionset API is getting used :)

PR #423, opened 2020-06-17
222 Write to project dcp_anticipatedyearbuilt

godfreyyeung:
There are two alternatives to this implementation, see below. I explored alternative 1 but got stuck modifying the JSON API Serializer. 

## PR 
Adds a new RWCDS Service in the backend in charge of updating
the RWCDS Form. While patching the RWCDS Form, it also copies
the RWCDS buildyear property over to the associated Project's
anticipatedyearbuilt property.

The service first needs to acquire the form's projectid through
a network request for the associated package. This is because
projectid information doesn't come from the frontend.

## Alternatives
1. Create a relationship between Forms (RWCDS, PAS) and Projects in the frontend, so that that when PATCHing a form, the request payload can include the Project Id. This will avoid a double network request within the RWCDS Patch. 

2. Create a separate endpoint  (controller) for PATCHing Projects. Have the frontend set the Project.dcpAnticipatedyearbuilt and call Project.save(). At first, this seems a little less sturdy since the frontend could potentially successfully PATCH a package but then abort before PATCHing the Project. However, the frontend already faces a similar risk with uploading documents, saving applicant team, bbls, etc.

### DIFF
--- a/server/src/packages/rwcds-form/rwcds-form.service.spec.ts
+++ b/server/src/packages/rwcds-form/rwcds-form.service.spec.ts
@@ -1,0 +1,44 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { CrmModule } from '../../crm/crm.module';
+import { RwcdsFormService } from './rwcds-form.service';
+import { CrmService } from '../../crm/crm.service';
+
+describe('PackagesService', () => {
+  let rwcdsFormService: RwcdsFormService;
+  let crmService: CrmService;
+
+  beforeEach(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      imports: [CrmModule],
+      providers: [RwcdsFormService],
+    }).compile();
+
+    rwcdsFormService = module.get<RwcdsFormService>(RwcdsFormService);
+    crmService = module.get<CrmService>(CrmService);
+  });
+
+  it('RWCDS Form Service Should be defined', () => {
+    expect(rwcdsFormService).toBeDefined();
+  });
+
+  it('update() should update both RWCDS form and project if rwcdsForm.buildyear is included ', async () => {
+    jest.spyOn(rwcdsFormService, '_getPackageProjectId').mockImplementation(async () => '123');
+    const updateMock = jest.spyOn(crmService, 'update').mockImplementation(async () => true);
+
+    await rwcdsFormService.update('123', { dcp_buildyear: 2020 });
+
+    expect(updateMock).toHaveBeenCalledWith('dcp_rwcdsforms', '123', { dcp_buildyear: 2020 });
+    expect(updateMock).toHaveBeenCalledWith('dcp_projects', '123', { dcp_anticipatedyearbuilt: 2020 });
+    expect(updateMock).toHaveBeenCalledTimes(2);
+  });
+
+  it('update() should only update RWCDS form if rwcdsForm.buildyear is NOT included ', async () => {
+    jest.spyOn(rwcdsFormService, '_getPackageProjectId').mockImplementation(async () => '123');
+    const updateMock = jest.spyOn(crmService, 'update').mockImplementation(async () => true);
+
+    await rwcdsFormService.update('123', { dcp_projectname: 'Chelsea' });
+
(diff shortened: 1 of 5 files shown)


server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    if(!projectId) {
+      return new HttpException('Package not found for RWCDS Form', HttpStatus.NOT_FOUND);
+    }
+
+    await this.crmService.update('dcp_rwcdsforms', id, allowedAttrs);
+
+    await this.crmService.update('dcp_projects', projectId, {
+      'dcp_anticipatedyearbuilt': allowedAttrs.dcp_buildyear,
+    });
```
allthesignals, line 38: Why does this need to happen every time we update something in the rwcds form? Can this not happen when we update the project model? Can we make an update method on the project service? Or can a separate call be made to the projects controller? What about CRM — do we know that CRM doesn't not synchronize these things?

I'm not sure it makes sense to include an _update to a property on a project_ in the rwcds. If i didn't know anything about this code, I would have no idea why every single update to the rwcds form should require an update to the related project's dcp_anticipatedyearbuilt.

Also, it's a patch request. So what if the body doesn't include `dcp_anticipatedyearbuilt`? Does it get set to undefined? Or worse, does it error out because it doesn't allow null for some reason?

I think making the rwcds service was definitely the right move, but the one-off look of `dcp_anticipatedyearbuilt` is confusing to me. Let's talk approach because this won't be the first time we'll run into this.

Otherwise, great work, _very concise_ — I know this is an awkward requirement, and I think this is a great start.

godfreyyeung: Per our standup discussion, I'll look into:
  - Whether RwcdsForm.buildyear is NULL if user does not modify the buildyear
  - Using ProjectService instead of CrmService to patch projects

godfreyyeung: - I've confirmed that RwcdsForm.buildyear _isn't_ NULL if user doesn't modify the buildyear
- The ProjectService doesn't yet have any update() methods (we haven't been writing to Project yet). @allthesignals Do you think I should just add one?

allthesignals: @godfreyyeung 

> * I've confirmed that RwcdsForm.buildyear _isn't_ NULL if user doesn't modify the buildyear

It doesn't matter. According to the JSON:API spec, we can't guarantee that all the attributes will be provided in the request body. Did you read the link I sent you?

Just because the FE application makes a request and it "works" doesn't make it a good design. The point is to _follow the spec as written_. Your code here violates the spec by overwriting the value in CRM if the body doesn't happen to include the attribute. Please fix it. If you are still confused about this please give me a call. Thanks.

godfreyyeung: Per standup this morning, the crux of the matter is that, even though Applicant Portal FE always send RwcdsForm.buildyear, another consumer of the API may not. If buildyear is not a property in the request body, the backend will interpret it as NULL when it shouldn't be. I'll write a guard for this.

server/src/packages/rwcds-form/rwcds-form.service.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  it('update() should only update RWCDS form if rwcdsForm.buildyear is NOT included ', async () => {
+    jest.spyOn(rwcdsFormService, '_getPackageProjectId').mockImplementation(async () => '123');
+    const updateMock = jest.spyOn(crmService, 'update').mockImplementation(async () => true);
+
+    await rwcdsFormService.update('123', { dcp_projectname: 'Chelsea' });
+
+    expect(updateMock).toHaveBeenCalledWith('dcp_rwcdsforms', '123', { dcp_projectname: 'Chelsea' });
+    expect(updateMock).toHaveBeenCalledTimes(1);
+  });
```
allthesignals, line 43: This is great but I thought we talked in Teams about just writing e2e tests?

Issue #222, opened 2020-05-12
RWCDS Project Description Section


nnxka:
[Azure Requirement USER STORY 10972](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/10972)
[AB#10972](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10972)
[AB#10974](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10974)

### User Story
As an applicant, I want to complete the required project information needed from me in order so that I can save and submit my RWCDS application. 

### [Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP---Applicant-Portal?node-id=0%3A1)
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/7da543d5-8760-4d84-a542-67babec4e7fc)

## Display Logic
For "Identify and describe existing conditions" question: 
  - If `dcp_projectsitedescription` (RWCDS)  IS NULL, display `dcp_projectdescriptionproposedarea` (PAS)
  - If `dcp_projectsitedescription`(RWCDS) IS NOT NULL, display `dcp_projectsitedescription`
  - Write to `dcp_projectsitedescription`(RWCDS)  

For "Describe the proposed project" question: 
  - If `dcp_proposedprojectdevelopmentdescription` (RWCDS) IS NULL, display `dcp_projectdescriptionproposeddevelopment` (PAS)
  - If `dcp_proposedprojectdevelopmentdescription' IS NOT NULL, display `dcp_proposedprojectdevelopmentdescription' (RWCDS)
  - Write to the `dcp_proposedprojectdevelopmentdescription' (RWCDS)

For `dcp_constructionphasing` "Will construction be done in multiple phases?" question, 
- NULL should display as "Not Applicable"

## Backend
- Anticipated analysis year `dcp_buildyear` input should also write directly to the project record `dcp_anticipatedyearbuilt`
- Backend needs to query from PAS form data and get `dcp_projectdescriptionproposedarea` and `dcp_projectdescriptionproposeddevelopment`

godfreyyeung:
Just tracking progress here:

- [x] Basic Project Description inputs (PR #401)
- [ ] In backend, write to project `dcp_anticipatedyearbuilt` as well
[...]

PR #424, opened 2020-06-18
222 Prefill RWCDS description fields from PAS

godfreyyeung:
Modifies the Package Service getPackage() function
to check for if two description fields are
blank on retrieved RWCDS forms. If so, it queries
for the PAS Form on the latest PAS Package revision
(on the same Project as the RWCDS), and prefills
those the RWCDS form fields from the PAS Form fields.
The pertinent fields are specified in Issue #222.


server/src/packages/packages.service.ts
```diff
@@ -70,6 +70,31 @@ const PACKAGE_TYPE_OPTIONSET = {
 export class PackagesService {
   constructor(private readonly crmService: CrmService) {}
 
+  // Gets the lateset PAS Package on Project `projectId`
```
godfreyyeung, line 73: comment should actually say latest "PAS Form"

server/src/packages/packages.service.ts
```diff
@@ -102,6 +127,7 @@ export class PackagesService {
       dcp_project,
       dcp_packageid,
       dcp_name,
+      _dcp_project_value,
```
allthesignals, line 130: Isn't this already available in `dcp_project`? If so, is there a good request to add more code to get the project id?

server/src/packages/packages.service.ts
```diff
@@ -70,6 +70,31 @@ const PACKAGE_TYPE_OPTIONSET = {
...
+
+      return latestPasForm;
+    } catch(e) {
+      const errorMessage = `Error finding a PAS Package on given Project. ${e.message}`;
+      console.log(errorMessage);
+      throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+    }
+  }
+
   async getPackage(packageId) {
```
allthesignals, line 98: this is starting to do way more than get a package. It gets a package, gets related
  forms, detects the form type, includes that form with the payload, includes additional
  munging that's specific to the form, and also includes attached documents.

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      &$expand=
+        dcp_dcp_applicantinformation_dcp_pasform,
+        dcp_dcp_applicantrepinformation_dcp_pasform,
+        dcp_package($select=${PACKAGE_ATTRS.join(',')}),
+        dcp_dcp_projectbbl_dcp_pasform($filter=statecode eq 0),
+        dcp_dcp_projectaddress_dcp_pasform
+    `);
+
+    // CRM feature for providing the human-readable value for the related resource
+    const { '_dcp_projectname_value@OData.Community.Display.V1.FormattedValue': dcp_projectname } = pasForm;
```
allthesignals, line 25: this... is probably more confusing that just:

```javascript
 const dcp_projectname = pasForm['_dcp_projectname_value@OData.Community.Display.V1.FormattedValue'];
```

server/src/packages/pas-form/pas-form.controller.ts
```diff
@@ -111,6 +111,9 @@ export const PAS_FORM_ATTRS = [
   'dcp_projectdescriptionproposedarea',
   'dcp_projectdescriptionsurroundingarea',
   'dcp_projectattachmentsotherinformation',
+
+  // Internals
+  '_dcp_projectname_value',
```
godfreyyeung, line 116: What's the difference between this and `dcp_projectname`?

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return {
+      ...pasForm,
+
+      dcp_revisedprojectname: pasForm.dcp_revisedprojectname || dcp_projectname,
+    };
+  }
+
+  // Gets the lateset PAS Package on Project `projectId`
```
godfreyyeung, line 34: We should change this comment to 
`// Gets the lateset PAS Form on Project `projectId``

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      `);
+
+      // if pasA is of a higher version than pasB, it should come first
+      const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
+        return pasB.dcp_packageversion - pasA.dcp_packageversion;
+      });
+
+      return latestPasForm;
+    } catch (e) {
+      const errorMessage = `Error finding a PAS Package on given Project. ${e.message}`;
```
godfreyyeung, line 52: Also need to change the error message here to ``Error finding a PAS Form...`

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      ...pasForm,
+
+      dcp_revisedprojectname: pasForm.dcp_revisedprojectname || dcp_projectname,
+    };
+  }
+
+  // Gets the lateset PAS Package on Project `projectId`
+  async getLatestPasForm(projectId): Promise<any> {
+    try {
+      const { records: pasPackages } = await this.crmService.get('dcp_packages', `
```
godfreyyeung, line 37: Just curious/for discussion: should this function go int he Package service, since it acts on dcp_packages?
Or should it stay here in the pas-form service, since it returns a PAS Form?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PACKAGE_ATTRS } from '../packages.controller';
+
+@Injectable()
+export class RwcdsFormService {
+  constructor(
+    private readonly crmService: CrmService,
+    private readonly pasFormService: PasFormService,
+  ) {}
+
+  async find(id) {
```
godfreyyeung, line 13: I think this function does too much to simply be called a generic `find` function.  Should be something like `getPrefilled`

server/src/packages/pas-form/pas-form.controller.ts
allthesignals: Because we need that data in the request.

`_dcp_projectname_value` is the project identifier.

server/src/packages/pas-form/pas-form.service.ts
allthesignals: go for it

server/src/packages/pas-form/pas-form.service.ts
allthesignals: go for it

server/src/packages/pas-form/pas-form.service.ts
allthesignals: That's a good question — it does call `dcp_packages`, but it's ultimately supposed to return a PAS-FORM. It takes a project ID as an argument, which is a little confusing as well. Maybe the argument should be changed?

I find that querying `dcp_packages` is just a detail of the implementation. We want to create a concise abstraction, concealing the nitty-gritty underneath.

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -0,0 +1,57 @@
...
+        $filter=
+          _dcp_project_value eq ${projectId}
+          and dcp_packagetype eq ${PACKAGE_TYPE_OPTIONSET['PAS_PACKAGE'].code}
+        &$expand=
+          dcp_pasform
+      `);
+
+      // if pasA is of a higher version than pasB, it should come first
+      const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
+        return pasB.dcp_packageversion - pasA.dcp_packageversion;
```
allthesignals, line 47: FYI you can sort using the OData API

server/src/packages/rwcds-form/rwcds-form.service.ts
allthesignals: Hmmm `getPrefilled` seems wrong to me — don't we always want it to do that, no matter what? AFAICT, it's a full part of the RWCDSForm "model" — there isn't really an exception where we _don't_ want it prefilled with some data.

If we had proper Model objects, it could be handled there. I know this is possible in Rails, but here we are.

server/src/packages/pas-form/pas-form.service.ts
godfreyyeung: :o whoa

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: That's an interesting point, that we can lean on the assumption that the consumers of the API will never want an RWCDS without pre-filled fields. 

I guess that's like the backend to pre-filling the PAS name from the Project name

server/src/projects/projects.controller.ts
```diff
@@ -23,13 +23,14 @@ export const PROJECT_ATTRS = [
...
   '_dcp_applicant_customer_value',
   'dcp_dcp_project_dcp_projectapplicant_Project',
-  'packages',
 ];
 
 @UseInterceptors(new JsonApiSerializeInterceptor('projects', {
   id: 'dcp_name',
   attributes: [
     ...PROJECT_ATTRS,
+
```
godfreyyeung, line 32: @allthesignals  Can you explain why you broke this out?

allthesignals: @godfreyyeung because `packages` isn't a real CRM field. All exported _ATTRS really must be real life CRM fields.

This is enforced additionally through the way we add _ATTRS into the `$select=` parts of the queries (for speed improvements).

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { records: [rwcdsForm] } = await this.crmService.get(`dcp_rwcdsforms`, `
+      $filter=
+        dcp_rwcdsformid eq ${id}
+      &$expand=dcp_package($select=${PACKAGE_ATTRS.join(',')})
```
godfreyyeung, line 17: @allthesignals Hmm, do we really need to expand packages on the rwcds form here? 

Wouldn't this be redundant, since the packages.service already requests the package?

I think expanding it here makes the final returned payload look like this:
```
  {
      ...package attrs
      ...RWCDS attrs,
      package: {
            ... package attrs
         },
        documents: {
           ....
        } 
     }
 ```

server/src/projects/projects.controller.ts
godfreyyeung: Oh, you mean it's a relationship instead of a static property?

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: Yeah, the package info is repeated in the nesting:
image
Is this necessary?

allthesignals: FWIW the package isn't included for serialization unless explicitly added. 

That said, probably not, no — I think this is a second request being added on after a package request. Makes sense. 

One other thing: we should add a `$select` here so we ask for explicit rwcds fields. That way, the request is fast.

godfreyyeung: Per standup, this is an extra expand that is ignored by the JSON API Serializer anyway, so we can likely get rid of it. I'll look into it.

allthesignals:
closing in favor of https://github.com/NYCPlanning/labs-applicant-portal/pull/437

PR #427, opened 2020-06-19
Removes unused components; cleans up some naming

allthesignals:
🧹 🧹 🧹 

The naming makes the Form.Field component a little more generic.


client/app/components/saveable-form/field.js
```diff
@@ -2,15 +2,15 @@ import Component from '@glimmer/component';
...
   }
 
   set value(newValue) {
-    this.args.saveableChanges[this.args.attribute] = newValue;
+    this.args.data[this.args.attribute] = newValue;
   }
 
   get error() {
-    return this.args.submittableChanges[`error.${this.args.attribute}`];
+    return this.args.error[this.args.attribute];
```
godfreyyeung, line 13: At this point maybe we can just treat error as a string? And not assume we need an error object? What other properties could the error object hold?

allthesignals: @godfreyyeung 

Ah you mean pass the error prop? I'll try that but I'm not sure it'll work because the `attribute` key isn't always present in the error. But let me give it a shot because that would be a simpler design.

allthesignals: @godfreyyeung oh, the attribute isn't know up front. So we need the full error object.

godfreyyeung: 👍

allthesignals: @godfreyyeung  I don't follow. Do you mean `get @error @attribute`? If you're referring to parentField, it needs to have the same interface. We curry this component from saveableForm.

allthesignals: approve?

client/app/components/saveable-form/field/radio-group.hbs
```diff
@@ -1,21 +1,23 @@
...
-        {{radio.label}}
-      </RadioButton>
-    </span>
-  {{/each}}
-</@parentField>
+{{#let @parentField as |Field|}}
+  <Field
+    @type="radio"
+    @hideError={{true}}
+    as |RadioButton|
```
godfreyyeung, line 5: This is a bit confusing to me. Can we just directly use the Radio component here?

client/app/components/saveable-form/field.js
godfreyyeung: sorry was trying to wrap my head around the radio yielding

client/app/components/saveable-form.hbs
```diff
@@ -12,12 +12,8 @@
       submittableChanges=submittableValidator.changeset
 
       Field=(component 'saveable-form/field'
-        saveableChanges=saveableValidator.changeset
-        submittableChanges=submittableValidator.changeset
-      )
-
-      Radio=(component 'radio-button'
```
godfreyyeung, line 19: Oh okay, I see you deleted it. (Since the Radio here isn't needed at a top level anymore since it's accessed through Field)

client/app/components/saveable-form/field/radio-group.hbs
godfreyyeung: What's the pros/cons of accessing Radio through Field vs directly, like
`<Package::SaveableForm::Field:Radio />`?
One observation: we wouldn't have to specify `@hideError={{true}}` in that case.

allthesignals: All this change does is use `let` to alias the component so that it's cleaner syntax? You're commenting about code that's already merged. I'm confused why this blocks an approval. But, here we go:

> This is a bit confusing to me. Can we just directly use the Radio component here?
It already does that. Could you clarify?

> What's the pros/cons of accessing Radio through Field vs directly, like
> `<Package::SaveableForm::Field:Radio />`?
I mean, do you really think invoking components directly like that is a better design? Accessing it directly would require passing all arguments every single time. That'll be like 5 arguments. We would have the same number of lines of code before we refactored. Using contextual components prevents that and is more semantic. Are you proposing a different API? What alternatives would you like to see?

> One observation: we wouldn't have to specify `@hideError={{true}}` in that case.
In what case?

allthesignals: Hmm just FYI, all this change does is use `let` to alias the component so that it's cleaner syntax.

> This is a bit confusing to me. Can we just directly use the Radio component here?

It already does that. Could you clarify?

> What's the pros/cons of accessing Radio through Field vs directly, like
> `<Package::SaveableForm::Field:Radio />`?

I mean, are you saying that invoking directly is a better design? Accessing it directly would require passing all arguments every single time. That'll be like 5 arguments. We would have the same number of lines of code before we refactored. Using contextual components prevents that and is more semantic. Are you proposing a different API? What alternatives would you like to see?

> One observation: we wouldn't have to specify `@hideError={{true}}` in that case.

In what case?

PR #428, opened 2020-06-19
ExternalLink 

andycochran:
This PR implements the `<Ui::ExternalLink />` component throughout the content. 

It also adds splatributes to the component (e.g. so it can be passed a `.nowrap` class to prevent the icon wrapping to the next line in certain cases).

### DIFF
--- a/client/app/components/packages/pas-form/project-area.hbs
+++ b/client/app/components/packages/pas-form/project-area.hbs
@@ -57,7 +57,7 @@
 
     <Ui::Fieldset as |Q|>
       <Q.legend>
-        What is the <a href="https://streets.planning.nyc.gov/about" target="_blank" rel="noopener noreferrer">Legal Street Frontage Status in the Project Area<sup><FaIcon @icon="external-link-alt" @prefix="fas" @fixedWidth={{true}} /></sup></a>?
+        What is the <Ui::ExternalLink @href="https://streets.planning.nyc.gov/about">Legal Street Frontage Status in the Project Area</Ui::ExternalLink>?
       </Q.legend>
 
       <Form.Field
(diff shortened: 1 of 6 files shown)


client/app/components/packages/pas-form/project-area.hbs
```diff
@@ -80,12 +80,7 @@
     <Ui::Fieldset as |Q|>
       <Q.legend>
         Do all land use actions meet
-        <a
-          href="https://govt.westlaw.com/nycrr/Document/I4ec3a767cd1711dda432a117e6e0f345?viewType=FullText&originationContext=documenttoc&transitionType=CategoryPageItem&contextData=(sc.Default)"
-          target="_blank"
-          rel="noopener noreferrer">
-          SEQRA or CEQR criteria<sup><FaIcon @icon="external-link-alt" @prefix="fas" @fixedWidth={{true}} /></sup>
-        </a>
+        <Ui::ExternalLink @href="https://govt.westlaw.com/nycrr/Document/I4ec3a767cd1711dda432a117e6e0f345?viewType=FullText&originationContext=documenttoc&transitionType=CategoryPageItem&contextData=(sc.Default)">SEQRA or CEQR criteria</Ui::ExternalLink>
```
allthesignals, line 83: I think splitting this out would be easier to read:

```handlebars
<Ui::ExternalLink
  @href="https://govt.westlaw.com/nycrr/Document/I4ec3a767cd1711dda432a117e6e0f345?viewType=FullText&originationContext=documenttoc&transitionType=CategoryPageItem&contextData=(sc.Default)"
>
  SEQRA or CEQR criteria
</Ui::ExternalLink>
```

Up to you, I suppose =/ is there a reason you like to keep things inline that I don't understand? Spacing issues?

andycochran: I did them all the same for consistency and meaning. 

I do understand that splitting tags to multiple lines can make _attributes_ more readable. But I find it confusing when inline elements are code-styled like block elements. Whitespace inside an inline tag has meaning; it's not just a style preference. 

Components like this — ones that spit out inline elements — are most commonly used in the middle of paragraph text, which very often makes whitespace a problem. For example, when there's punctuation (a period or question mark) after this component, splitting to multiple lines would add a space between the link and the punctuation. 

I suppose it's a matter of how important consistency is. If consistency isn't important, we could default to coding inline elements as if they're block elements, knowing that we'll sometimes need to code them inline when whitespace would have meaning or undesired effect.

PR #432, opened 2020-06-22
Ui::Question component

andycochran:
This PR refactors the `<Ui::Fieldset>` component so that it works for all form questions (including text inputs and textareas, which need to be wrapped in a `<label>` instead of a `<fieldset>`). 

The new API is: 

```
<Ui::Question [@group={{true}}] as |Q|>
  <Q.prompt>
  </Q.prompt>
  ...
</Ui::Question>
```

If you pass the optional `@group={{true}}`, the component will render a `<fieldset>` element instead of the default `<label>` element, and call the `Ui::label-text` component instead of the default `Ui::FieldsetLegend` component.

This PR also 
- Changes the validation message classes to follow the style guide
- Nixes some unnecessary CSS

### DIFF
--- a/client/app/components/land-use-action.hbs
+++ b/client/app/components/land-use-action.hbs
@@ -25,10 +25,14 @@
   {{/if}}
 
   {{#each this.sortedSelectedActions as |landUseAction|}}
-    <Ui::Question class="fieldset relative scale-fade-in" as |Q|>
-      <Q.Prompt data-test-action-name="{{landUseAction.name}}">
+    <Ui::Question
+      @group={{true}}
+      class="fieldset relative scale-fade-in"
+      as |Q|
+    >
+      <Q.prompt data-test-action-name="{{landUseAction.name}}">
         {{landUseAction.name}}
-      </Q.Prompt>
+      </Q.prompt>
 
       {{#if (or (eq landUseAction.name "Zoning Special Permit") (eq landUseAction.name "Zoning Certification") (eq landUseAction.name "Zoning Authorization") (eq landUseAction.name "Zoning Map Amendment") (eq landUseAction.name "Zoning Text Amendment") (eq landUseAction.name "Modification") (eq landUseAction.name "Renewal"))}}
         {{#if (or (eq landUseAction.name "Zoning Special Permit") (eq landUseAction.name "Zoning Certification") (eq landUseAction.name "Zoning Authorization"))}}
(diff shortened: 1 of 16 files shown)


client/app/components/land-use-action.hbs
```diff
@@ -25,10 +25,10 @@
   {{/if}}
 
   {{#each this.sortedSelectedActions as |landUseAction|}}
-    <Ui::Fieldset class="fieldset relative scale-fade-in" as |Q|>
-      <Q.legend data-test-action-name="{{landUseAction.name}}">
+    <Ui::Question class="fieldset relative scale-fade-in" as |Q|>
+      <Q.prompt data-test-action-name="{{landUseAction.name}}">
```
allthesignals, line 29: I think the convention goes that components should always be capitalized. So this should be 👍 
```
<Q.Prompt />
```

client/app/components/ui/question.hbs
```diff
@@ -0,0 +1,18 @@
...
+      )}}
+    </fieldset>
+  {{else}}
+    <label class="medium-margin-bottom" ...attributes>
+      {{yield (hash
+        prompt=(component promptComponent)
+      )}}
+    </label>
+  {{/if}}
+{{/let}}
```
allthesignals, line 18: I'd simplify this a bit more:

```handlebars
  {{#if @group}}
    <fieldset class="medium-margin-bottom" ...attributes>
      {{yield (hash
        Prompt=(component 'ui/fieldset-legend')
      )}}
    </fieldset>
  {{else}}
    <label class="medium-margin-bottom" ...attributes>
      {{yield (hash
        Prompt=(component 'ui/label-text')
      )}}
    </label>
  {{/if}}
```

client/app/components/land-use-action.hbs
andycochran: Changed.

client/app/components/ui/question.hbs
andycochran: Duh! I totally overcomplicated this because I arrived at it from a different angle where I needed to dynamically bind some args. Thanks for catching that.

andycochran:
Thanks for the feedback, @allthesignals. I've made your suggested improvements and I went ahead and implemented this method thru the whole PAS form.

andycochran:
Yeah, RWCDS will need this too. I just didn't wanna touch that while the other pieces are in pretty nascent shape.

PR #433, opened 2020-06-23
Create ConfirmationModal for SaveableForms

allthesignals:
Integrate into both PAS and RWCDS forms.

Separately, updates Rwcds form to use new APIs. Fixes some markup issues.

Adds named blocks polyfill for better composability.


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -46,56 +46,47 @@
...
-              </div>
-            </div>
-            {{!-- Error Handling --}}
-            <Packages::PasForm::PasFormError @package={{@package}}/>
-          </RevealModal>
-        </EmberWormhole>
-      {{/if}}
+
+      <pasForm.ConfirmationModal>
+        <:message>
```
godfreyyeung, line 68: I wonder if there's someway to scope this named block to ConfirmationModal

allthesignals: @godfreyyeung submit an RFC.

client/app/components/ui/modal.hbs
```diff
@@ -0,0 +1,49 @@
+{{#if @open}}
+  <EmberWormhole @to={{if @container @container "reveal-modal-container"}}>
```
andycochran, line 2: I don't see `@container` defined anywhere. Is this just so the `@to` can be optionally passed? Would we ever need to wormhole the modal to a different place in the markup? I think the layout will only work if the modal is rendered in that one place, z-indexed after everything in the main application template.

andycochran: Regardless, I love that you've moved the wormhole from the content template to this component. ❤️

client/app/components/ui/modal.hbs
```diff
@@ -0,0 +1,49 @@
...
+              class="button expanded large secondary"
+              {{on "click" @toggle}}
+            >
+              Continue Editing
+            </button>
+          </div>
+          <div class="cell large-6 small-padding-left">
+            {{yield @toggle to="action"}}
+          </div>
+        </div>
```
andycochran, line 33: This new component is named `Ui::Modal` but its hardcoded content makes it specifically an "[action]/continue editing" modal. If we need a modal for other purposes, we'd have to refactor this. 

We could just assume this component is specifically for the form edit confirmation modal, and when we need another modal we'll rename this one, e.g. `Ui::EditConfirmationModal` and make another `Ui::WhateverModal`. 

Or, should this modal be more generic? The confirm/cancel buttons could just be passed with the `:message`. But then, named blocks wouldn't be getting us much in a generic modal component, since we could just pass the `:message`, `:action`/buttons, and `:footer` in one block.

I'm fine either way. This works for now. I just wanted to point out that this isn't a generic modal component.

client/app/components/ui/modal.hbs
allthesignals: @andycochran hmm, this code makes 'reveal-modal-container' the default target. This is a design choice. Basically, I want the modal to make no assumptions about outside context. Having the default sort of goes against that, but this kind of approach makes it easier, especially in testing, to create whatever outside context you need. It's really a matter of object-oriented design, specifically encapsulation.

client/app/components/ui/modal.hbs
allthesignals: @andycochran yep, I like your idea of splitting this out further — modal should be more generic, and we could have another component that uses that modal. I'll do that. Thanks.

client/app/components/ui/modal.hbs
andycochran: Cool. If it helps testing, great. But I can't think of a situation where a modal would need a different target; it must be added to the DOM in one specific place in the markup.

allthesignals: @andycochran I get that, but the point is to design this as if it doesn't know anything about the outside markup.

In fact, we should have this paradigm in mind, always: design components to be unaware of outside context. The more we do that, the easier it is to maintain our code.

I get that you can't think of a situation now, but maybe there is one later? I don't know. I think if I follow the comment you make below, this'll be clearer, but I hope the principle makes sense.

client/tests/integration/components/ui/modal-test.js
```diff
@@ -0,0 +1,30 @@
...
+        @open={{this.open}}
+        @toggle={{this.toggle}}
+      >
+        template block text
+      </Ui::Modal>
+    `);
+
+    await this.toggle();
+
+    assert.dom().containsText('template block text');
```
trbmcginnis, line 28: Nit picky thing: maybe you should also include a negative test before you run `await this.toggle()`

e.g. `assert.dom()doesNotContainText('template block text')` so that you know the modal is in fact opening when you toggle it (and not open by default or opening for some other reason)

client/app/components/ui/modal.hbs
trbmcginnis: Going off of what @andycochran said about this being more generic. I think that our modals will most likely always have a button similar to (although might not be exactly the same) the `continue editing` button. The user is always able to toggle the modal through 1) a button similar to `Cancel` or `Continue Editing` or 2) the Close Modal button. So I think while it might not always have to say `continue editing`, I think we should keep this button here and the only thing that should be changeable would be the text on the button (e.g. `continue editing` or `return to xyz` etc.)

(also I really like that we have a passable action here for this button too)

client/tests/integration/components/ui/modal-test.js
allthesignals: Aah! Yep. Actually... I bet you the negative test fails. Will fix.

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -46,56 +46,47 @@
...
+          <h6>Confirm Pre-Applicant Statement submission</h6>
+          <p class="header-large medium-margin-top small-margin-bottom">Are you sure?</p>
+          <p>Before submitting, ensure that your answers are accurate and complete, and that necessary attachments (including the
+            signature form) have been uploaded. If City Planning does not receive enough accurate information to provide guidance,
+            the lead planner will notify you and request that this form be resubmitted with needed materials, corrections, or
+            clarifications.</p>
+        </:message>
+
+        <:action>
+          <pasForm.SubmitButton
```
trbmcginnis, line 78: Just another nit picky thing I'm noticing -- also maybe not relevant to this specific PR but figured I'd note it anyway. Noticing that we use `pasForm` in `pas-form/edit.hbs` here but then use `saveableForm` in `rwcds-form/edit.hbs`. Just for consistency, should probably update rwcds-form.

allthesignals: @trbmcginnis yeah, I'm not sure what is the best convention. `pasForm` seems right because it makes it clear that it's validating that kind of thing and that the scope has options for that stuff.

godfreyyeung:
I'm wary of introducing edge features and using edge addons. I think we should wait until it's part of Ember to use named blocks.  Sorry, I should have brought this up on Monday. 

Here, can you do the same thing by passing a hash of components, or providing an attribute for each block?
Something like

modal.hbs
```hbs
....
{{yield}}
...
<@actionBlock />
...
<@footerBlock />
```

edit.hbs
```
 <pasForm.ConfirmationModal
         @action={{component pasForm.SubmitButton
            @onClick=(this.submitPackage)
            @isEnabled=(pasForm.isSubmittable)
            class="no-margin"
            data-test-confirm-submit-button
        />
        @footer={{component Packages::PasForm::PasFormError @package=(@package){{ />
  > 
          <h6>Confirm Pre-Applicant Statement submission</h6>
          <p class="header-large medium-margin-top small-margin-bottom">Are you sure?</p>
          <p>Before submitting, ensure that your answers are accurate and complete, and that necessary attachments (including the
            signature form) have been uploaded. If City Planning does not receive enough accurate information to provide guidance,
            the lead planner will notify you and request that this form be resubmitted with needed materials, corrections, or
            clarifications.</p>
</pasForm.ConfirmationModal>
```

allthesignals:
Can't test for submission until the rest of the questions are in

PR #437, opened 2020-06-23
Preview 222 prefill fields from pas

godfreyyeung:
Using this instead of https://github.com/NYCPlanning/labs-applicant-portal/pull/424


allthesignals:
Can we close? I Don't get why we need a draft PR to do this.

allthesignals:
Going to take a look and add our fix to this

Issue #396, opened 2020-06-09
Attachments section for RWCDS (possible refactor/redo of Attachments) 


godfreyyeung:
Wire up the Attachments section for the RWCDS form. 

We discussed possibly removing the ember-file-upload dependency.

Do we want to take the time here to explore that?

PR #442, opened 2020-06-24
234 Add Attached Documents section for RWCDS Form

godfreyyeung:
Creates an attached-documents component for RWCDS form, similar to the one for PAS form. Creates an "download link" component, which is similar to the "external link" component, to support the attached-documents text.

### DIFF
--- a/client/app/components/packages/rwcds-form/attached-documents.hbs
+++ b/client/app/components/packages/rwcds-form/attached-documents.hbs
@@ -1,0 +1,57 @@
+<@form.Section @title="Attached Documents">
+  <div class="grid-x grid-margin-x">
+    <div class="cell large-12">
+      <p>
+        Please attach the following required documents.
+      </p>
+      <ul class="text-small">
+        <li class="small-margin-bottom">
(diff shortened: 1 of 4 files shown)


godfreyyeung:
> Thanks — this is good, thanks for the Download Link component.
> 
> We could implement this now, but it won't address the bug that [this pr](https://github.com/NYCPlanning/labs-applicant-portal/pull/318) surfaced/addresses. Have you noticed that uploading files then clicking "back" breaks that app?
> 
> If you think it's best to add this and fix it later, that's fine — I do think we should be a little more proactive, though.

Thanks, yeah I think we can do this in steps. Yep, I noticed it, too.  #318 and #396  we can address in separate PRs without affecting this template code

godfreyyeung:
@andycochran Let me know if you think the design looks good 
image

godfreyyeung:
Save Andy's comments for when I return to this:
       Nix the "Please" to change that sentence to "Attach the following required documents:"
	(with a ":" instead of a ".")
	Make the bold paragraphs normal weight (they're less important than the required docs list)
	All the text can be the default size except the parens after "Massing diagrams"

PR #443, opened 2020-06-24
RWCDS show route and template

andycochran:
This PR…
- Add the RWCDS show route and template
- Adds a `<Ui:Answer>` component that reduces the repetition and brittle markup of the PAS and RWCDS show routes — which works much like the `<Ui:Question>` component for the edit routes
  ```
  <Ui::Answer [@beside={{true}}] as |A|>
    <A.Prompt>
      ...
    </A.Prompt>
    <A.Field>
      ...
    </A.Field>
  </Ui::Answer>
  ```
  (Passing the optional `@beside={{true}}` will put the prompt and field in a grid so that they're beside each other. Without it, the prompt appears above the filed.) 
- Updates the `{{optionset}}` helper so that it can be passed falsey args

NOTE: Still need to get the list of actions and attachments in the RWCDS show template

### DIFF
--- a/client/app/components/packages/rwcds-form/show.hbs
+++ b/client/app/components/packages/rwcds-form/show.hbs
@@ -223,7 +223,6 @@
         Attachments
       </h2>
 
-      {{log @package.rwcdsForm}}
     </section>
 
   </div>{{! end left/main column }}
--- a/client/app/components/packages/rwcds-form/show.hbs
+++ b/client/app/components/packages/rwcds-form/show.hbs
@@ -1,0 +1,250 @@
+<div class="grid-x grid-margin-x">
+  <div class="cell large-8">
+
+    <section class="form-section">
+      <h1 class="no-margin header-large">
+        <span id="project-info" class="section-anchor"></span>
+        Reasonable Worst Case Development Scenario
+      </h1>
+
+      <h2 class="section-header header-xxlarge"
+        data-test-show="dcpProjectname">
+        {{@package.rwcdsForm.dcpProjectname}}
+      </h2>
+    </section>
+
+    <section class="form-section">
+      <h2 class="section-header">
+        <span id="" class="section-anchor"></span>
+        Description
+      </h2>
+
+      <p class="grid-x">
+        <strong class="cell auto large-padding-right">
+          Has the project changed since submission of the Pre-Application Statement (PAS)?
+        </strong>
+        <span class="cell large-3">
+          {{if @package.pasForm.dcpHasprojectchangedsincesubmissionofthepas "Yes" "No"}}
+        </span>
+      </p>
+
+      <p class="ruled-top">
+        <strong>
+          Project site description:
+        </strong>
+        <br />
+        {{@package.rwcdsForm.dcpProjectsitedescription}}
+      </p>
+
+      <p class="ruled-top">
+        <strong>
+          Proposed development description:
+        </strong>
+        <br />
+        {{@package.rwcdsForm.dcpProposedprojectdevelopmentdescription}}
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          Will construction be done in multiple phases?
+        </strong>
+        <span class="cell large-3">
+          {{#if (eq @package.rwcdsForm.dcpConstructionphasing 717170000)}}
+            Single
+          {{else if (eq @package.rwcdsForm.dcpConstructionphasing 717170001)}}
+            Multiple
+          {{else}}
+            Not Applicable
+          {{/if}}
+        </span>
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          What is the proposed analysis year of the proposed action?
+        </strong>
+        <span class="cell large-3">
+          {{@package.rwcdsForm.dcpBuildyear}}
+        </span>
+      </p>
+
+      <p class="ruled-top">
+        <strong>
+          Analysis year rationale:
+        </strong>
+        <br />
+        {{@package.rwcdsForm.dcpRationalbehindthebuildyear}}
+      </p>
+
+      <p class="ruled-top">
+        <strong>
+          History of sites within the Project Area:
+        </strong>
+        <br />
+        {{@package.rwcdsForm.dcpSitehistory}}
+      </p>
+    </section>
+
+    <section class="form-section">
+      <h2 class="section-header">
+        <span id="" class="section-anchor"></span>
+        Proposed Actions
+      </h2>
+
+      {{!-- TODO: add list of actions --}}
+
+      <p class="ruled-top">
+        <strong>
+          Purpose and need for all proposed actions:
+        </strong>
+        <br />
+        {{@package.rwcdsForm.dcpPurposeandneedfortheproposedaction}}
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          Does the applicant plan to develop a 100% affordable housing development?
+        </strong>
+        <span class="cell large-3">
+          {{if @package.pasForm.dcpIsplannigondevelopingaffordablehousing "Yes" "No"}}
+        </span>
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          Is the applicant seeking actions from other City/State/Federal agencies?
+        </strong>
+        <span class="cell large-3">
+          {{if (eq @package.rwcdsForm.dcpIsapplicantseekingaction 717170000) "Yes"}}
+          {{if (eq @package.rwcdsForm.dcpIsapplicantseekingaction 717170001) "No" }}
+          {{if (eq @package.rwcdsForm.dcpIsapplicantseekingaction 717170002) "Unsure at this time"}}
+        </span>
+      </p>
+
+      {{#if (eq @package.rwcdsForm.dcpIsapplicantseekingaction 717170000)}}
+        <p class="ruled-top">
+          <strong>
+            Actions sought from other agencies:
+          </strong>
+          <br />
+          {{@package.rwcdsForm.dcpWhichactionsfromotheragenciesaresought}}
+        </p>
+      {{/if}}
+    </section>
+
+    <section class="form-section">
+      <h2 class="section-header">
+        <span id="" class="section-anchor"></span>
+        Analysis
+      </h2>
+
+      <p class="">
+        <strong>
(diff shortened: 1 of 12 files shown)


client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,250 @@
...
+        <span id="" class="section-anchor"></span>
+        Description
+      </h2>
+
+      <p class="grid-x">
+        <strong class="cell auto large-padding-right">
+          Has the project changed since submission of the Pre-Application Statement (PAS)?
+        </strong>
+        <span class="cell large-3">
+          {{if @package.pasForm.dcpHasprojectchangedsincesubmissionofthepas "Yes" "No"}}
```
godfreyyeung, line 27: 2 things:
1) should say `@package.rwcdsForm.dcpHas...`
2) We should use the `{{optionset ...}}` helper here, so that we don't need to hardcode the labels.
```
{{#if @package.rwcdsForm.dcpHasprojectchangedsincesubmissionofthepas}}
   {{optionset 'rwcdsForm' 'dcpHasprojectchangedsincesubmissionofthepas' 'label' @package.rwcdsForm.dcpHasprojectchangedsincesubmissionofthepas}}
{{/if}}
```

I know it seems verbose, but programmatically it is actually cleaner

client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,249 @@
...
+        <br />
+        {{@package.rwcdsForm.dcpProposedprojectdevelopmentdescription}}
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          Will construction be done in multiple phases?
+        </strong>
+        <span class="cell large-3">
+          {{#if (eq @package.rwcdsForm.dcpConstructionphasing 717170000)}}
```
godfreyyeung, line 52: Same as (2) in the comment above. we should use the optionset helper here.
```
{{optionset 'rwcdsForm' 'dcpConstructionphasing' 'label' @package.rwcdsForm.dcpConstructionphasing}}
```
This time, the optionset helper will significantly reduce line count :) ⛳

client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,249 @@
...
+          {{if @package.pasForm.dcpIsplannigondevelopingaffordablehousing "Yes" "No"}}
+        </span>
+      </p>
+
+      <p class="grid-x ruled-top">
+        <strong class="cell auto large-padding-right">
+          Is the applicant seeking actions from other City/State/Federal agencies?
+        </strong>
+        <span class="cell large-3">
+          {{if (eq @package.rwcdsForm.dcpIsapplicantseekingaction 717170000) "Yes"}}
```
godfreyyeung, line 118: The optionset helper can be used here, too. 
This time, you'll have to 
1) add the optionset export to `app/models/rwcds-form.js`, because it isn't already in there. e.g.

```
export const DCPISAPPLICANTSEEKINGACTION_OPTIONSET = {
  YES: {
    code: true,
    label: 'Yes',
  },
  NO: {
    code: false,
    label: 'No',
  },
};
```

2) register it in `helpers/optionset.js`, following the other examples. e.g.
```
  rwcdsForm: {
    dcpHasprojectchangedsincesubmissionofthepas: DCPHASPROJECTCHANGEDSINCESUBMISSIONOFTHEPAS_OPTIONSET,
    dcpConstructionphasing: DCPCONSTRUCTIONPHASING_OPTIONSET,
    dcpIsapplicantseekingaction: DCPISAPPLICANTSEEKINGACTION_OPTIONSET,
  },
```

client/app/components/packages/rwcds-form/show.hbs
andycochran: Ack! Thanks for catching that copy/paste "pas".

client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,249 @@
...
+      <h3>
+        No-Action Scenario
+      </h3>
+
+      <p class="grid-x">
+        <strong class="cell auto large-padding-right">
+          Is the No-Action scenario the same as the Existing Conditions?
+        </strong>
+        <span class="cell large-3">
+          {{if @package.rwcdsForm.dcpExistingconditions "Yes" "No" }}
```
godfreyyeung, line 158: Optionset helper also needed here.

client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,249 @@
...
+      <h3>
+        With-Action Scenario
+      </h3>
+
+      <p class="grid-x">
+        <strong class="cell auto large-padding-right">
+          Is the proposed project the same as the proposed With-Action scenario?
+        </strong>
+        <span class="cell large-3">
+          {{if @package.rwcdsForm.dcpIsrwcdsscenario "Yes" "No" }}
```
godfreyyeung, line 191: Optionset helper also needed here

client/app/components/packages/rwcds-form/show.hbs
andycochran: Why do we need the if here?

andycochran: The code you provide will work for "YES" but not for "NO".

client/app/components/packages/rwcds-form/show.hbs
```diff
@@ -0,0 +1,249 @@
...
+        (hash anchorId="project-description"
+          label="Description")
+        (hash anchorId="proposed-actions"
+          label="Proposed Actions")
+        (hash anchorId="analysis"
+          label="Analysis")
+        (hash anchorId="attachments"
+          label="Attachments")
+        }}
+    />
```
allthesignals, line 244: I think we need to extract some of the less-form-related stuff so we can use the `saveableForm.Section` API

andycochran: Can we handle this is a separate PR? I just did this the same way it's done in the PAS form. But I did note that the StickySidebar needs some attention. We need to be able to use it more generally, not only for saveable forms. I want to refactor the StickySidebar component so that it's actually just a navigation component. The "sticky" part is just a class on its container, and we need to be able to render other things in that container.

andycochran:
I've updated this PR quite a bit to include @godfreyyeung's suggestions and to add components that clean up the template markup.

client/app/helpers/optionset.js
```diff
@@ -54,40 +68,34 @@ const OPTIONSET_LOOKUP = {
  * @return     {string, number, array or Object}
  */
 export function optionset([model, optionsetId, returnType, lookupToken]) {
-  const optionset = OPTIONSET_LOOKUP[model][optionsetId];
+  const optionset = OPTIONSET_LOOKUP[model][optionsetId]; // eslint-disable-line
```
allthesignals, line 71: gah, disable line? why???

client/app/helpers/optionset.js
```diff
@@ -54,40 +68,34 @@ const OPTIONSET_LOOKUP = {
...
     case 'code':
-      if (lookupToken) {
-        const optionById = optionset[lookupToken];
-
-        if (optionById) {
-          return optionById.code;
-        }
+      if (optionById) {
+        return optionById.code;
+      }
```
allthesignals, line 82: I really don't want to change this stuff right now. I'm not confident we're asserting that lookups are working correctly anyway.

Please use ember-composable-helpers and build on top of the existing code, or add your own helper if this doesn't do the job.

client/tests/integration/components/ui/answer-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Template block usage:
+    await render(hbs`
+      <Ui::Answer>
+        template block text
+      </Ui::Answer>
+    `);
+
+    assert.equal(this.element.textContent.trim(), 'template block text');
+  });
+});
```
allthesignals, line 26: Please add some assertions here

client/tests/integration/components/ui/answer-prompt-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Template block usage:
+    await render(hbs`
+      <Ui::AnswerPrompt>
+        template block text
+      </Ui::AnswerPrompt>
+    `);
+
+    assert.equal(this.element.textContent.trim(), 'template block text');
+  });
+});
```
allthesignals, line 26: Plz assert something real

client/tests/integration/components/ui/answer-field-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Template block usage:
+    await render(hbs`
+      <Ui::AnswerField>
+        template block text
+      </Ui::AnswerField>
+    `);
+
+    assert.equal(this.element.textContent.trim(), 'template block text');
+  });
+});
```
allthesignals, line 26: real assertion

client/app/helpers/optionset.js
andycochran: Removed. I think the was accidentally left in there. (@godfreyyeung and I had added it while improving this helper, but looks like it's not necessary now.)

client/tests/integration/components/ui/answer-test.js
andycochran: Updated the test to assert that passing optional `beside` boolean does/doesn't make the component have a grid layout.

client/tests/integration/components/ui/answer-prompt-test.js
andycochran: Updated the test to assert that passing optional `beside` boolean does/doesn't make the component have a grid cell layout.

client/tests/integration/components/ui/answer-field-test.js
andycochran: Updated the test to assert that passing optional `beside` boolean does/doesn't make the component have a grid cell layout.

client/app/helpers/optionset.js
andycochran: I paired on this with @godfreyyeung and we manually checked the other places where this helper is used to make sure if still functions as intended. This change allows the `optionset` helper to look up falsey values. In refactoring this helper, we realized that this `if()` was never actually necessary. If you want to keep it this helper unchanged, I'll need to back out all the changes that make use of it—which is fine with me, it just makes the template code more verbose.

allthesignals: Alright, if ya'll checked how it'd affect the other parts of the code, I think that's fine.

Thanks andy!

Issue #329, opened 2020-05-20
Reorganize Component Files


allthesignals:
![image.png](https://images.zenhubusercontent.com/5ca23922d9fcf5278d715977/d9d125af-1e8f-4203-8468-e179debfdfd9)

Let's meet to discuss how we want to organizze our components. How? domain-specific, reusable, etc?

godfreyyeung:
**Current structure is** 
- components/
  - ... lots of misc ui components (breadcrumbs, character counter, sign in)
  - packages/
    - ... various form components (applicant fieldset, attachments)
    - pas-form/
      - ...edit/show components
  - projects/
    - ... project card components
  - saveable-form/

**Possible proposed structure:**
- components/
  - forms/
    - ... common form components (character counter, applicant fieldset, attachments)
    - pas-form/
      - ... edit/show components
    - rwcds/
      - ... edit/show components
  - projects/
    - ... project card components
  - saveable-form/

andycochran:
Most of the components in the root `components/` folder could be organized into subfolders… 

```
components/
  auth/
    do-logout
    sign-in-button
  ui/
    breadcrumbs-item
    breadcrumbs
    confirmation-modal // Rename confirmation-message (It renders modal and buttons. The "message" is passed to it.)
    page-nav
  messages/
    assistance // Or this could go under ui/ until we have other reusable "messages"
  packages/
    land-use-actions
    project-geography
    proposed-development-types
  saveable-form/
    character-counter
    form-validator
```

This would leave only `saveable-form` and `validation-message` in the root `components/` folder. But `validation-message` seems unnecessary; it only renders a span, invoked in 1 place.

allthesignals:
that looks great!

andycochran:
PR forthcoming…

PR #449, opened 2020-06-26
Sidebar refactor

andycochran:
This PR… 
- Adds a `<NeedAssistanceMessage />` component (which has DCP contact into that's repeated in all the edit/show form sidebars)
- Adds comments to the edit/show form templates to indicate where the main/sidebar columns start and end
- Renames the `<StickySidebar>` component to `<PageNav>` (for a couple reasons):
  - The `.sticky-sidebar` is only a class on the right grid column that makes it sticky
  - This component itself isn't what needs to be sticky; it doesn't need to be concerned with grid layout
  - The sidebar grid column needs to contain other content that isn't part of form buttons and page navigation (and that content also needs to be sticky)
- Moves the `.sticky-sidebar` class out of the component that handles the form buttons and page navigation


andycochran:
@allthesignals I've renamed it, but I don' think it belongs under the `packages/` folder, as it's also used for the project template and would be included on other non-package templates as we build them out.

allthesignals:
> @allthesignals I've renamed it, but I don' think it belongs under the `packages/` folder, as it's also used for the project template and would be included on other non-package templates as we build them out.

Okay yeah you're right. Don't you think the /components folder is getting a little messy? Is there another folder it might go under? I need some help keeping the components folder from getting out of control

andycochran:
Maybe `<Ui::Messages::Assistance />`?

allthesignals:
> Maybe `<Ui::Messages::Assistance />`?

But it's pretty domain-specific, right?

andycochran:
Adding a note to #329 ("Reorganize Component Files").

Issue #223, opened 2020-05-12
RWCDS Proposed Action(s) Section


nnxka:
[Azure Requirement 10972](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/10972)
[AB#10972](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10972)
[AB#10974](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10974)

#### Zoning Actions subsection
1. The user sees a list of actions that they previously provided to NYC Planning listed in the form section for them
2. Depending on the type of zoning action, there are form inputs where the user needs to submit additional details
3. The user fills in missing details about the missing actions where it is necessary
4. If the user wants to add or delete zoning actions types, they have to contact DCP and follow the instructions

#### General Zoning Questions subsection
1. Depending on the users' zoning actions, some questions will be displayed
2. Regardless of the zoning actions, there are a few questions that are always asked

### [Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP---Applicant-Portal?node-id=0%3A1)
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/ea39475c-c72a-4155-91d9-5ce1ed5bbb9d)

#### "Zoning Action Type" subsection
  - Form should display all of the actions from the projects actions `dcp_projectaction`. Depending on the type of action, the form needs to show additional field inputs. Please refer to wireframe to see which actions have conditional questions. 
#### "Which actions from other agencies" question
  - If applicant answers TRUE for previous "Is applicant seeking action question", then show "which actions from other agencies" question and write to `dcp_whichactionsfromotheragenciesaresought`

#### General Zoning Questions subsection
There are 1 condition question and one piece of conditional text that gets displayed
1. Conditional text: Mandatory Inclusionary Housing (MIH)
2. Conditional question: Which actions from other agencies

### Attributes
  - [ ] `dcp_zoningresolution` (project action, entity = `dcp_projectaction` or 10127)
  - [ ] `dcp_zrmodifyingzrtxt` (project action, entity = `dcp_projectaction` or 10127)
  - [ ] `dcp_actiontype` (action, entity = `dcp_action` or 10094)

### Filter Logic
[...]

PR #454, opened 2020-06-30
223 create syncactions method for proposed actions

trbmcginnis:
- set up backend patch for updating `dcp_affectedzoningresolution` entity
- when rwcds' `find` method is run, a new method `syncActions` will copy `dcp_projectaction` entity associated with current project to `dcp_affectedzoningresolution`. The frontend then only interacts with `dcp_affectedzoningresolution`

Addreses issue #223 (reference this issue for scope of problem).

### DIFF
--- a/server/src/packages/rwcds-form/rwcds-form.service.ts
+++ b/server/src/packages/rwcds-form/rwcds-form.service.ts
@@ -49,9 +49,14 @@ export class RwcdsFormService {
     const { records: [rwcdsForm] } = await this.crmService.get(`dcp_rwcdsforms`, `
       $filter=
         dcp_rwcdsformid eq ${id}
+      &$expand=dcp_rwcdsform_dcp_affectedzoningresolution_rwcdsform
     `);
 
     const { _dcp_projectid_value } = rwcdsForm;
+
+    // run syncActions to synchronize dcp_projectaction and dcp_affectedzoningresolution entities
+    const affectedZoningResolutions = rwcdsForm.dcp_rwcdsform_dcp_affectedzoningresolution_rwcdsform;
+    this.syncActions(_dcp_projectid_value, id, affectedZoningResolutions);
 
     // requires info from adjacent latest pasForm
     if (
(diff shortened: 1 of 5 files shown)


server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -84,6 +84,19 @@ export class RwcdsFormService {
   }
 
   async find(id) {
+    // Because we post data from dcp_projectaction to dcp_affectedzoningresolution, we need
+    // to GET the rwcdsForm twice. Once to pass the form into the syncActions method. And 
+    // a second time to include the updated dcp_affectedzoningresolution relationship. 
+    const { records: [rwcdsFormWithoutUpdatedZr] } = await this.crmService.get(`dcp_rwcdsforms`, `
+      $filter=
+        dcp_rwcdsformid eq ${id}
+      &$expand=dcp_rwcdsform_dcp_affectedzoningresolution_rwcdsform
```
godfreyyeung, line 93: It looks like you can delete the expand here, since the second rwcdsForm GET also expands `dcp_rwcdsform_dcp_affectedzoningresolution_rwcdsform`. The first expand of affectedzoningresolution is to support creating `zrTypes` within `syncActions()` , but I think you can just use the expand from the second GET.

trbmcginnis: But don't we need the first `expand` to be able to access `dcp_affectedzoningresolution` in order to pass it into `syncActions`

godfreyyeung: Thanks for clarifying this during standup! I think I didn't have enough coffee on Thursday.

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -74,4 +126,40 @@ export class RwcdsFormService {
...
+    const zrTypes = affectedZoningResolutions.map(zr => zr['dcp_zoningresolutiontype@OData.Community.Display.V1.FormattedValue']);
+    const { records: projectActions } = await this.crmService.get(`dcp_projectactions`, `
+      $filter=_dcp_project_value eq ${_dcp_projectid_value}
+    `);
+
+    await Promise.all(projectActions.map(action => {
+      const projectActionLabel = action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'];
+
+      // Lookup: dcp_affectedzoningresolutions asks for a ZR type,
+      const [actionLabel, actionCode] = ZONING_RESOLUTION_TYPES.find((label) => label === projectActionLabel) || ['ZA', 717170000];
```
allthesignals, line 149: I might have stored this lookup data a little differently. I think an object would be just fine:

```
const ZONING_RESOLUTION_TYPES = {
  ZA: 717170000,
  // others
};
```

Then,

```
const [actionLabel, actionCode] = ZONING_RESOLUTION_TYPES[projectActionLabel] || ZONING_RESOLUTION_TYPES.ZA;
```

That way we don't have to loop or put constants inside the code after the `||` pipes.

Also, `projectActionLabel` is the lookup key as well as the label, so it doesn't seem worth it to have another value.

PR #456, opened 2020-07-02
223: add zrsectiontitle question to proposed actions component

trbmcginnis:
- Add `dcpZrsectiontitle` as new field on `affected-zoning-resolution` model
- Add new input question on rwcds form for `dcpZrsectiontitle`
- update styling on `proposed-actions-editor` component and `with-action-no-action` component

Addresses #223

### DIFF
--- a/client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
+++ b/client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
@@ -1,39 +1,62 @@
-{{#let @form as |Form|}}
-  <Ui::Question class="fieldset relative scale-fade-in" as |Q|>
-    <span class="label">
-      {{~@form.saveableChanges.dcpZoningresolutiontype~}}
-      {{#if (lookup-action-type @form.saveableChanges.dcpZoningresolutiontype)}}
-        <LabsUi::IconTooltip
-          @tip={{lookup-action-type @form.saveableChanges.dcpZoningresolutiontype}}
-          @icon="info-circle"
-          @side="bottom"
-          @transform="shrink-4 up-2"
-          @fixedWidth={{true}}
-        />
+{{#let @zrForm as |zrForm|}}
+  <Ui::Question @fieldset={{true}} class="fieldset" as |Q|>
+    <Q.Prompt>
+      <span class="label">
+        {{~zrForm.saveableChanges.dcpZoningresolutiontype~}}
+        {{#if (lookup-action-type zrForm.saveableChanges.dcpZoningresolutiontype)}}
+          <LabsUi::IconTooltip
+            @tip={{lookup-action-type zrForm.saveableChanges.dcpZoningresolutiontype}}
+            @icon="info-circle"
+            @side="bottom"
+            @transform="shrink-4 up-2"
+            @fixedWidth={{true}}
+          />
+        {{/if}}
+      </span>
+      {{#if this.zrSectionNumber}}
+        <span data-test-zrsectionnumber>{{this.zrSectionNumber}}</span>
       {{/if}}
-    </span>
-    {{#if this.zrSectionNumber}}
-      <span data-test-zrsectionnumber>{{this.zrSectionNumber}}</span>
-    {{/if}}
-    {{#if this.hasModifiedZrSectionNumberQuestion}}
-      <Q.Prompt>
-        Which sections of the Zoning Resolution does this modify?
-      </Q.Prompt>
-      <small class="help-text display-block">
-        Provide the Zoning Resolution section number(s). Ex. ZR Sec. 42-10 and 43-17
-      </small>
-      <Form.Field
-        @attribute="dcpModifiedzrsectionnumber"
-        as |TextInput|
-      >
-        <TextInput
-          type="text"
-          maxlength="250"
-          data-test-input="dcpModifiedzrsectionnumber"
-        />
-      </Form.Field>
-    {{else if this.zrAppendixF}}
-      This action {{if @rwcdsForm.saveableChanges.dcpIncludezoningtextamendment "includes" "does not include"}} MIH. If this is a mistake, please contact City Planning. 
+    </Q.Prompt>
+    {{#if (or this.hasModifiedZrSectionNumberQuestion this.hasZrSectionTitleQuestion)}}
+      {{#if this.hasModifiedZrSectionNumberQuestion}}
+        <Q.Prompt>
+          Which sections of the Zoning Resolution does this modify?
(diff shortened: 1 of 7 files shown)


client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -1,39 +1,62 @@
 {{#let @form as |Form|}}
```
godfreyyeung, line 1: I know this isn't part of the diff, but can we change this to lowercase `|form|` here, too?

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -1,39 +1,62 @@
...
-          @icon="info-circle"
-          @side="bottom"
-          @transform="shrink-4 up-2"
-          @fixedWidth={{true}}
-        />
+  <Ui::Question @fieldset={{true}} class="fieldset" as |Q|>
+    <Q.Prompt>
+      <span class="label">
+        {{~@form.saveableChanges.dcpZoningresolutiontype~}}
+        {{#if (lookup-action-type @form.saveableChanges.dcpZoningresolutiontype)}}
```
godfreyyeung, line 6: We should be decisive about whether to use `@form` or the block param  (from `as |form|` ). If we introduce the block param, we don't need to use the argument `@form` here. Or, I think we can just get rid of `@form as |Form|` here, and use `@form` throughout.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
trbmcginnis: for sure!

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
trbmcginnis: okay good point, I'll just use `@form as |Form|` and then use `form` throughout

client/app/components/packages/rwcds-form/proposed-project-actions-editor.js
```diff
@@ -1,12 +1,12 @@
...
 
 export default class ProposedActionsComponent extends Component {
-  actionsWithSectionNumber = [
+  actionsWithSectionNumberAndSectionTitle = [
     'ZS', 'ZA', 'ZC', 'SD', 'SA', 'SC', 'RS', 'RA', 'RC',
   ];
 
   actionsWithModifiedZrSectionNumberQuestion = [
-    'ZS', 'ZA', 'SD', 'SA', 'RS', 'RA',
+    'ZA', 'ZS', 'SD', 'SA', 'RA', 'RS',
```
godfreyyeung, line 9: Thanks for alphabetizing!!! :D ❤️

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -1,39 +1,62 @@
...
+        >
+          <TextInput
+            type="text"
+            maxlength="100"
+            data-test-input="dcpZrsectiontitle"
+          />
+        </form.Field>
+      {{/if}}
+      {{#if this.zrAppendixF}}
+        This action {{if @rwcdsform.saveableChanges.dcpIncludezoningtextamendment "includes" "does not include"}} MIH. If this is a mistake, please contact City Planning.
```
godfreyyeung, line 58: Sorry, I should have caught this last time. This here should be just `form.saveableChanges. dcpIncludezoningtextamendment`, right? Not `@rwcdsform.saveableChanges.dcpIncludezoningtextamendment`

godfreyyeung: Otherwise, this PR looks great!! Thanks Taylor!

trbmcginnis: So we need two models in this component. 1) affected-zoning-resolution and 2) rwcds-form. 

This attribute does not come from affected-zoning-resolution but instead is an outlier and comes from rwcds-form so I passed in rwcds-form as well.

affected-zoning-resolution has a MANY to ONE relationship with rwcdsForm. But based on how the component is set up (with validations on saveableForm and passing in one affected-zoning-resolution at a time), it didn't make sense to me to pass in rwcdsForm.affectedZoningResolutions. Since it's just that one little display (and not based on anything being patched or changed) I figured I'd just pass in rwcdsForm as well.

Let me know if you disagree with this approach. I tried to implement the one that felt easiest to understand.

client/app/models/affected-zoning-resolution.js
```diff
@@ -12,4 +12,7 @@ export default class ProjectActionModel extends Model {
```
godfreyyeung, line 12: Should the class name here be `AffectedZoningResolutionModel`, instead of `ProjectActionModel`?

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
godfreyyeung: Ohhh. Sorry my bad, thank you for explaining. This is an interesting caveat with our nested form design. If we want a component to interact with multiple forms, we have to pass each one into that component separately.

godfreyyeung: I think we should leave it this way for this PR for now, since it follows the current pattern. Though we should definitely think about how we can make this workflow a bit nicer. 

I did just notice how the forms are passed into the editor:
**proposed-project-actions.hbs**
```
          <Packages::RwcdsForm::ProposedProjectActionsEditor
            @form={{affectedZoningResolutionForm}}
            @rwcdsForm={{@form}}
          />
```

I don't know what's best.... Just thinking out loud but perhaps
- If we do decide to pass each form in as a separate parameter, it would be cleaner to make the parameter names consistently expressive. `@form`  should probably be something like `@zrForm`
- Or perhaps we can pass in a hash...
```
          <Packages::RwcdsForm::ProposedProjectActionsEditor
            @forms={{hash
                 zForm=affectedZoningResolutionForm
                 rwcdsForm=@form
             }}
          />
```

trbmcginnis: Okay that's a good point! I'll update the form names

client/app/models/affected-zoning-resolution.js
trbmcginnis: Oh whoops yes it should -- looks like that was left over from previous stuff

PR #471, opened 2020-07-07
add RequiredAsterisk component and call it in prompts

andycochran:
This PR adds a `<Ui::RequiredAsterisk />` component which is called within prompts to indicate that its field is required. 

```
<Ui::Question as |Q|>
  <Q.Prompt>
    What's your name?
    <Ui::RequiredAsterisk />
  </Q.Prompt>
  ...
</Ui::Question>
```

It inserts a FontAwesome asterisk icon with a tooltip. 

image

-----

NOTE: As is, the `<Ui::Question>` component is not aware of the `<Form.Field>` and this PR is more evidence that they could be refactored as one component that generates both the prompt and the field(s), auto generating `for=""` attributes. We're already flagging in the "saveable" validations which fields are required, and this PR flags the labels separately. It would be nice if we could easily keep the prompt asterisk in sync with the validation.

### DIFF
--- a/client/app/components/packages/applicant-fieldset.hbs
+++ b/client/app/components/packages/applicant-fieldset.hbs
@@ -53,6 +53,7 @@
         <Ui::Question as |Q|>
           <Q.Prompt>
             <span class="text-weight-normal">First Name</span>
+            <Ui::RequiredAsterisk />
           </Q.Prompt>
 
           <Form.Field @attribute="dcpFirstname"/>
@@ -63,6 +64,7 @@
         <Ui::Question as |Q|>
           <Q.Prompt>
             <span class="text-weight-normal">Last Name</span>
+            <Ui::RequiredAsterisk />
           </Q.Prompt>
 
           <Form.Field @attribute="dcpLastname"/>
@@ -84,6 +86,7 @@
     <Ui::Question as |Q|>
       <Q.Prompt>
         <span class="text-weight-normal">Email Address</span>
+        <Ui::RequiredAsterisk />
       </Q.Prompt>
 
       <Form.Field @attribute="dcpEmail"/>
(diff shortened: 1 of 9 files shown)


client/app/components/ui/required-asterisk.hbs
```diff
@@ -0,0 +1,6 @@
+<sup>
+  <FaIcon @icon="asterisk" class="text-red text-weight-normal" @transform="shrink-4" />
```
godfreyyeung, line 2: Is this icon accessible? Should we add a hidden * screen readers?

client/app/components/ui/required-asterisk.hbs
```diff
@@ -0,0 +1,6 @@
...
+  <EmberTooltip>
```
godfreyyeung, line 3: What's the difference between this and `<LabsUi:Icontooltip>` ?
We should be consistent with use of components if possible

andycochran: ! Good call. Thanks. I'll revise.

client/app/components/ui/required-asterisk.hbs
andycochran: I'll switch to using `<LabsUi::IconTooltip />` and it will be accessible. Screen readers can read the tooltip text in the Ember Tooltip.

client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -53,6 +53,7 @@
         <Ui::Question as |Q|>
           <Q.Prompt>
```
allthesignals, line 54: So, I might have made this a part of the `Prompt` component, with an option:

```handlebars
<Q.Prompt @required={{true}}>
  Name of your cat?
</Q.Prompt>
```

PR #475, opened 2020-07-08
313 Enforce data types for applicant fields

godfreyyeung:
Use changeset validator to regulate format for phone and email fields.
Use HTML Number field to force number entry for zipcode.

Addresses #313. Addresses #298.

### DIFF
--- a/client/app/components/packages/applicant-fieldset.hbs
+++ b/client/app/components/packages/applicant-fieldset.hbs
@@ -139,7 +139,10 @@
             <span class="text-weight-normal">ZIP</span>
           </Q.Prompt>
 
-          <Form.Field @attribute="dcpZipcode"/>
+          <Form.Field
+            @attribute="dcpZipcode"
+            type="number"
+          />
         </Ui::Question>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


hannahkates:
This also addresses #298.

godfreyyeung:
@hannahkates For phone numbers, do you think it's fine to strictly mandate a 10-digit format?

hannahkates:
@godfreyyeung Hema's comment said that the phone in CRM is a numeric with a max of 10 digits, so we should use that for the validation.

godfreyyeung:
> @godfreyyeung Hema's comment said that the phone in CRM is a numeric with a max of 10 digits, so we should use that for the validation.

oh, perfect!

godfreyyeung:
> Would you mind adding these validations to the `pas-form-validation-messages-display-test`? It's kind of a hefty test but might be nice to have these in there. I don't think applicant validations have been added to that test yet

Sorry saw this after I merged! I'll put it up in a new PR.

allthesignals:
@hannahkates is this still an issue? https://github.com/NYCPlanning/labs-applicant-portal/issues/313

Did this PR not fix it?

PR #487, opened 2020-07-10
313 Add tests for applicant validation

godfreyyeung:
Also use HTML to prevent user from entering letters for
dcpPhone

### DIFF
--- a/client/tests/acceptance/pas-form-validation-messages-display-test.js
+++ b/client/tests/acceptance/pas-form-validation-messages-display-test.js
@@ -19,7 +19,7 @@ module('Acceptance | pas form validation messages display', function(hooks) {
     });
   });
 
-  test('Certain fields display both Saveable and Submittable validation errors', async function(assert) {
+  test('Certain PAS Form fields display both Saveable and Submittable validation errors', async function(assert) {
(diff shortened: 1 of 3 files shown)


client/tests/acceptance/pas-form-validation-messages-display-test.js
```diff
@@ -155,4 +155,79 @@ module('Acceptance | pas form validation messages display', function(hooks) {
...
+    this.server.create('applicant', 'individualApplicant', { pasForm });
+
+    this.applicants = [
+      await this.owner.lookup('service:store').findRecord('applicant', 1),
+    ];
+
+    await visit('/pas-form/1/edit');
+
+    // email
+    assert.dom('[data-test-validation-message="dcpEmail"]').doesNotExist();
```
trbmcginnis, line 172: Doesn't email also have a "field is required" validation coming from `submittable-applicant-form.js`? Were you thinking of testing for that as well?

godfreyyeung: @trbmcginnis  Great comment! Thank you. This made me also realize I should revise how I created the applicants within the tests. I've now added tests for the required validation, which should definitely be included according to the test title.

Issue #479, opened 2020-07-09
RWCDS: Missing history question


hannahkates:
This question is missing from the form.
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/cdf8ef3a-2bf2-4785-84f4-7b2501c879c2)

PR #488, opened 2020-07-10
479 Add field for build year rational

godfreyyeung:
Also update copy based on updates to wireframe

Addresses #479  
The issue was not that the input for `dcpSitehistory` was missing. It was that it was accidentally associated with the question for build year rational, and there was no input for `dcpRationalbehindthebuildyear`

### DIFF
--- a/client/app/components/packages/rwcds-form/project-description.hbs
+++ b/client/app/components/packages/rwcds-form/project-description.hbs
@@ -93,16 +93,45 @@
         Explain the rationale behind the analysis year
       </strong>
       <small class="help-text display-block">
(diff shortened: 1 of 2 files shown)


andycochran:
I think this might be handled by #489, except for the test. Let's get it merged first, then we might just need the test from this PR.

godfreyyeung:
> I think this might be handled by #489, except for the test. Let's get it merged first, then we might just need the test from this PR.

Can we get this merged first since it's such a simple PR? #489 is doing a lot. This PR directly addresses a specific issue.

PR #489, opened 2020-07-10
379 - Ui::Question Refactor to implement RWCDS instructions

andycochran:
This PR completes the RWCDS instructions and includes a necessary refactor of the `Ui::Question` component in order to do so. Unfortunately, this touches a ton of files, so please review carefully and let me know if you have questions (pun intended). 

**API/component changes:**

- `Ui::Question` no longer takes a `@fieldset` argument — All questions are now `<fieldset>` elements. Previously, questions were either `<fieldset>` elements (for grouped radios, checkboxes) or `<label>` elements (for non groups) which wrapped the entire question. This caused trouble when block content including links or other interactivity was wrapped by a clickable `<label>`. A `<fieldset>` element is still semantic for wrapping non-grouped inputs, as they also include other contextual pieces (such as help text, additional guidance, counters, validation messages, etc). 

   Note: Some questions do not need to be wrapped in `Ui:Question` components — e.g. the applicant contact info does not require the same markup/classes/styles as the main form questions, so they can be simple inputs wrapped in labels. 

- The abstract `<Q.Prompt>` has been replaced by `<Q.Legend>` and `<Q.Label>` — Instead of hiding the logic for whether a prompt is a `<legend>` in a `<fieldset>` or `<span>` in a `<label>`, we call them explicitly in the template. 

- `Ui::Question` takes a `@required` boolean and passes it to the contextual `<Q.Legend>` and `<Q.Label>`. 

- `Ui::Question` generates a `questionId`  that is passed to`<Q.Label>` as a `for=""` value. To make labels clickable, `id={{Q.questionId}}` should be passed to the associated `<Form.Field>`. 

- Improves the layout of the `Ui::CollapsibleText` component (plus a new/improved animation) 

- The `Ui::CollapsibleText` component can be called within `Ui::Question` with `<Q.AdditionalGuidance>` (which standardizes the button text)

- Adds a `.q-help` class to style the "help text" between labels and inputs. This class should be used instead of `.help-text` as [defined in our style guide](https://nyc-planning-style-guide.netlify.app/forms/#help-text), as it's meant to be used below inputs. (Per customer request, we're using a pattern outside style guide recommendations.)

**Content/instructions changes:** 

- Makes all the RWCDS instructions match the wireframes. (leveraging `<Q.Helptext>` and `<Q.AdditionalGuidance>`) 

- Revised the Attachment section layout: 

  - Indicate file size limit on all forms (not just RWCDS)

  - Move the upload UI above the detailed instructions (users were confused by the long instruction causing them to not see the upload UI)

- Improves the PAS, RWCDS form header styles


andycochran:
😩 Once again a `<U+2028>` character copied from Figma had me hunting to find why tests fail.

client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -36,12 +36,10 @@
...
-      <Ui::Question data-test-applicant-organization as |Q|>
-        <Q.Prompt>
-          <span class="text-weight-normal">Organization Name</span>
-        </Q.Prompt>
-        <Form.Field @attribute="dcpOrganization"/>
-      </Ui::Question>
+      <label data-test-applicant-organization>
+        Organization Name <Ui::RequiredAsterisk />
+        <Form.Field @attribute="dcpOrganization" />
+      </label>
```
godfreyyeung, line 42: Why are we using just <label> here, and not Ui::Question and Q.label?

godfreyyeung:
>   Note: Some questions do not need to be wrapped in `Ui:Question` components — e.g. the applicant contact info does not require the same markup/classes/styles as the main form questions, so they can be simple inputs wrapped in labels.
Can you explain more?

client/app/components/packages/rwcds-form/with-action-no-action.hbs
```diff
@@ -128,68 +146,75 @@
...
+      <Q.Label>
         How did you determine this With-Action scenario?
-      </Q.Prompt>
-      <p class="help-text">
+      </Q.Label>
+
+      <Q.HelpText>
         Explain the rationale supporting the With-Action Scenario described above. If future conditions are not "maximized", please indicate why this is the case (e.g., existing site controls, site configuration).
-      </p>
+      </<Q.HelpText>
```
godfreyyeung, line 211: This looks like a typo/error

client/app/components/packages/land-use-action.hbs
```diff
@@ -25,47 +25,56 @@
...
-                Ui::Question component doesn't work when a grid splits up label and input.
-                The label needs a functioning for="" if it doesn't wrap the input.
-              --}}
-              <label class="middle">
-                <strong>How many {{landUseAction.name}} actions?</strong>
-                <Ui::RequiredAsterisk />
-              </label>
+      </Ui::Legend>
+
+      {{#if (or
```
godfreyyeung, line 33: Thank you for the reformatting! 🙏  Looks so much clearer. However this highlights  that these land use names should be moved into an array within the component. like `["Zoning Special Permit", "Zoning Certification"]` 
Then in the template we can use a helper like `{{in-array}}`

client/app/components/ui/question.hbs
```diff
@@ -1,13 +1,9 @@
...
-      Prompt=(component 'ui/label-text')
-    )}}
-  </label>
-{{/if}}
+<fieldset class="ui-question" ...attributes>
+  {{yield (hash
+    questionId=this.questionId
+    Legend=(component 'ui/legend' required=@required)
+    Label=(component 'ui/label' required=@required for=this.questionId)
+    HelpText=(component 'ui/help-text')
```
godfreyyeung, line 6: Food for thought: Here, HelpText is yielded from this component just because it is related to the Question component and other yields. From a developer standpoint, they use  `Q.HelpText` instead of `UI::HelpText` just because the help text is within a `UI::Question`. This is a nuance that may not be immediately obvious to the developer. We should add docs

client/app/components/ui/help-text.hbs
```diff
@@ -0,0 +1,3 @@
+<p class="text-small tiny-margin-bottom" ...attributes>
```
godfreyyeung, line 1: I wonder if help-text can just be a class

andycochran: `.help-text` already is a class defined in our style guide. But (per the customer's request) we are not following the [style guide's recommendation](https://nyc-planning-style-guide.netlify.app/forms/#help-text) (in which help text is below the input). We'd need to come up with a more semantic class name for this custom thing we're doing — like `.inline-guidance` or `.sub-label`… or maybe modify `.help-text` with a second class like `.help-text.above`.

client/app/components/packages/rwcds-form/with-action-no-action.hbs
andycochran: Ack! Fixed.

client/app/components/packages/land-use-action.hbs
andycochran: Let's create a separate issue for this?

client/app/components/ui/question.hbs
andycochran: You're right. I don't think it's necessary for this to be a contextual component. Or even a component at all. I'm going to look into just making it a class.

client/app/components/ui/help-text.hbs
godfreyyeung: Oh, I see. Though, I guess guides are just guides in the end and we have a customer request :p so maybe using it above an input is okay? I like the `.help-text.above` idea!

client/app/components/packages/land-use-action.hbs
godfreyyeung: Sounds good

client/app/components/packages/applicant-fieldset.hbs
andycochran: The main benefit of `Ui::Question` is to standardize our markup, avoid duplicating classes that style labels, and to avoid labels wrapping inputs (by leveraging `for=""`) so that questions can contain complex/interactive UI (help text, additional guidance, validation, counter). 

In order to give visual hierarchy to the applicant info, its inputs are not styled the same as top-level questions. And they don't require any classes, or contain any complex help text or additional guidance. They're much simpler. 

In order to use `Ui::Question` for these simple inputs, we'd need to complicate the component with a way to conditionally style questions — with regular instead of bold labels, and without margin between the questions. 

I'd rather have these few inputs done differently than overcomplicate the `Ui::Question` API.

godfreyyeung: Okay 👍 Thanks for the explanation! I really think we should have documentation laying out the different uses of these components

godfreyyeung: I'll create an issue for that

trbmcginnis:
Any chance we could get a readme for these similar to how we have a readme for our optionset helper? I feel like the PR description could actually work as a readme -- maybe with some examples added

client/app/components/packages/applicant-fieldset.hbs
```diff
@@ -1,23 +1,23 @@
 {{#let @form as |Form|}}
   <Ui::Question
-    @fieldset={{true}}
     class="fieldset relative"
     ...attributes
     as |Q|
   >
-    <Q.Prompt data-test-applicant-title="{{@applicant.friendlyEntityName}}">
+    <Q.Legend data-test-applicant-title="{{@applicant.friendlyEntityName}}">
```
allthesignals, line 7: "Legend" makes less sense to me. I really liked "Prompt" because it adds clarity to what we're trying to do...

Issue #482, opened 2020-07-09
RWCDS: Conditional logic is wrong for no-action, with-action questions


hannahkates:
The logic was mislabeled on the wireframes. They are corrected now.

**No-Action**
- The "Description of No-Action scenario" is currently set up to only conditionally appear if the answer to the previous question is no. This is incorrect. It should always appear.
- "How did you determine this No-Action scenario?" should appear when the previous answer is no, not yes.

**With-Action**
- "Description of With-Action scenario" is currently set up to only conditionally appear if the answer to the previous question is no. This is incorrect. It should always appear.

PR #491, opened 2020-07-13
482: fix with-action-no-action conditional display logic

trbmcginnis:
Fix the conditional display logic for the with-action-no-action component. 

Addresses #482

### DIFF
--- a/client/app/components/packages/rwcds-form/with-action-no-action.hbs
+++ b/client/app/components/packages/rwcds-form/with-action-no-action.hbs
@@ -51,9 +51,9 @@
             (hash label='Yes' code=true)
             (hash label='No' code=false)
           }}
-          @onChange={{fn (mut @form.saveableChanges.dcpDescribethenoactionscenario) ""}} as |value|
+          @onChange={{fn (mut @form.saveableChanges.dcpHowdidyoudeterminethenoactionscenario) ""}} as |value|
         >
-          {{#if value}}
+          {{#if (eq value false)}}
             <Ui::Question as |Q|>
               <Q.Prompt>
                 How did you determine this No-Action scenario?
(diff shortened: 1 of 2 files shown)


client/app/components/packages/rwcds-form/with-action-no-action.hbs
```diff
@@ -51,9 +51,9 @@
             (hash label='Yes' code=true)
             (hash label='No' code=false)
           }}
-          @onChange={{fn (mut @form.saveableChanges.dcpDescribethenoactionscenario) ""}} as |value|
```
godfreyyeung, line 54: so this was previously a mistake?

client/tests/acceptance/user-can-click-rwcds-form-edit-test.js
```diff
@@ -129,32 +129,30 @@ module('Acceptance | user can click rwcds edit', function(hooks) {
...
 
+    // no-action
     await fillIn('[data-test-textarea="dcpDevelopmentsiteassumptions"]', exceedMaximum(2400, 'String'));
     assert.dom('[data-test-validation-message="dcpDevelopmentsiteassumptions"').hasText('Text is too long (max 2400 characters)');
 
-    await click('[data-test-radio="dcpExistingconditions"][data-test-radio-option="Yes"]');
+    await fillIn('[data-test-textarea="dcpDescribethenoactionscenario"]', exceedMaximum(1500, 'String'));
+    assert.dom('[data-test-validation-message="dcpDescribethenoactionscenario"').hasText('Text is too long (max 1500 characters)');
+
+    await click('[data-test-radio="dcpExistingconditions"][data-test-radio-option="No"]');
```
godfreyyeung, line 139: Why do we click "No"  after testing the validation for  dcpDescribethenoactionscenario ? should it be before, since dcpDescribethenoactionscenario should appear regardless of whether dcpExistingconditions is yes or no?

godfreyyeung: I might just not be understanding the form business requirements

client/tests/acceptance/user-can-click-rwcds-form-edit-test.js
```diff
@@ -129,32 +129,30 @@ module('Acceptance | user can click rwcds edit', function(hooks) {
...
     assert.dom('[data-test-validation-message="dcpHowdidyoudeterminethenoactionscenario"').hasText('Text is too long (max 1500 characters)');
 
-    await click('[data-test-radio="dcpExistingconditions"][data-test-radio-option="No"]');
+    // with-action
+    await fillIn('[data-test-textarea="dcpDescribethewithactionscenario"]', exceedMaximum(1500, 'String'));
+    assert.dom('[data-test-validation-message="dcpDescribethewithactionscenario"').hasText('Text is too long (max 1500 characters)');
 
-    await fillIn('[data-test-textarea="dcpDescribethenoactionscenario"]', exceedMaximum(1500, 'String'));
-    assert.dom('[data-test-validation-message="dcpDescribethenoactionscenario"').hasText('Text is too long (max 1500 characters)');
+    await fillIn('[data-test-textarea="dcpHowdidyoudeterminethiswithactionscena"]', exceedMaximum(600, 'String'));
```
godfreyyeung, line 148: 🏕️ 🧹 ? :)

client/app/components/packages/rwcds-form/with-action-no-action.hbs
trbmcginnis: It wasn't necessarily a mistake but it did only take into account ONE "reset" of a field when they're technically should have been TWO, since two fields were conditional. BUT now that only one field depends on this condition, I just needed to update it to reflect that one.

client/tests/acceptance/user-can-click-rwcds-form-edit-test.js
trbmcginnis: Correct, it does not depend on the `dcpExistingconditions` question. We click No afterwards because `dcpHowdidyoudeterminethenoactionscenario` only shows up when that radio question is No.

I put the test for all fields that are not conditional based on a question (will always be there) BEFORE we do anything with the radio buttons

godfreyyeung: Ah okay, thanks!

trbmcginnis:
Closing and remaking due to conflicts

PR #497, opened 2020-07-13
485 add what-input

andycochran:
This PR adds ember-cli-what-input, which improves `:focus` styles (using the CSS in the Style Guide) — e.g. it prevents outlines from displaying unless the user is navigating with a keyboard.


godfreyyeung:
Hmm, I'm wary of using this addon. It doesn't seem the benefits, which appear mostly to be appearance, outweigh the risks of using an outdated and low support, low popularity addon. 
On ember observer, this addon has a score of 1.5/10. Also the master branch was last updated september 2017?

andycochran:
The `ember-cli-what-input` addon might not be active/popular, but all it does is inject [What Input](https://www.npmjs.com/package/what-input) which _is_ popular/active. I don't think this addon is risky. There's no need for active development on it, as its only function is to require the latest version of What Input and add its path to a `<script>`. I suppose we could require What Input directly if you're wary of this addon.

allthesignals:
The only code it adds is here: https://github.com/MariusRumpf/ember-cli-what-input/blob/master/index.js

If you want to avoid an addon, you might try installing what-input without the addon. If that proves too hard to deal with, I think the addon is just fine.

andycochran:
If you're worried about accessibility, that's covered by the CSS from the style guide. What Input just adds a data attr to the window that our existing styles leverage — e.g. appropriately using `:focus` styles when users are navigating w/ the keyboard, only hiding them for mouse users.

PR #501, opened 2020-07-14
fix typo in proposed actions component

trbmcginnis:


client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -69,7 +69,7 @@
       {{/if}}
 
       {{#if this.zrAppendixF}}
-        This action {{if @rwcdsForm.saveableChanges.dcpIncludezoningtextamendment "includes" "does not include"}} MIH. If this is a mistake, please contact City Planning.
+        This action {{if (eq @rwcdsForm.saveableChanges.dcpIncludezoningtextamendment 717170000) "includes" "does not include"}} MIH. If this is a mistake, please contact City Planning.
```
godfreyyeung, line 72: Can we handle this with the optionset helper, so we don't have static codes within our templates?

trbmcginnis: Yep! Will do

trbmcginnis: Actually not sure using an optionset would be beneficial here since we're just checking that it's equal to one thing, not displaying a code/labe.

godfreyyeung: Maybe something like

```
 This action {{if (eq
    @rwcdsForm.saveableChanges.dcpIncludezoningtextamendment
    (optionset 'rwcdsForm' 'dcpIncludezoningtextamendment' 'code' 'YES'))
    "includes"...
    }}
```
Where 'YES' is the identifier (key) for the optionset value. I looked up the optionset in CRM. If registering it with the optionset helper, it would something look like:
```
YES: {
  code: 717170000,
  label: 'Yes',
},
NO: {
  code: 717170001,
  label: 'No',
},
DONT_KNOW: {
  code: 717170002,
  label: 'Don\u2019t Know',
}
```

godfreyyeung: I guess in this case it's a minor improvement, but it means we as humans don't have to remember that `717170000` translates to a yes answer. Also, for some fields 717170000 is actually the negative answer.

trbmcginnis: Okay yes that's how I was imagining it too -- but I figured that putting in 'YES' would still be hardcoding it (even if it's easier to understand than an actual code). Plus it would have the extra work of an onlooking developer to understand how the optionset worked / take up a lot more space. So it didn't feel particularly beneficial to change it. If you disagree I'll happily update it, but in my opinion just having `717170000` there is easier to read. Maybe I can add a comment above the code? Let me know your thoughts

client/app/components/packages/rwcds-form/proposed-project-actions-editor.js
```diff
@@ -18,7 +18,7 @@ export default class ProposedActionsComponent extends Component {
 
   get zrAppendixF() {
     const zrSectionNumber = this.args.zrForm.saveableChanges.dcpZrsectionnumber;
-    return this.zrTypeLabel === 'ZR' && zrSectionNumber === 'Appendix F';
+    return this.zrTypeLabel === 'ZR' && zrSectionNumber === 'AppendixF';
```
allthesignals, line 21: Let's use a lookup for this kind of thing — don't necessarily have to use optionset helper, just a dictionary will do

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
godfreyyeung: I see your point and think it's a fine distinction.

I think 'YES' is more maintainable, self-documenting and less error prone for developers. It's hard to track the difference between `717710000` and `71717000`, but it's easier to understand 'YES' or 'NO'. Also, probably most important, is having the code consistent throughout the app. It feels confusing to have some parts of the code use optionsets but not others. If I was someone new stepping into the code, I would be confused why that is the case.

Also if the CRM codes ever change, we'll only need to update them in the optionset constant in the model file.  
I'd still vouch for using the optionset.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.js
godfreyyeung: I'm gonna vouch for  `{{optionset}}` helper all the things :p to avoid errors like this.
Optionset helper will spit out a console assertion error if the passed in identifier/code/label doesn't exist.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
godfreyyeung: (sidenote: This is a situation where I wish Javascript had something called enums. They are basically dictionary "constants", which force you to choose from a limited set of identifiers ('YES', 'NO', 'MAYBE') each of which translate to a specific number/value.  )

trbmcginnis: Okay cool you make a good point. I think there are a couple places in the code where we are still hardcoding things like `717170000` so maybe another maintenance issue to update these would be good in the future. I'll make the changes! Thanks

trbmcginnis:
Handling optionset issues in #533

Issue #499, opened 2020-07-14
RWCDS: Unable to Submit RWCDS form


trbmcginnis:
When a user fills out all required fields on the RWCDS form, they should be able to submit. For some reason the Submit button remains disabled even after everything is filled out.

PR #508, opened 2020-07-15
499: fix submit on rwcds form

trbmcginnis:
Add `submitPackage` action to rwcds-form edit.js so that it is accessible in the template.

Add optionsets to rwcds-form for use on show page.

Addresses #499

### DIFF
--- a/client/app/helpers/optionset.js
+++ b/client/app/helpers/optionset.js
@@ -19,6 +19,9 @@ import {
 import {
   DCPHASPROJECTCHANGEDSINCESUBMISSIONOFTHEPAS_OPTIONSET,
   DCPCONSTRUCTIONPHASING_OPTIONSET,
+  DCPEXISTINGCONDITIONS_OPTIONSET,
+  DCPISRWCDSSCENARIO_OPTIONSET,
+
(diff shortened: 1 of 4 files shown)


client/app/helpers/optionset.js
```diff
@@ -19,6 +19,9 @@ import {
 import {
   DCPHASPROJECTCHANGEDSINCESUBMISSIONOFTHEPAS_OPTIONSET,
   DCPCONSTRUCTIONPHASING_OPTIONSET,
+  DCPEXISTINGCONDITIONS_OPTIONSET,
```
godfreyyeung, line 22: I don't see these properties on the rwcds-form model. Can you clarify where these will be used/for what part of the wrieframes?

trbmcginnis: `dcpIsrwcdsscenario` (line 45) and `dcpExistingconditions` (line 97) are both on the rwcds model. If you look at show.hbs, it's line 159 and 194.

godfreyyeung: Not sure why I couldn't find them earlier on the model. Thanks!!

Issue #358, opened 2020-05-26
PAS submission: require at least one applicant type="Applicant"


hannahkates:
[AB#10745](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/10745)

In order to submit the PAS form, the applicant needs to have at least one applicant team member where type="Applicant".

PR #510, opened 2020-07-15
358 Require at least one applicant to submit PAS

godfreyyeung:
Addresses #358 

The way we've designed our form architecture, there are two types of validations: 
  1. **Field-specific** Those sitting within validators (like `validators/submittable-pas-form.js` and `validators/saveable-rwcds-form.js`, etc.)
  2. **Entity/Macro-level** Those sitting within our package logic, namely the isDirty and isValid getters. These usually deal with macro-level validations. Validations which span across fields and entities (e.g. validating one or more attachment or applicant has been added) 

image

The Save and Submit buttons depend on both types of validations to determine if they are enabled.
image

image

For this issue, I added the isValid() getter to the package in order to add an "entity/macro-level validation" to support the Save enabled logic. This follows the pattern for the Submit button. (In the screenshot above, pasForm.isSubmittable is the field-level validation)

Note that I have a PR #513   to standardize and improve the variable names in the `pas-form/edit.hbs` file. Namely, renaming the SaveableForm block parameter from `|pasForm|` to `|saveableForm|` (like how we do it for `rwcds-form/edit.hbs`.  This will remove the potential confusion here as to the difference between `this.pasForm` and `pasForm`
image

However it's an extensive change, so needed to do it in a separate PR to keep this one clear and focused.

### DIFF
--- a/client/app/components/packages/pas-form/applicant-team.hbs
+++ b/client/app/components/packages/pas-form/applicant-team.hbs
@@ -54,5 +54,13 @@
         </div>
       </div>
     </div>
+    {{#unless @pasForm.hasApplicants}}
+      <span
+        class="form-error is-visible"
+        data-test-validation-message="has-applicants"
+      >
+        One or more applicant team members is required.
+      </span>
+    {{/unless}}
   </Form.Section>
 {{/let}}
--- a/client/app/components/packages/pas-form/edit.hbs
+++ b/client/app/components/packages/pas-form/edit.hbs
@@ -31,6 +31,7 @@
 
       <Packages::PasForm::ApplicantTeam
         @form={{pasForm}}
+        @pasForm={{this.pasForm}}
         @addApplicant={{this.addApplicant}}
         @removeApplicant={{this.removeApplicant}}
         @validations={{this.validations}}
@@ -63,7 +64,7 @@
 
         <Packages::PasForm::PasFormError @package={{@package}} />
         <pasForm.SubmitButton
-          @isEnabled={{pasForm.isSubmittable}}
+          @isEnabled={{and @package.isValid  pasForm.isSubmittable}}
           class="secondary"
           data-test-submit-button
         />
--- a/client/app/models/package.js
+++ b/client/app/models/package.js
@@ -227,4 +227,11 @@ export default class PackageModel extends Model {
 
     return isPackageDirty;
   }
+
+  get isValid() {
+    if (this.dcpPackagetype === PACKAGE_TYPE_OPTIONSET.PAS_PACKAGE.code) {
+      return this.pasForm.hasApplicants;
+    }
+    return true;
+  }
 }
--- a/client/app/models/pas-form.js
+++ b/client/app/models/pas-form.js
@@ -883,4 +883,8 @@ export default class PasFormModel extends Model {
 
     return dirtyApplicants.length > 0;
   }
+
+  get hasApplicants() {
+    return this.applicants.length > 0;
+  }
 }
--- a/client/tests/acceptance/pas-form-validation-messages-display-test.js
+++ b/client/tests/acceptance/pas-form-validation-messages-display-test.js
@@ -3,6 +3,7 @@ import {
   visit,
   fillIn,
   click,
+  waitFor,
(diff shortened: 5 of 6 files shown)


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -63,7 +64,7 @@
 
         <Packages::PasForm::PasFormError @package={{@package}} />
         <pasForm.SubmitButton
-          @isEnabled={{pasForm.isSubmittable}}
+          @isEnabled={{and @package.isValid  pasForm.isSubmittable}}
```
andycochran, line 67: Doesn't `isSubmittable` imply `isValid`? A form isn't submittable if it isn't valid.

client/app/components/packages/pas-form/applicant-team.hbs
```diff
@@ -54,5 +54,13 @@
...
       </div>
     </div>
+    {{#unless @pasForm.hasApplicants}}
+      <span
+        class="form-error is-visible"
+        data-test-validation-message="has-applicants"
+      >
+        One or more applicant team members is required.
+      </span>
+    {{/unless}}
```
allthesignals, line 64: I think this works nicely, but it seems like we're missing out on the features of changeset validations. Is it possible to validate the `.length` property of `applications`?

To me, this is a validation and should leverage the validator features we have. I'm happy to take a look! I've done a lot of research on ways to use changeset validations on related models.

client/app/components/packages/pas-form/edit.hbs
godfreyyeung: There are two different types of validity -- Validator (ember-changeset) validity, and macro/entity-level validity for things like number of related entities

client/app/components/packages/pas-form/applicant-team.hbs
godfreyyeung: That would be sweet, if you could do that! Maybe it's possible to use a custom validator. 
We would have to spike test if the oldValue and newValue parameters reflect related models
```
![image](https://user-images.githubusercontent.com/3311663/87803412-479cb500-c807-11ea-90e8-a233b5f84d8f.png)

```

godfreyyeung: Though would this be muddying the focus of the validators? Should we keep them focused to just field-specific validations? We've made an effort to move the changesets off of the models and into their own component. Would this potentially couple the changesets and models again?

allthesignals: > Though would this be muddying the focus of the validators? Should we keep them focused to just field-specific validations? We've made an effort to move the changesets off of the models and into their own component. Would this potentially couple the changesets and models again?

It is field-specific: the model has a field called `applicants` and we want to validate that it has at least one.

I don't quite follow your last point... changesets are really just a proxy of the model itself, but the proxy intercepts changes to it. When were changesets coupled to the model? I think it should behave just as it does with any other field. There's an issue in ember changeset validations addon about this, I'll just have to dig it up.

godfreyyeung: Oops sorry, I misspoke the changesets used to be created on the edit Packages::_Form::Edit component. I thought the changesets used to be on the package model itself. Scratch the model thing.

Okay, sounds good, thanks!

client/app/components/packages/pas-form/edit.hbs
allthesignals: Yeah but I think we're going to look at using ember changeset validations so that `isSubmittable` can capture validations on the # of models

allthesignals:
@godfreyyeung can I help with this? I did some research on this sort of thing before, so I can dig up what I found.

allthesignals:
https://davidtang.io/2018/07/28/tracking-has-many-relationship-changes-with-ember-changeset.html

allthesignals:
@godfreyyeung take a look at this and lemme know what you think: https://github.com/NYCPlanning/labs-applicant-portal/pull/530

allthesignals:
Closing in favor of https://github.com/NYCPlanning/labs-applicant-portal/pull/530

PR #511, opened 2020-07-15
Remove legacy use of data-test-textarea

godfreyyeung:
🧹🧹🧹 
Previously, we had two different test selectors, data-test-textinput
and data-test-textarea, for single-line inputs and text areas respectively.

However the new Field component assigns data-test-input as the selector
for all input types, even radio inputs. Thus we no longer need
data-test-textarea selectors on text areas. In fact, prior to this
commit, text areas had both data-test-textarea and data-test-input
selectors attached to the element.

### DIFF
--- a/client/app/components/saveable-form/field/text-area.hbs
+++ b/client/app/components/saveable-form/field/text-area.hbs
@@ -1,6 +1,5 @@
 <Textarea
   @value={{@value}}
-  data-test-textarea={{@attribute}}
   rows="5"
   ...attributes
 />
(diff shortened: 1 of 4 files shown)


andycochran:
Nice! `data-test-field=""` might be even more semantic, as there are both `<textarea>` tags and `<input>` tags (but that's nit-picky).

godfreyyeung:
> Nice! `data-test-field=""` might be even more semantic, as there are both `<textarea>` tags and `<input>` tags (but that's nit-picky).

That's a good suggestion! I wonder the others think

godfreyyeung:
> I'd prefer `data-test-field=""` — as it's assigned to `<textarea>` in addition to `<input>` elements, and since it's handled by the `Field` component — but approving. Looks good!

Okay! After this PR, all the selectors will be consistent, and it will be simple to find and replace `data-test-input` with `data-test-field` across the app! We can put up a separate PR for that

Issue #504, opened 2020-07-14
Standardize block parameter variable names


godfreyyeung:
We need to standardize our block parameters. They're a bit all over the place currently.

We should change
- `SaveableForm as |saveableForm|` instead of `... as |pasForm|`
- `|Form|` to `|form|` (use lower case)

godfreyyeung:
Sorry, realized I should have brought this up in Sprint Planning. Forgot about it since it got moved to backlog, but we need to fix this in this sprint, too.

PR #513, opened 2020-07-15
504 rename the pasForm block param to saveableForm

godfreyyeung:
Addresses #504 

- Commit 1: clarifies which variable refers to the form, which to the model.
 It also is consistent with rwcds-form/edit.hbs

- Commit 2: use lowercase form block params to indicate a form hash/object
Change to |form| (lowercase form) to follow community conventions.
This conveys that the variable represents the SaveableForm yielded
hash, instead of a rendered component.

### DIFF
--- a/client/app/components/packages/pas-form/project-area.hbs
+++ b/client/app/components/packages/pas-form/project-area.hbs
@@ -1,5 +1,5 @@
-{{#let @form as |Form|}}
-  <Form.Section @title="Project Area">
+{{#let @form as |form|}}
+  <form.Section @title="Project Area">
     <p>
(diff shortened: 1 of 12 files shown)


client/app/components/packages/pas-form/project-area.hbs
```diff
@@ -9,7 +9,7 @@
         Does the proposed Project, or portion thereof, require a <Ui::ExternalLink @href="https://www1.nyc.gov/site/dep/water/stormwater-management.page">NYC DEP Stormwater Construction Permit</Ui::ExternalLink>?
       </Q.Legend>
 
-      <Form.Field
+      <form.Field
```
allthesignals, line 12: Nice, thank you

Issue #469, opened 2020-07-07
Create project route and modify project model


hannahkates:
Related to #468
[Azure Requirement](#)

### User Story
As an applicant, I want to see an overview of my project so that I can understand its status and see what materials I need to submit.

### [Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP---Applicant-Portal?node-id=0%3A1)

- [ ] Create route `project/[projectid]`
- [x] Modify model: add dcp_publicstatus from dcp_project
- [ ] Load the model into the route

trbmcginnis:
`dcp_publicstatus` already added to model in this PR:

https://github.com/NYCPlanning/labs-applicant-portal/pull/527

hannahkates:
Once this is implemented, the draft PR for #378 will need to be updated to include the real model attributes.

Issue #378, opened 2020-06-01
Create project template


hannahkates:
[Azure Requirement](#)

### User Story
As an applicant, I want to see an overview of my project so that I can understand its status and see what materials I need to submit.

### [Wireframes](https://www.figma.com/file/aYB9O9Vpu31XvHubhLAusX/ZAP---Applicant-Portal?node-id=0%3A1)
![image.png](https://images.zenhubusercontent.com/5ca2425fd9fcf5278d7159e7/25e326c2-8d77-403f-9191-9b00120e09df)

### Display Logic

Sorting:
- Package revisions: Sort package version descending by version number
- Package types: Hardcoced order. PAS always first at bottom

Icons:
- Check mark for most recent package, if submitted
- Pencil for for most recent package, if 'Package Preparation'
- Triangle/exclamation mark warning icon for old package revisions

Copy - make sure to include:
"If something isn’t working, you can’t see your data, or you need assistance, please contact City Planning at ZAP_feedback_DL@planning.nyc.gov or 212-720-3300."

PR #517, opened 2020-07-16
378: project template

hannahkates:
Template for singular project page. 

This PR includes:
- a new `project` route
- a template for a singular project's page
- a component `package-list.hbs` (I think we should rename this to `package-list-item.hbs`) that displays links to different versions of rwcds and pas forms (one is editable and the others are just show)
- a component `date-display` (borrowed from zap-search) that formats dates appropriately in templates
- addition of relevant attributes to project and package models
- update of mirage factories for new relevant attributes

**NOTE TODO:** Versioning is a bit more complicated of a topic to address. So I am going to create another PR directly after this focused on assuring that versioning is working correctly on the singular project page. This will include thorough tests and some organizing of traits/factories that have to do with the project page. Outlined in issue #542 

Addresses #378 and #469


client/app/routes/project.js
```diff
@@ -1,4 +1,17 @@
 import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default class ProjectRoute extends Route {
+export default class ProjectRoute extends Route.extend(AuthenticatedRouteMixin) {
+  authenticationRoute = '/';
+
+  async model(params) {
+    console.log('params', params);
```
godfreyyeung: nit: nix this :)

client/mirage/factories/project.js
```diff
@@ -15,6 +16,15 @@ export default Factory.extend({
     return SAMPLE_NAMES[i % SAMPLE_NAMES.length];
   },
 
+  dcpName(i) {
+    const SAMPLE_CODENAMES = [
+      '2019R0380',
+      '2020K0445',
+      '1990M0111'];
```
godfreyyeung: super nit: closing bracket should be on a newline here.

client/app/routes/project.js
```diff
@@ -1,4 +1,17 @@
...
+    const project = await this.store.findRecord('project', params.id, {
+      reload: true,
+      include: 'packages.pasForm.applicants',
```
godfreyyeung, line 10: We should also load `packages.rwcdsForm`

client/app/routes/project.js
trbmcginnis: this is outdated on an older commit

client/app/components/project/package-list.hbs
```diff
@@ -1,19 +1,17 @@
...
-    <!-- TODO update all ifs with real logic -->
     <a href="#">
-      {{#if (eq status "revision required")}}
+      {{#if (eq @package.statuscode (optionset 'package' 'status' 'code' "Reviewed - Revisions Required"))}}
         <FaIcon @icon="exclamation-triangle" class="text-gold" />
       {{else}}
         <FaIcon
-          @icon={{if (eq status "package prep") 'edit' 'check'}}
-          class={{if (eq status "package prep") '' 'text-green-dark'}}
+          @icon={{if (eq @package.statuscode (optionset 'package' 'status' 'code' "Package Preparation")) 'edit' 'check'}}
```
godfreyyeung: Yay! thanks for using optionset. 
The way you've written here works. As the last parameter, you're passing in the "human readable" label, and that is valid code.

However it's better to use the identifier `PACKAGE_PREPARATION`. This makes it slightly less error prone, and clear that it is a constant. 

e.g. 
```
(optionset 'package' 'status' 'code' 'PACKAGE_PREPARATION')
```

client/app/controllers/project.js
```diff
@@ -0,0 +1,12 @@
...
+import { optionset } from '../helpers/optionset';
+
+export default class ProjectController extends Controller {
+
+  get publicStatusNotNullVisibilityGeneralPublic() {
+    const publicStatus = optionset(['project', 'dcpPublicstatus', 'label', this.model.dcpPublicstatus]);
+    const visibility = optionset(['project', 'dcpVisibility', 'label', this.model.dcpVisibility]);
+    return publicStatus && visibility === 'General Public';
+  }
+}
```
godfreyyeung: nit: newline

client/app/controllers/project.js
```diff
@@ -0,0 +1,12 @@
+import Controller from '@ember/controller';
+import { sort } from '@ember/object/computed';
+import { optionset } from '../helpers/optionset';
+
+export default class ProjectController extends Controller {
+
+  get publicStatusNotNullVisibilityGeneralPublic() {
+    const publicStatus = optionset(['project', 'dcpPublicstatus', 'label', this.model.dcpPublicstatus]);
```
godfreyyeung, line 6: instead of getting the label, you can just directly match against the codes. 
```
const isGeneralPublic = this.model.dcpVisibility === optionset(['project', 'dcpVisibility', 'code', 'GENERAL_PUBLIC']);

return this.model.publicStatus && isGeneralPublic;
```

client/app/controllers/project.js
```diff
@@ -0,0 +1,10 @@
+import Controller from '@ember/controller';
+import { optionset } from '../helpers/optionset';
+
+export default class ProjectController extends Controller {
+  get publicStatusNotNullVisibilityGeneralPublic() {
```
godfreyyeung, line 5: very long name 😅  
Maybe something just like `publicStatusGeneralPublicProjects()` ?
Also including `Projects1 in the name makes clear what this getter returns

client/app/components/project/package-list.hbs
trbmcginnis: good to know, will do!

client/app/router.js
```diff
@@ -22,4 +23,5 @@ Router.map(function() { // eslint-disable-line
     this.route('edit');
     this.route('show', { path: '/' });
   });
+  this.route('project', { path: 'projects/:id' });
```
godfreyyeung, line 26: Nice, thanks!

client/app/templates/project.hbs
```diff
@@ -0,0 +1,61 @@
...
+      <strong>Primary Applicant:</strong> {{@model.dcpApplicantCustomerValue}}
+    </p>
+
+    <p>
+      <strong>Project Status:</strong> {{@model.statuscode}}
+      <span class="text-gray"> | </span>
+      <strong>Public Status:</strong>
+      <FaIcon @icon={{if this.publicStatusNotNullVisibilityGeneralPublic 'eye' 'eye-slash'}} @prefix="far" class="text-gray"/>
+      {{#if this.publicStatusNotNullVisibilityGeneralPublic}}
+        <Ui::ExternalLink @href="https://zap.planning.nyc.gov/projects/{{@model.dcpName}}"> {{optionset 'project' 'dcpPublicstatus' 'label' @model.dcpPublicstatus}} </Ui::ExternalLink>
```
godfreyyeung, line 26: yay, thanks!

client/app/controllers/project.js
trbmcginnis: sounds good -- I'll shorten it :)

allthesignals: Yeah, I've been seeing a little bit of this — the naming of things should abstract away what the code is doing, not really describe the code itself (in camelcase).

allthesignals: Another thing I like to keep in mind when I'm naming stuff (and this comes from Sandi Metz) is that if my method can be named with a conjunction "_and_", it probably needs to be split up.

client/app/helpers/optionset.js
```diff
@@ -51,6 +52,7 @@ const OPTIONSET_LOOKUP = {
   project: {
     dcpPublicstatus: DCPPUBLICSTATUS,
     dcpVisibility: PROJECT_VISIBILITY,
+    status: PROJECT_STATUS,
```
godfreyyeung, line 55: This is my own fault, but I realize I'm re-aliasing some properties. Like package.status and package.type. I was trying to simplify, but I should have just stuck with the full property name to be consistent. Just a note to self, I may refactor this later

server/src/projects/projects.controller.ts
```diff
@@ -31,6 +31,11 @@ import { PROJECT_ATTRS } from './projects.attrs';
       'dcp_packagetype',
       'dcp_visibility',
       'dcp_packageversion',
+      '_dcp_rwcdsform_value',
```
godfreyyeung, line 34: Sorry, one change actually. 
Here I think we can just do 
``1
attributes: [
    ...PACKAGE_ATTRS,
]
```
Just like in  Packages.controller.ts

Good catch here to add the attributes!

PR #518, opened 2020-07-16
379: add definitions of no-action and with-action

hannahkates:
- Adds definitions for no-action and with-action
- Fixes URL in a hyperlink to the CEQR manual


client/app/components/packages/rwcds-form/with-action-no-action.hbs
```diff
@@ -36,7 +36,7 @@
     </h3>
 
     <p>
-      [defnition] [link to CEQR manual]
+      Scenario of the future without the Proposed Project, used as a baseline against which incremental changes generated by a Project are evaluated in Environmental Review. <Ui::ExternalLink @href="https://www1.nyc.gov/site/oec/environmental-quality-review/technical-manual.page">CEQR Technical Manual</Ui::ExternalLink>
```
allthesignals, line 39: We should just have a lint rule for this but I normally like to split this out into multiple lines when possible.

PR #519, opened 2020-07-16
Grabs the full project object and the correct name field

### LINKED ISSUE #493: PAS: The project name field is getting populated with the Project ID, Upon submission, the Project Name changes to the Project ID.
Browser: Chrome,Firefox, IE

The project name field is prepopulated with the Project ID, which seems to serve the same function as the "revised Project name" field from before. Upon submission, the Project Name changes to the Project ID.

image

allthesignals:
Addresses #493


server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -17,11 +17,13 @@ export class PasFormService {
...
         dcp_dcp_applicantrepinformation_dcp_pasform,
         dcp_dcp_projectbbl_dcp_pasform($filter=statecode eq 0),
-        dcp_dcp_projectaddress_dcp_pasform
+        dcp_dcp_projectaddress_dcp_pasform,
+        dcp_projectname($select=dcp_projectname)
     `);
 
-    // CRM feature for providing the human-readable value for the related resource
-    const { '_dcp_projectname_value@OData.Community.Display.V1.FormattedValue': dcp_projectname } = pasForm;
+    // dcp_projectname is the name of the CRM navigation link.
```
godfreyyeung, line 24: did you mean to type another property here?

godfreyyeung: Oh, I see. So pasForm.dcpProjectname is an object with a property of the same name?
```
pasForm = {
    dcp_projectname: {
       dcp_projectname: "My project name", 
    }
}
```

And is the nested `dcp_projectname value the longform value, like "My project name", whereas `_dcp_projectname_value@OData.Community.Display.V1.FormattedValue` is the project id, like `20202MK1`?

allthesignals: Yes exactly. It's a little weird but that's how it's set up in CRM.

Issue #468, opened 2020-07-07
Create new API end point for user's individual project


hannahkates:
Related to #469

[Azure Requirement](#)

### User Story
As an applicant, I want to see an overview of my project so that I can understand its status and see what materials I need to submit.

### Attributes
What attributes should the API response include?

- **source: dcp_project**
  - dcp_projectname
  - dcp_name
  - statuscode
  - dcp_publicstatus
  - dcp_projectbrief
  - dcp_applicant_customer
- **source: dcp_package**
  - dcp_packagetype
  - dcp_packageversion
  - dcp_statusdate
  - id of package
  - package status to determine icons

### Filter Logic

- dcp_project records WHERE **visibility = 'General Public'**
- dcp_package WHERE **statecode = 'Active' and visibility = 'Applicant Only'**
- dcp_projectapplicant WHERE contact is associated with the project ID and 'Active'. otherwise return false. (already implemented on my project end point)

godfreyyeung:
Going to pick this up. @allthesignals This is my plan so far:

- new GET endpoint on the Project controller
- The Project service will use the provided project id to get the specific project
- It will expand packages and applicants on the project

This will be similar to the crmService.get() Web API request in the `findManyByContactId()` function, only we aren't requesting multiple projects by contact id.

This largely follows existing patterns, but let me know if you have additional thoughts!

godfreyyeung:
Oh, also we should put guards in place to prevent those outside the project team from loading the project, if they're not in place

allthesignals:
> Going to pick this up. @allthesignals This is my plan so far:
> 
> * new GET endpoint on the Project controller
> * The Project service will use the provided project id to get the specific project
> * It will expand packages and applicants on the project
> 
> This will be similar to the crmService.get() Web API request in the `findManyByContactId()` function, only we aren't requesting multiple projects by contact id.
> 
> This largely follows existing patterns, but let me know if you have additional thoughts!

I think this makes sense. You might have trouble expanding more than 1 level deep. In that case you can query for the data a little differently — we do it in a few places to save a network request.


> Oh, also we should put guards in place to prevent those outside the project team from loading the project.
> 
> I think we can do this by just including the filter condition that checks that the current contact is within the associated applicant team. The findManyByContactId() function already does this.

That's a good point. I think we should capture this in a separate issue, though, because it starts to get into "authorization" part of our auth strategy. A guard is a great pattern for that.

hannahkates:
[...]

PR #528, opened 2020-07-17
468 Backend projects/id endpoint

godfreyyeung:
Addresses #468 

Sets up an endpoint to GET a project of specific ID.
It will sideload packages and applicant team members.

It only returns the project if the signed in contact
is part of the project's applicant team.

### DIFF
--- a/server/src/projects/projects.attrs.ts
+++ b/server/src/projects/projects.attrs.ts
@@ -5,7 +5,8 @@ export const PROJECT_ATTRS = [
   'statecode',
   'statuscode',
   'dcp_visibility',
+  'dcp_publicstatus',
+  'dcp_projectbrief',
   '_dcp_applicant_customer_value',
   'dcp_dcp_project_dcp_projectapplicant_Project',
-  'dcp_publicstatus',
 ];
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -37,12 +37,52 @@ export class ProjectsService {
             $filter= statuscode eq ${APPLICANT_ACTIVE_STATUS_CODE}
           )
       `);
-   
       return this.overwriteCodesWithLabels(records);
     } catch(e) {
       const errorMessage = `Error finding projects by contact ID. ${e.message}`;
       console.log(errorMessage);
       throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+    }
+  }
+
+  /**
+   * Gets a Project for given dcp_projectid and contactid
+   *
+   * @param      {string}  projectId   CRM Project dcp_projectid
+   * @param      {string}  contactId   CRM Contact contactid
+   * @return     {any}     { ...Project Attributes }
+   */
+  public async getProject(projectId: string, contactId: string) {
+    try {
+      const { records } = await this.crmService.get('dcp_projects', `
+        $filter=
+          dcp_dcp_project_dcp_projectapplicant_Project/
+            any(o:
+              o/_dcp_applicant_customer_value eq '${contactId}'
+              and o/statuscode eq ${APPLICANT_ACTIVE_STATUS_CODE}
+            )
+          and (
+            dcp_visibility eq ${PROJECT_VISIBILITY_APPLICANT_ONLY}
+            or dcp_visibility eq ${PROJECT_VISIBILITY_GENERAL_PUBLIC}
+          )
+          and statecode eq ${PROJECT_ACTIVE_STATE_CODE}
+          and dcp_projectid eq '${projectId}'
+        &$expand=
+          dcp_dcp_project_dcp_package_project,
+          dcp_dcp_project_dcp_projectapplicant_Project(
+            $filter= statuscode eq ${APPLICANT_ACTIVE_STATUS_CODE}
+          )
+      `);
+
+      const [ project ] = this.overwriteCodesWithLabels(records);
+      return project;
+    } catch(e) {
+      const errorMessage = `Error finding project by project ID. ${e.message}`;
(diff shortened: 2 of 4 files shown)


server/src/projects/projects.service.ts
```diff
@@ -37,7 +37,7 @@ export class ProjectsService {
             $filter= statuscode eq ${APPLICANT_ACTIVE_STATUS_CODE}
           )
       `);
-   
+
```
godfreyyeung, line 40: This is just removing whitespace characters

server/src/projects/projects.service.ts
```diff
@@ -46,6 +46,40 @@ export class ProjectsService {
...
+            or dcp_visibility eq ${PROJECT_VISIBILITY_GENERAL_PUBLIC}
+          )
+          and statecode eq ${PROJECT_ACTIVE_STATE_CODE}
+          and dcp_name eq '${projectId}'
+        &$expand=
+          dcp_dcp_project_dcp_package_project,
+          dcp_dcp_project_dcp_projectapplicant_Project(
+            $filter= statuscode eq ${APPLICANT_ACTIVE_STATUS_CODE}
+          )
+      `);
```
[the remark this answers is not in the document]
godfreyyeung: The client only passes `dcp_name` to the backend, though. How would we get `dcp_projectid`, without doing a separate lookup using `dcp_name`? 

I say the client only passes `dcp_name` to the backend assuming a few things:
  1. The client URLs are designed similar to ZAP search, where the URL for an individual project page is something like `/projects/2020MK2`
  2. If a user visits the individual project page directly, the route won't preload all the proejcts. It will only send one request to the backend for the project specified in the URL param. 

I guess we could modify the route to preload all projects (hit the /projects endpoint) first. Thoughts?

server/src/projects/projects.attrs.ts
```diff
@@ -5,6 +5,8 @@ export const PROJECT_ATTRS = [
   'statecode',
   'statuscode',
   'dcp_visibility',
+  'dcp_publicstatus',
```
trbmcginnis, line 8: Adding this in another PR #527 but don't think it should cause much conflict!

server/src/projects/projects.service.ts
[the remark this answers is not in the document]
allthesignals: Why does it only pass `dcp_name` to the backend, though? I'm not sure it should be doing that anyway.

> How would we get `dcp_projectid`, without doing a separate lookup using `dcp_name`?
Not sure I follow — the client just needs to update the `link-to`s to use the correct id.

> 1. The client URLs are designed similar to ZAP search, where the URL for an individual project page is something like `/projects/2020MK2`

Doesn't mean we should be continue doing it that way. If we need this for vanity reasons, we should design it that way intentionally.


> 2\. If a user visits the individual project page directly, the route won't preload all the projects (like by hitting the `/projects` endpoint). It will only send one request to the backend for the project specified in the URL param.

Exactly — that's what projects/id should do. Why does a single project page need all the projects?

Is there a wireframe for this?

This implementation-specific issue style is a little too presumptuous....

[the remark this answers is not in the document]
godfreyyeung: I assumed that planners/applicants would want that predictable human-understandable ulurp number in the URL. But you're right, we don't need to stick with that.

On a separate note, Ryan mentioned we shouldn't be using CRM record ids in the url. Though we can address this in a separate issue.

The single project page imo shouldn't load all projects. I was thinking through a workaround for only passing dcp_name

I'll change it to use projectid in the URL, and let others know of the change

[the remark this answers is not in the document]
allthesignals: > Ryan mentioned we shouldn't be using CRM record ids in the url

Really? Ack. Why? We do that _everywhere_. Lemme know if there's an issue created...

> I assumed that planners/applicants would want that predictable human-understandable ulurp number in the URL. But you're right, we don't need to stick with that ... I'll change it to use projectid in the URL, and let others know of the change

Thanks... I'm pushing for consistency here because it's how our other routes are (mostly) designed. If we need to move away from that, let's make an issue of it and do it. Vanity URLs are a totally valid request, but let's not implement it on accident.

server/src/projects/projects.service.ts
```diff
@@ -46,6 +45,47 @@ export class ProjectsService {
...
+   * @param      {string}  contactId   CRM Contact contactid
+   * @return     {any}     { ...Project Attributes }
+   */
+  public async getProject(projectId: string, contactId: string) {
+    try {
+      const { records } = await this.crmService.get('dcp_projects', `
+        $filter=
+          dcp_dcp_project_dcp_projectapplicant_Project/
+            any(o:
+              o/_dcp_applicant_customer_value eq '${contactId}'
```
allthesignals, line 61: This looks good but I'm still not sure why we check contactId here? Is this an authorization step?

I'm fine with this for now, but we need to come up with a broader authorization strategy. It's weird that projects authorize but not packages, etc.

godfreyyeung: Yep, for auth. Yeah, agreed we need a broader strategy. We had planned that during the next sprint (or sooner) we will make sure packages are locked down. I just wanted to be proactive here about locking it down, since it was an easy step to take here

https://app.zenhub.com/workspaces/labs-sprint-5e56ebd113bedb2575e694da/issues/nycplanning/labs-applicant-portal/507

server/src/projects/projects.attrs.ts
godfreyyeung: yep, frontend and backend! :)

PR #529, opened 2020-07-17
Include default object for destructuring

allthesignals:
Although the default behavior of CRM for 0 related entities is to return a blank array, this isn’t true for a single related entity (a has one). If there is no association, it returns null.

This fix accounts for that.

### DIFF
--- a/server/src/packages/pas-form/pas-form.service.ts
+++ b/server/src/packages/pas-form/pas-form.service.ts
@@ -23,7 +23,7 @@ export class PasFormService {
 
     // dcp_projectname is the name of the CRM navigation link.
     // then dcp_projectname is the real project name
-    const { dcp_projectname } = pasForm.dcp_projectname;
+    const { dcp_projectname } = pasForm.dcp_projectname || {};
 
     return {
       ...pasForm,


server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -23,7 +23,7 @@ export class PasFormService {
 
     // dcp_projectname is the name of the CRM navigation link.
     // then dcp_projectname is the real project name
-    const { dcp_projectname } = pasForm.dcp_projectname;
+    const { dcp_projectname } = pasForm.dcp_projectname || {};
```
godfreyyeung, line 26: Oh huh.. I thought dcp_projectname was just a standard field. I didn't know it was a related entity

allthesignals: @godfreyyeung yeahhh, it's a "navigation link". I don't know why it's called that or behaves this way. I'd like to understand better how these navigation links get created and how they are named. Some or all seem autogenerated, likely through some kind of Access-style model builder UI.

Although this makes me wonder if I can just grab the label metadata... going to check really fast.

allthesignals: Yeah, it's just deceptive: 

```js
  '_dcp_projectname_value@OData.Community.Display.V1.FormattedValue': '2020M0451',
  '_dcp_projectname_value@Microsoft.Dynamics.CRM.associatednavigationproperty': 'dcp_projectname',
  '_dcp_projectname_value@Microsoft.Dynamics.CRM.lookuplogicalname': 'dcp_project',
  dcp_projectname: {
    dcp_projectname: 'Bubba Gump Shrimp Company',
    dcp_projectid: '48e2e539-96a6-ea11-a811-001dd8309f0e'
  }
```

All navigation links get a "FormattedValue", a label representation of the relationship/association, which is inferred from some probably-configured display value, which in the case of dcp_project entity, is probably the ULURP number, _not_ the actual project name.

Still don't know why this navigation link is _called_ "dcp_projectname", but it's really just the dcp_project.

PR #530, opened 2020-07-18
510: Validate has many length

allthesignals:
Draft solution for #510, uses changeset validations.

This does not show an error message yet, but it does validate.

@godfreyyeung , this mostly follows https://davidtang.io/2018/07/28/tracking-has-many-relationship-changes-with-ember-changeset.html

### DIFF
--- a/client/app/utils/ember-changeset.js
+++ b/client/app/utils/ember-changeset.js
@@ -1,0 +1,9 @@
+export function addToHasMany(changeset, relationship, item) {
+  let hasMany = changeset.get(relationship).pushObjects([item]);
+  changeset.set(relationship, hasMany.toArray());
+}
+
+export function removeFromHasMany(changeset, relationship, item) {
+  let hasMany = changeset.get(relationship).removeObject(item);
+  changeset.set(relationship, hasMany.toArray());
+}
--- a/server/src/packages/pas-form/pas-form.service.ts
+++ b/server/src/packages/pas-form/pas-form.service.ts
@@ -23,7 +23,7 @@ export class PasFormService {
 
     // dcp_projectname is the name of the CRM navigation link.
     // then dcp_projectname is the real project name
-    const { dcp_projectname } = pasForm.dcp_projectname;
+    const { dcp_projectname } = pasForm.dcp_projectname || {};
 
     return {
       ...pasForm,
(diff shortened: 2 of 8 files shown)


server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -23,7 +23,7 @@ export class PasFormService {
 
     // dcp_projectname is the name of the CRM navigation link.
     // then dcp_projectname is the real project name
-    const { dcp_projectname } = pasForm.dcp_projectname;
+    const { dcp_projectname } = pasForm.dcp_projectname || {};
```
godfreyyeung, line 26: The commit description says this is to fix something about CRM's return values for related entities. I'm not following.. how does this relate to dcp_projectname?

allthesignals: It doesn't, I just accidentally branched off of that. That's addressed in another PR.

allthesignals: This PR is about validating hasMany length. The diff here is for another PR: https://github.com/NYCPlanning/labs-applicant-portal/pull/529

client/app/utils/ember-changeset.js
```diff
@@ -0,0 +1,9 @@
+export function addToHasMany(changeset, relationship, item) {
+  let hasMany = changeset.get(relationship).pushObjects([item]);
+  changeset.set(relationship, hasMany.toArray());
+}
+
+export function removeFromHasMany(changeset, relationship, item) {
+  let hasMany = changeset.get(relationship).removeObject(item);
+  changeset.set(relationship, hasMany.toArray());
+}
```
godfreyyeung, line 9: This is a cool approach!

allthesignals: Cribbed from here: https://davidtang.io/2018/07/28/tracking-has-many-relationship-changes-with-ember-changeset.html

server/src/packages/pas-form/pas-form.service.ts
godfreyyeung: 👍  Gotcha

godfreyyeung:
Thanks! This is nice and simple -- way more simple than I thought! Let me know if you want me to look into adding the template error messages

godfreyyeung:
Also, there are tests in #510  you can snatch!

allthesignals:
> Also, there are tests in #510 you can snatch!

Perfect! Will grab those. Thank you!

client/app/utils/ember-changeset.js
trbmcginnis: Oh yeah I remember reading about this a while back, looks great!

PR #547, opened 2020-07-23
New "My Projects" layout/UX

andycochran:
This PR… 

- Updates the layout and UX of "My Projects" 
  - No longer shows PAS links
  - Sorts projects into "Projects Awaiting Submission" and "Other Projects"
  - Display all the project list items with the same meta info and layout
- Refactors the components for listing projects on the `projects/` route
  - Combines `<Projects::ApplicantProjectCard>` and `<Projects::PlanningProjectCard>` into the more semantic  `<MyProjects::ListItem>`
  - Makes the vars more semantic:
    - Old: `applicantProjects`, `planningProjects`, `sortedApplicantProjects`, `sortedPlanningProjects`
    - New: `toDoProjects`, `doneProjects`, `sortedToDoProjects`, `sortedDoneProjects`


client/app/templates/projects.hbs
```diff
@@ -6,38 +6,43 @@
...
     {{#if (gt this.model.length 0)}}
-      {{#if (gt this.applicantProjects.length 0)}}
-        <ul class="no-bullet">
-          {{#each this.sortedApplicantProjects as |project|}}
-            <Projects::ApplicantProjectCard @project={{project}} />
+
+      <h2>Projects Awaiting Submission</h2>
+
+      {{#if (gt this.toDoProjects.length 0)}}
+        <ul class="no-bullet" data-test-to-do-projects-list>
```
godfreyyeung, line 13: For the most part we've been naming our test selectors so that the specific variant or modifier on a selector is assigned as the selector value. It's somewhat similar to the CSS concept of [BEM](http://getbem.com/introduction/)

i.e. `data-test-<element>="<modifier>"` 

So in this case, instead of `data-test-to-do-projects-list` and `data-test-done-projects-list`, it should be:
`data-test-projects-list="to-do"` and `data-test-projects-list="done"`

andycochran: Updated.

godfreyyeung: @andycochran  hmm I don't see the changes yet.. maybe it didn't push?

PR #553, opened 2020-07-27
Reset server readme to make it our own

godfreyyeung:
Prepping for a wider project documentation effort

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -1,51 +1,15 @@
-<p align="center">
-  <a href="http://nestjs.com/" target="blank"><img src="https://nestjs.com/img/logo_text.svg" width="320" alt="Nest Logo" /></a>
-</p>
+## Applicant Portal Backend
 
-[travis-image]: https://api.travis-ci.org/nestjs/nest.svg?branch=master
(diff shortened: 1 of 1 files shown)


server/README.md
```diff
@@ -1,51 +1,15 @@
...
-## Description
-
-[Nest](https://github.com/nestjs/nest) framework TypeScript starter repository.
-
-## Installation
+## Setting up
 
 ```bash
-$ npm install
+$ yarn
```
andycochran, line 6: Don't we need to `cd server && yarn` and `cd client && yarn`? Should we make a default script that does this for running a plain `yarn` in the root directory?

andycochran: It's cool if that's part of the wider project documentation effort.

godfreyyeung: Oh yeah, I was thinking that part should come in a top-level README. When in the server folder, you just `yarn`

andycochran: Still might be worth pointing out to `cd server` in this server/README (similar to how most repo README files tell you to `cd [repo-name]`). We shouldn't assume that the user is in the server/ directory because they're reading this file.

Issue #526, opened 2020-07-17
Improve Error logging infrastructure 


godfreyyeung:
Let's improve the robustness of our backend error logging and its passing error messages to the frontend.

Ideas:
  - Standardize format of all our thrown HttpExceptions. 
    - This PR contains a proposed format: https://github.com/NYCPlanning/labs-applicant-portal/pull/92
  - Standardize the codes that we assign to our exceptions. 
      - e.g. `<SHORT_DESCRIPTION>_<FUNCTION>`
      -  e.g.  `GET_PROJECT_SERVICE`

Questions:

 - Will thrown HttpExceptions from multiple layers get logged and passed to the frontend?
    - E.g. Say there is a try/catch in a controller endpoint, and it calls a service that also contains a try/catch. If an error occurs within the service, will the HttpExceptions thrown from both the service and parent controller be visible?
      - Should we only throw HttpExceptions at the "leaf-level" layers like service? Or only throw them at root level layers like the controller?

hannahkates:
Related to #170

hannahkates:
This was 21 story points.

PR #554, opened 2020-07-27
526 Guide for backend error handling

godfreyyeung:
This is a DRAFT (maybe I should have opened this PR as draft, sorry) guide for error logging in the backend. 
Thinking that this doc is something we can react to and drive the conversation about improving and standardizing our error logging.

Please take a look at it in a Markdown viewer, like in VS code (or copy pasting into a github comment box 😛 )

Written to address #526


server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
+# Exception handling in our backend
+
+## Quickstart
+
+To generate errors in the backend that will be passed to the
```
trbmcginnis, line 5: Yay thanks for explaining how to mimic errors! super helpful

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+```json
+{
+  errors: [
+    response: {
+      message: "our error message", 
+      code: OUR_ERROR_CODE, // We also determine this value
+    },
+    status: [401|500|etc...],
+    message: "our error message"
+  ]
```
andycochran, line 149: Could the example strings in this code sample help clarify the difference in the two `message` values?

godfreyyeung: They're actually the same! The value of response.message is copied over to the top-level message property

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+
+```js
+// services/my-project.service.ts
+getProject(){
+  try {
+    return myProject = this.get('dcp_project', ...);
+  } catch (e) {
+    throw new HttpException({
+      code: 'GET_PROJECTS',
+      message: `Failed to find a project for given id. CRM: ${e}`,
```
godfreyyeung, line 39: This should actually be ${e.message}

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+    } catch (e) {
+      throw e; // simply relay the bubbling error
+    }
+  }
+}
+```
+
+Note how it is not necessary to construct a new HttpException in the controller to relay the error from the service. Instead, we simply use `throw` to pass along the error `e` from the service.
+
+Errors will *bubble* through the function call stack. The `catch` block in a try-catch will also catch errors generated from *within* functions that are called in the `try` block.
```
trbmcginnis, line 66: cool :)

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+
+Only the HttpException caught and thrown closest to or on the controller
+endpoint will be returned for a given request. To prevent confusion over which
```
trbmcginnis, line 69: So does this mean that if a controller employs TWO services and both services error out that only one service's error will be shown?

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+backend layer should provide the most detailed error handling, this document prescribes generating new HTTPExceptions in services, and simply relaying them in controllers.
+
+## General Errors / Preventing Gaps
+Since we are trying to `throw new HttpException` at the service level, we may
+want to write a generic, controller-level catch-all for errors. This will help catch
```
trbmcginnis, line 74: okay like an even bigger safety net

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+
+Our nest backend has a custom [Nest exception filter](https://docs.nestjs.com/exception-filters#throwing-standard-exceptions) in place to catch all thrown HttpExceptions and format them before relaying them to the frontend. It will wrap the errors inside an `errors` array in an object:
+
+```json
+  {
+    errors: [ ...thrownHttpExceptions]
+  }
+```
+
+This means in our code we don't need to manually wrap the error. Below is what manually wrapping the error would look like, but our HttpException does this for us:
```
trbmcginnis, line 110: You show an example below of our code if we were handling it manually, can you also show an example of the code where we don't handle it manually (so how we have it now since HttpException handles it for us)? Would be nice to see explicit comparison side by side

server/EXCEPTIONS.md
godfreyyeung: I actually realize I need to rewrite this part because of a nuance I missed. Standby..

godfreyyeung: Have to handle things a little different with async/await. Reading up on it now.

godfreyyeung: So essentially we need to make sure we also use `await` when we directly return async functions. This article was wonderful shedding light on this: https://itnext.io/error-handling-with-async-await-in-js-26c3f20bc06a

I also need to clarify the nuance that that any outer try-catch will only catch _uncaught_  errors. If a nested try-catch wants to relay an error (let an outer try-catch to handle something), it has to re-throw the error or generate a new error.
Updating the README now. I'll also try to update it to address your original question and your other comments. Thanks for these!

godfreyyeung: Re your question:
If a controller employs two services, and both error out, I think only the first one that errors out will be shown. I think code execution stops when the first error is hit

server/EXCEPTIONS.md
godfreyyeung: The contrasting example would actually be the one in the previous section, "General Errors / Preventing Gaps". I'll clarify this!

server/EXCEPTIONS.md
godfreyyeung: > I guess if we want to know more specifics we have to be reallyyy diligent about wrapping most stuff in the services in try-catch blocks
Yes, I think this is the case. But curious about thoughts from @allthesignals 
>  is there another strategy for making sure errors don't seep through services uncaught?
Yep, I think this is an open question!

server/EXCEPTIONS.md
trbmcginnis: okay sweet, thanks!

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,168 @@
...
+```js
+// controllers/my-project.controller.ts
+@Get('/')
+async getProject(@Session() session) {
+  try {
+    return this.myProjectService.getProject();
+  } catch (e) {
+    if (e instanceof HttpException) {
+      throw e;
+    else {
```
godfreyyeung, line 89: need to add a `}` here

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,192 @@
...
+        code: 'APPLICANT_PORTAL_ERROR_CODE',
+        message: 'Applicant Portal could not find coffee. CRM: CRM ran out of tea.',
+      }
+    ```
+    - The `code` property is developer-defined, and should be a unique, concise,
+    descriptive identifier for the type of error. Users can provide this code to
+    help developers locate the code that yielded the error.
+    - The `message` property is a short, human-readable sentence firstly
+    describing the error for the benefit of end users. Secondly, it should contain
+    errors relayed from CRM, if available.
```
allthesignals, line 26: Sorry, just now seeing this, but this isn't an allowed key name in the JSON:API error response: https://jsonapi.org/format/#errors

Please be sure to check existing standards before extending them. If we need our own key names, JSON:API allows for a `meta` property whose value is an object with whatever non-standard key names we need. In this case, I think the spec provides a nice option for this, namely in `detail`.

> detail: a human-readable explanation specific to this occurrence of the problem. Like title, this field’s value can be localized.

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,192 @@
...
+the following try-catch scopes the thrown error to an attempt to find a project.
+
+```ts
+// services/my-project.service.ts
+public async getProject(){
+  try {
+    return myProject = this.get('dcp_project', ...);
+  } catch (e) {
+    throw new HttpException({
+      code: 'GET_PROJECTS',
```
allthesignals, line 38: Not sure I agree with this style...

in HTTP, a status _code_ is something like "404 NOT FOUND" — it tries to explain the status.

In this case, the code describes what named object it comes from, including method. That's useful information, but I think it would go somewhere else, possible in `meta` or just added to the `detail` key

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,192 @@
...
+  - `responseBody` argument is an object with the following schema.
+    ```js
+       const responseBody = {
+        code: 'APPLICANT_PORTAL_ERROR_CODE',
+        message: 'Applicant Portal could not find coffee. CRM: CRM ran out of tea.',
+      }
+    ```
+    - The `code` property is developer-defined, and should be a unique, concise,
+    descriptive identifier for the type of error. Users can provide this code to
+    help developers locate the code that yielded the error.
```
allthesignals, line 23: I think this description conflicts with itself a bit... I don't think it needs to be unique to the `throw new` instance. that might be an `id` in JSON:API errors.

I think the code should attempt to convey what is reasonably inferable from a given point in the code — either through a try/catch block (promise failure or other error), using the the `catch` block's error param _OR_ can be explicitly inferred through logic outside of try/catch, for example:

`if (!contact) throw new HTTPException({ code: 'CONTACT_NOT_FOUND });`

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,192 @@
...
+    throw e; // simply relay the bubbling error
+  }
+}
+```
+
+Note how it is not necessary to construct a new HttpException in the controller to relay the error from the service. Instead, we simply use `throw` to pass along the error `e` from the service.
+
+The `catch` block in a try-catch will catch uncaught errors generated *within* functions that are called in the `try` block. Errors not yet caught will *bubble* through the function call stack until caught.
+
+<u>Ultimately, we are relaying errors up to the Nest global HttpException filter.</u> By re-throwing or generating a new error in catch blocks, we keep the error in an "uncaught" state. The global filter catches any uncaught errors and standardizes how they are wrapped in a request before being sent to the frontend (more in a following section). If in any try-catch blocks caught an error but did nothing in the catch block (relay it or generate a new error), the filter would never detect an uncaught exception.
```
allthesignals, line 72: Interesting! I think I've been doing it wrong then...

allthesignals: Have we confirmed this is true? I have been confused by this a lot

server/EXCEPTIONS.md
godfreyyeung: @allthesignals  the top-level `message` property isn't something I introduced. It's something that already exists, not sure what part of the code is generating it.

allthesignals: Can you point to where it's being generated? Not sure I know where to find that... our Http Exception filter itself just wraps the error in an array.

godfreyyeung: Okay, so I found out that it's generated automatically from the Nest HttpException class, but on the docs it's marked as _private_:

http://nestjs-doc.exceptionfound.com/classes/httpexception.html#error
That'll teach me to only rely on browsing response formats in the dev console

godfreyyeung: @allthesignals There are two catches: 
  1. it seems Nest HTTPExceptions out of the box do not conform to JSONAPI Errors specification. According to the docs, by default when constructing a Nest HTTPException without overriding the response body (simply passing a string instead of an object to the first argument)...
```ts
throw new HttpException('Forbidden', HttpStatus.FORBIDDEN); 
```
then the generated error object looks like this:
```json
{
  "statusCode": 403,
  "message": "Forbidden"
}
```

Note that `message` and `statusCode` were both top-level properties automatically generated by Nest.

Strangely, when I try to verify this, the response looks slightly different in that `statusCode` is instead `status`:
```
     const errorMessage = `Error finding projects by contact ID. ${e.message}`;
     throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
```
... generates the following response:

```json
  {
    error: "Error finding projects by contact ID. Could not find a property named 'dcp_dcp_project_dcp_packae_project' on type 'Microsoft.Dynamics.CRM.dcp_project'.
 at ProjectsService.findManyByContactId (/Users/[...]/projects.service.js:49:19)",
    response: "Error finding projects by contact ID. Could not find a property named \'dcp_dcp_project_dcp_packae_project\' on type \'Microsoft.Dynamics.CRM.dcp_project\'.'",
    status: 401,
    message: "Error finding projects by contact ID. Could not find a property named \'dcp_dcp_project_dcp_packae_project\' on type \'Microsoft.Dynamics.CRM.dcp_project\'.'"
  }
```

Note that both in the docs, and in the live response, the `message` property was provided automatically by Nest. 
The live response does provide the `status` property, which conforms to the JSON API. 

2. Nest's preferred way of editing and adding error properties is to provide an object to the first argument:
```ts
@Get()
async findAll() {
  throw new HttpException({
    status: HttpStatus.FORBIDDEN,
    error: 'This is a custom message',
  }, HttpStatus.FORBIDDEN);
}
```
And the docs suggest this will place those properties in the top-level of the error:
```json
{
  "status": 403,
  "error": "This is a custom message"
}
```

However, in a live test, they are actually placed within a `response` property.
```ts
const errorMessage = `Error finding project by project ID. ${e.message}`;
      console.log(errorMessage);
      throw new HttpException({
        "code": "GET_PROJECT_SERVICE_ERROR",
        "detail": errorMessage,
        "title": "My title",
      }, HttpStatus.UNAUTHORIZED);
```
 yields
```
   {
    Error: Http Exception
      at ProjectsService.getProject (/Users/[...]/projects.service.js:78:19),
     response:
        { 
          code: 'GET_PROJECT_SERVICE_ERROR',
          detail: 'Error finding project by project ID. Could not find a property named \'dcp_dcp_project_dcp_pakage_project\' on type \'Microsoft.Dynamics.CRM.dcp_project\'.',
          title: 'My title'
        },
        status: 401,
        message: 'Http Exception'
   }
```


One thing we could do is modify our filter to mold HTTPException to comply to JSONAPI spec.

Perhaps:
```ts
@Catch(HttpException)
export class HttpExceptionFilter implements ExceptionFilter {
  catch(exception: HttpException, host: ArgumentsHost) {
    ...
    // this shapes the error responses into JSON:API
    // see https://jsonapi.org/format/#errors
    response
      .status(status)
      .json({
        errors: [{
            detail: exception.message,
            code: exception.response.status,
            status: exception.status,
         }],
      });
  }
}
```

server/EXCEPTIONS.md
godfreyyeung: Hmm. The JSON API do suggest the `code` is specific to the application.
```
code: an application-specific error code, expressed as a string value.
```

> code is similar to title in that both identify a general type of problem (unlike detail, which is specific to the particular instance of the problem), but dealing with code is easier programatically,

https://jsonapi.org/examples/#error-objects-error-codes


Their examples also suggest the same:

image

server/EXCEPTIONS.md
allthesignals: Oh! Interesting. I think that's exactly what has to happen...

@godfreyyeung thanks for sleuthing this... I didn't know Nest was opinionated about the shape of the exception. I guess it makes sense given that it's called an _HTTPException_, so some of the language is there.

I think re-mapping is the right way to go... but we might need to figure out how we want to handle each "feature" in the JSON:API error spec — what is the difference between a "title" and a "detail"? Should we use "meta"? Should we use "source"? And so on.

Anyway, great find!

server/EXCEPTIONS.md
allthesignals: Glad we're leaning more on JSON:API. I don't see how my comment conflicts with the language you're quoting. We want  to point to a "general type of problem", not scope the general problem in the code identifier with the named subclass itself.

godfreyyeung: Ohh, I thought you were suggesting the code should be a slightly longer version of the `status` property.

server/EXCEPTIONS.md
godfreyyeung: Yep, I've experimented and verified this

server/EXCEPTIONS.md
godfreyyeung: Ah okay, gotcha, so like what you're saying in your comment below
https://github.com/NYCPlanning/labs-applicant-portal/pull/554/files#r463195525

server/EXCEPTIONS.md
godfreyyeung: Yes, will revise the guide to account for this! 
Thinking about using "status", "detail" and "code" at least. 
Maybe also "meta" to hold controller/service information.

server/EXCEPTIONS.md
```diff
@@ -0,0 +1,185 @@
...
+
+```ts
+// services/my-project.service.ts
+public async getProject(){
+  try {
+    return myProject = this.get('dcp_project', ...);
+  } catch (e) {
+    throw new HttpException({
+      code: 'NO_PROJECT_FOR_ID',
+      message: `Failed to find a project for given id. CRM: ${e.message}`,
```
godfreyyeung, line 42: change this to `detail:`

PR #555, opened 2020-07-27
542: update logic for frontend package versioning

trbmcginnis:
**Problem 1:**
- Previously, `pasPackages` on the project model was only returning the first package in the array (which represented the most recent version). 
- This PR separates out `pasPackages` computed property and `latestPasPackage` computed property so that we have access to all the package versions in an array as well as the most recent one. 
- This was necessary for displaying ALL versions of a package on the project.hbs template.

**Problem 2:**
- the filtering logic for `done` vs `todo` projects was pulled out of the `todo` and `done` computed properties on the projects.js controller and placed in an exported function at the top of the file.
- This not only simplifies the logic but also fixes faulty display logic that was causing projects to end up in both columns rather than just one.
- Checks for rwcds package (instead of just PAS package) were also added in.

### DIFF
--- a/client/app/controllers/projects.js
+++ b/client/app/controllers/projects.js
@@ -2,6 +2,20 @@ import Controller from '@ember/controller';
 import { sort } from '@ember/object/computed';
 import { PACKAGE_STATUS, PACKAGE_VISIBILITY } from '../optionsets/package';
 
+export function packageIsToDo(projectPackages) {
+  return projectPackages.some((projectPackage) => {
+    if (
+      projectPackage.statuscode === PACKAGE_STATUS.PACKAGE_PREPARATION.code
+      && [
+        PACKAGE_VISIBILITY.APPLICANT_ONLY.code,
+        PACKAGE_VISIBILITY.GENERAL_PUBLIC.code,
+      ].includes(projectPackage.dcpVisibility)
+    ) {
+      return true;
+    } return false;
+  });
+}
+
 export default class ProjectsController extends Controller {
   queryParams = ['email'];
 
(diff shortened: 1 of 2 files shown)


client/app/controllers/projects.js
```diff
@@ -31,18 +36,8 @@ export default class ProjectsController extends Controller {
...
-            PACKAGE_VISIBILITY.APPLICANT_ONLY.code,
-            PACKAGE_VISIBILITY.GENERAL_PUBLIC.code,
-          ].includes(projectPackage.dcpVisibility)
-        ) {
-          return false;
-        }
-        return true;
-      });
+      // Check that NONE of the packages are currently editable
+      return !packageIsToDo(project.pasPackages) && !packageIsToDo(project.rwcdsPackages);
```
godfreyyeung, line 40: We can just negate the `toDoProjects` getter to compute this getter. e.g.
```
get doneProjects() {
  return this.model.filter((project) => {
      return !this.toDoProjects.includes(project);
    })
}
```

client/app/controllers/projects.js
```diff
@@ -10,18 +24,9 @@ export default class ProjectsController extends Controller {
...
-          PACKAGE_VISIBILITY.GENERAL_PUBLIC.code,
-        ].includes(projectPackage.dcpVisibility)
-      ) {
-        return true;
-      }
-      return false;
-    }));
+    return this.model.filter((project) =>
+      // Check that at least ONE of the packages is currently editable
+      packageIsToDo(project.pasPackages) || packageIsToDo(project.rwcdsPackages));
```
godfreyyeung, line 29: nice! Can't believe we missed this in the other my-projects PR. good catch!

client/app/models/project.js
```diff
@@ -28,26 +28,28 @@ export default class ProjectModel extends Model {
...
       .reverse();
 
-    if (firstPackage) {
-      return [firstPackage];
-    }
-    return [];
+    return pasPackages;
+  }
+
+  get latestPasPackage() {
```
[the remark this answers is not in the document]
godfreyyeung: Can you just quickly add to the `project.hbs` template a `Project::PackageListItem` for both the `latestPasPackage` and `latestRWCDSPackage`? I think that will help cover these lines of code.

client/app/controllers/projects.js
trbmcginnis: Yeah that works!

client/app/models/project.js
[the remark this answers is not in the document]
trbmcginnis: Sorry not sure what you mean -- we need to be listing all of the possible versions of a package on the project page and if we only pass in `latestRWCDSPackage` we'll only be getting the most recent version. Is this what you were implying?

Maybe a better fix would be to just comment out/delete these computed properties for now since they're not being used anywhere? And then we can bring them back later when we actually employ them?

PR #562, opened 2020-07-29
526 Standardize Projects &Packages error handling

godfreyyeung:
This PR standardizes error handling (the format for how errors are thrown) for just the projects and packages controllers and services, and some CRM service methods. As a whole the error handling redesign is getting bigger and more tedious than I first thought, so wanted to first get another round of reactions just these three areas of code.

This standardization trying to implement the guide in #554

### DIFF
--- a/server/src/projects/projects.controller.ts
+++ b/server/src/projects/projects.controller.ts
@@ -65,12 +65,18 @@ export class ProjectsController {
 
     try {
       if (contactId) {
-        return this.projectsService.findManyByContactId(contactId);
+        return await this.projectsService.findManyByContactId(contactId);
       }
     } catch (e) {
-      const errorMessage = `${e}`;
-
-      throw new HttpException(errorMessage, HttpStatus.BAD_REQUEST);
+      if (e instanceof HttpException) {
+        throw e;
+      } else {
+        throw new HttpException({
+          code: 'FIND_USER_PROJECTS_FAILED',
+          title: 'Failed getting projects',
+          detail: `An unknown server error occured while getting projects. ${e.message}`,
+        }, HttpStatus.INTERNAL_SERVER_ERROR);
+      }
     }
   }
 
@@ -86,15 +92,18 @@ export class ProjectsController {
 
     try {
       if (contactId) {
-        return this.projectsService.getProject(id, contactId);
+        return await this.projectsService.getProject(id, contactId);
       }
     } catch (e) {
-      const errorMessage = `${e}`;
-
-      throw new HttpException({
-        "code": "GET_PROJECT_CONTROLLER_ERROR",
-        "message": errorMessage,
-      }, HttpStatus.BAD_REQUEST);
+      if (e instanceof HttpException) {
+        throw e;
+      } else {
+        throw new HttpException({
+          code: 'FIND_PROJECT_FAILED',
+          title: 'Failed getting a project',
+          detail: `An unknown server error occured while finding a project by ID. ${e.message}`,
+        }, HttpStatus.INTERNAL_SERVER_ERROR);
+      }
     }
   }
 }
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -39,9 +39,11 @@ export class ProjectsService {
       `);
       return this.overwriteCodesWithLabels(records);
     } catch(e) {
-      const errorMessage = `Error finding projects by contact ID. ${e.message}`;
-      console.log(errorMessage);
-      throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+      const errorMessage = `Unable to find projects for current user. ${e.message}`;
+      throw new HttpException({
+        "code": "USER_PROJECTS_NOT_FOUND",
+        "detail": errorMessage,
+      }, HttpStatus.NOT_FOUND);
     }
   }
 
@@ -77,12 +79,13 @@ export class ProjectsService {
       const [ project ] = this.overwriteCodesWithLabels(records);
       return project;
     } catch(e) {
-      const errorMessage = `Error finding project by project ID. ${e.message}`;
+      const errorMessage = `Could not find requested project ${projectId}.`;
       console.log(errorMessage);
       throw new HttpException({
-        "code": "GET_PROJECT_SERVICE_ERROR",
-        "message": errorMessage,
-      }, HttpStatus.UNAUTHORIZED);
+        "code": "PROJECT_NOT_FOUND",
+        "title": "Project not found",
+        "detail": errorMessage,
+      }, HttpStatus.NOT_FOUND);
     }
   }
 
--- a/server/src/crm/crm.service.ts
+++ b/server/src/crm/crm.service.ts
@@ -1,4 +1,8 @@
-import { Injectable } from '@nestjs/common';
+import {
+  HttpException,
+  HttpStatus,
+  Injectable
+} from '@nestjs/common';
 import * as zlib from 'zlib';
 import * as Request from 'request';
 import { ConfigService } from '../config/config.service';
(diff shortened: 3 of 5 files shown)


server/src/projects/projects.controller.ts
```diff
@@ -65,12 +65,17 @@ export class ProjectsController {
 
     try {
       if (contactId) {
-        return this.projectsService.findManyByContactId(contactId);
+        return await this.projectsService.findManyByContactId(contactId);
```
allthesignals, line 68: I don't think you need to `return await` — it's the responsibility of the method consumer to await it's response.

server/src/projects/projects.controller.ts
```diff
@@ -65,12 +65,17 @@ export class ProjectsController {
...
       }
     } catch (e) {
-      const errorMessage = `${e}`;
-
-      throw new HttpException(errorMessage, HttpStatus.BAD_REQUEST);
+      if (e instanceof HttpException) {
+        throw e;
+      } else {
+        throw new HttpException({
+          code: 'PROJECTS_CONTROLLER_UNKNOWN',
```
allthesignals, line 75: I think this could be worded better — not sure the codes should be named controller-specific, more generic. If you want to mention the controller, it might be done in "detail" or something else.

server/src/projects/projects.controller.ts
```diff
@@ -65,12 +65,17 @@ export class ProjectsController {
...
+          message: `An unknown server error occured while getting projects: ${e.message}`,
```
allthesignals, line 76: ... `message` isn't a standard JSON API field: https://jsonapi.org/format/#errors

server/src/projects/projects.controller.ts
```diff
@@ -86,15 +91,17 @@ export class ProjectsController {
 
     try {
       if (contactId) {
-        return this.projectsService.getProject(id, contactId);
+        return await this.projectsService.getProject(id, contactId);
```
allthesignals, line 94: Ditto — why do we need to `return await` ?

server/src/projects/projects.service.ts
```diff
@@ -39,9 +39,12 @@ export class ProjectsService {
       `);
       return this.overwriteCodesWithLabels(records);
     } catch(e) {
-      const errorMessage = `Error finding projects by contact ID. ${e.message}`;
+      const errorMessage = `Error finding projects by contact ID. CRM: ${e.message}`;
```
allthesignals, line 42: The promise above could fail for other reasons — it's not necessarily an error that originates from the CRM system itself.

server/src/projects/projects.service.ts
```diff
@@ -39,9 +39,12 @@ export class ProjectsService {
...
       console.log(errorMessage);
-      throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+      throw new HttpException({
+        "code": "FIND_MANY_BY_CONTACTID_PROJECT_SERVICE",
+        "message": errorMessage,
```
allthesignals, line 46: Yeah, I would say this code shouldn't be specific to the named object itself, but more of a generic description.

Something like `PROJECT_NOT_FOUND`. I would make the code try to describe as much as is reasonably inferable from the promise failure. In some other cases you do an `instance of` which is good. Some even more robust error handling might check the CRM error response body: https://docs.microsoft.com/en-us/dynamics-nav/api-reference/v1.0/dynamics_error_codes

Ditto on `message` — it's not an allowed attribute in JSON:API (unless you want to use `meta`, but `message` seems like what `detail` is trying to achieve.

server/src/projects/projects.service.ts
```diff
@@ -39,9 +39,12 @@ export class ProjectsService {
...
+      }, HttpStatus.UNAUTHORIZED);
```
allthesignals, line 47: I'm not sure we can assume this will always be an `unauthorized` error but I can see how the promise failure in the `try` could lead us to believe that.

If we know _why_ the promise failed, then we could infer that.

_however_ — looking at this, if CRM does not find a contact record, it will silently fail because it will return a blank array of records. Then, this error handling wouldn't even be triggered.

So, I would look into that some more... the actual logic of this throws an error if the promise fails, and assumes that promise failure is always an "unauthorized", even though that might not even be the case _and_ the situation when _no projects_ are found is not handled

server/src/projects/projects.controller.ts
godfreyyeung: The `await` here is necessary to prevent rejected promises (from `findManyByContactId`) from bypassing the `catch` block below. Since `findManyByContactId` is an async function, it'll return an unhandled reject promise, which bypasses the catch block.

I wrote up a section about it in the EXCEPTIONS.md guide
https://github.com/NYCPlanning/labs-applicant-portal/pull/554/files#diff-7ec9a2cb9b2a6a0502d76ee07c26d43cR78

More about it here: https://medium.com/javascript-in-plain-english/handling-exceptions-in-javascript-f14d855be11c


I _was_ wondering if they consumer would expect a promise instead. But it appears the nest guards/intereceptors and ember adapter integrate just fine even with the await....

Not sure if there is a better approach

server/src/projects/projects.controller.ts
godfreyyeung: 👍  gotcha, will rename

server/src/projects/projects.controller.ts
godfreyyeung: see comment above

server/src/projects/projects.service.ts
godfreyyeung: dah, true... hmm.

server/src/projects/projects.service.ts
godfreyyeung: You're right, UNAUTHORIZED may not be the right code here. 

I don't think the a blank array of records should be regarded as an error. In that case, the user's dashboard just shows no projects, as is already the case.

server/src/projects/projects.controller.ts
allthesignals: Aaah. Fascinating! Great find. Thanks for clarifying... I think I've definitely had this problem before and was confounded.

It'd be like:

```javascript
  try {
    const rejectedPromise = this.aPromiseWillreject();
    return rejectedPromise;
  } catch (e) {
    // uncaught!
  }
```

Where try / catch only become promise-aware because of the `await` syntax...

We would have to mix styles to only return:

```javascript
  try {
    return this.aPromiseWillreject
      .catch(e => throw new Error(e));
  } catch (e) {
    // caught! 
  }
```

But that looks bad/is confusing.

> I _was_ wondering if they consumer would expect a promise instead.

Hm, yeah, the method caller will always get a promise return because it's an `async` function, regardless of whether `await` was used. If it wasn't an `async` function, they'd still get a promise because we know the returning method call itself returns a promise. In that case, the caller would need to try/catch it — but then we'd be in a similar spot! Gah!

I'm guessing this is something you've also dug up in practice?

server/src/projects/projects.service.ts
allthesignals: You're right, I was thinking about something else.

I will say though that for routes in which we request a _single_ record by ID, we should use the RESTful feature in Web API: `dcp_projects/:guid`. If it's not found, you'll get a full error message. But it's not super important. Just an ideal.

server/src/projects/projects.controller.ts
godfreyyeung: > We would have to mix styles to only return:
Exactly, yuck 😬 

> Hm, yeah, the method caller will always get a promise return because it's an async function, regardless of whether await was used. If it wasn't an async function, they'd still get a promise because we know the returning method call itself returns a promise. In that case, the caller would need to try/catch it — but then we'd be in a similar spot! Gah!

Oh right, duh me 🤦 , it's still a promise. What do you mean by dug up in practice? 

In our case, I think it's good we catch the rejected promise and re-throw it as an HttpException so that it ultimately goes through our global filter. 

Aside from the examples in the article, I haven't looked into if there are other codebases that also do this specifically, but can look around.

Issue #515, opened 2020-07-16
Combine pas-form-error.hbs, rwcds-form-error.hbs, projects-error.hbs components into one component


trbmcginnis:
These three components are identical so we should create one main component and then reuse.

godfreyyeung:
Closing since the new `<Errors::List />`  component introduced as part of #596  and #598  takes care of much of the redundancy in the error pages.

PR #563, opened 2020-07-29
526 List errors on projects page like for forms

godfreyyeung:
Short and sweet PR to list errors on the Projects page just like how we decided to do so on the PAS and RWCDS form pages, and login page. 

We should also do this for `projects/id`, and work to componentize this code. Perhaps we can do so in #515

### DIFF
--- a/client/app/templates/projects-error.hbs
+++ b/client/app/templates/projects-error.hbs
@@ -6,7 +6,18 @@
   <span>Hmm, something went wrong:</span>
 
   {{#each this.model.errors as |error|}}
-    <span>{{error.message}}</span>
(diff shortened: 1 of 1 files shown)


godfreyyeung:
> Are you not rendering the `ProjectsError` component anywhere?

I believe you're referring to the `templates/project-error.hbs` file? It's actually a template that is automatically displayed whenever Ember detects there's an error from the backend

trbmcginnis:
> > Are you not rendering the `ProjectsError` component anywhere?
> 
> I believe you're referring to the `templates/project-error.hbs` file? It's actually a template that Ember automatically displayed for the projects route whenever Ember detects there's an error from the backend

cool!

Issue #560, opened 2020-07-28
RWCDS: Manually update dcp_includezoningtextamendment


trbmcginnis:
`dcp_includezoningtextamendment` is a field on the rwcds-form that should be true (`717170000`) if there exists a "ZR" `dcpZoningresolutiontype` and that zoning resolution's `dcpZrsectionumber` is "AppendixF". 

Previously this was thought to be autofilled, but it needs to be manually updated (PATCH) on the rwcds-form entity  similar to how we're updating dcp_affectedzoningresolution entity (POST).

PR #564, opened 2020-07-29
560: manually update dcp_includezoningtextamendment in syncActions function

trbmcginnis:
- this PR sets `dcp_includezoningtextamendment` on the rwcds-form to "Yes" (`717170000`) if there exists a "ZR" action and that action's `dcp_ZoningResolution.dcp_zoningresolution` attribute is equal to "AppendixF". 
- This update occurs every time the `syncActions` function is run in `rwcds-form.service.ts`, which means it runs every time the rwcds form page is visited. 
- Right now the way that it's built, it will PATCH this value each time the user visits the rwcds-form page whether the field is already filled or not. I built it out this way based on the assumption that a planner might go into CRM and update fields on the project actions entity after an action is already created. 

Addresses #560

### DIFF
--- a/client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
+++ b/client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
@@ -19,7 +19,7 @@
       {{/if}}
     </Ui::Legend>
 
-    {{#if (or this.hasModifiedZrSectionNumberQuestion this.hasZrSectionTitleQuestion)}}
+    {{#if this.isActionWithAdditionalInformation}}
       {{#if this.hasModifiedZrSectionNumberQuestion}}
         <Ui::Question as |Q|>
           <Q.Label>
@@ -67,9 +67,9 @@
           </zrForm.Field>
         </Ui::Question>
       {{/if}}
-
-      {{#if this.zrAppendixF}}
-        This action {{if (eq
+    
+      {{#if (eq this.zrTypeCode (optionset 'affectedZoningResolution' 'actions' 'code' 'ZR'))}}
+          This action {{if (eq
           @rwcdsForm.data.dcpIncludezoningtextamendment
           (optionset 'rwcdsForm' 'dcpIncludezoningtextamendment' 'code' 'YES')
           ) "includes" "does not include"}} MIH.&nbsp;
--- a/server/src/packages/rwcds-form/rwcds-form.service.ts
+++ b/server/src/packages/rwcds-form/rwcds-form.service.ts
@@ -43,6 +43,28 @@ const ZONING_RESOLUTION_TYPES = [
   { label: "BD", code: 717170038 },
 ];
 
+export const INCLUDE_ZONING_TEXT_AMENDMENT_OPTIONSET = {
+  YES: {
+    code: 717170000,
+    label: 'Yes',
+  },
+  NO: {
+    code: 717170001,
+    label: 'No',
+  },
+  DONT_KNOW: {
+    code: 717170002,
+    label: 'Don\u2019t Know',
+  },
+}
+
+export const ACTION_VALUE_OPTIONSET = {
+  ZR: {
+    code: '886ede3a-dad0-e711-8125-1458d04e2f18',
+    label: 'ZR',
+  }
+}
+
 @Injectable()
 export class RwcdsFormService {
   constructor(
@@ -143,6 +165,8 @@ export class RwcdsFormService {
       &$expand=dcp_ZoningResolution
     `);
 
+    await this.updateIncludeZoningTextAmendment(dcp_rwcdsformid, projectActions);
+
     await Promise.all(projectActions.map(action => {
       const projectActionLabel = action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'];
 
@@ -160,5 +184,26 @@ export class RwcdsFormService {
         });
       }
     }));
+  }
+
+  async updateIncludeZoningTextAmendment(dcp_rwcdsformid, projectActions) {
+    // dcp_includezoningtextamendment is a field on the rwcds-form that should be "Yes" (717170000) if...
+    // there exists a "ZR" action and that action's dcp_ZoningResolution.dcp_zoningresolution is "AppendixF"
+    const includeZoningTextAmendmentAction = projectActions.find(action =>
+      action._dcp_action_value === ACTION_VALUE_OPTIONSET['ZR'].code
+      && action.dcp_ZoningResolution.dcp_zoningresolution === 'AppendixF'
+    );
+
+    if (includeZoningTextAmendmentAction) {
+      await this.crmService.update('dcp_rwcdsforms', dcp_rwcdsformid, {
+        // set to "Yes"
+        dcp_includezoningtextamendment: INCLUDE_ZONING_TEXT_AMENDMENT_OPTIONSET['YES'].code,
+      });
+    } else {
+      await this.crmService.update('dcp_rwcdsforms', dcp_rwcdsformid, {
+        // set to "No"
+        dcp_includezoningtextamendment: INCLUDE_ZONING_TEXT_AMENDMENT_OPTIONSET['NO'].code,
+      });
+    }
   }
 }
--- a/client/app/components/packages/rwcds-form/proposed-project-actions-editor.js
+++ b/client/app/components/packages/rwcds-form/proposed-project-actions-editor.js
@@ -10,10 +10,12 @@ export default class ProposedActionsComponent extends Component {
     'ZA', 'ZS', 'SD', 'SA', 'RA', 'RS',
   ];
 
-  get zrTypeLabel() {
-    const zrTypeCode = this.args.zrForm.data.dcpZoningresolutiontype;
+  get zrTypeCode() {
+    return this.args.zrForm.data.dcpZoningresolutiontype;
+  }
 
-    return optionset(['affectedZoningResolution', 'actions', 'label', zrTypeCode]);
+  get zrTypeLabel() {
+    return optionset(['affectedZoningResolution', 'actions', 'label', this.zrTypeCode]);
   }
 
   get zrAppendixF() {
@@ -38,5 +40,11 @@ export default class ProposedActionsComponent extends Component {
   get hasZrSectionTitleQuestion() {
     const zrTypeInArray = this.actionsWithSectionNumberAndSectionTitle.includes(this.zrTypeLabel);
     return zrTypeInArray || this.zrAppendixF;
+  }
+
+  get isActionWithAdditionalInformation() {
+    return this.hasModifiedZrSectionNumberQuestion
+    || this.hasZrSectionTitleQuestion
+    || this.zrTypeCode === optionset(['affectedZoningResolution', 'actions', 'code', 'ZR']);
   }
 }


server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -143,6 +143,27 @@ export class RwcdsFormService {
       &$expand=dcp_ZoningResolution
     `);
 
+    // dcp_includezoningtextamendment is a field on the rwcds-form that should be "Yes" (717170000) if...
+    // there exists a "ZR" action and that action's dcp_ZoningResolution.dcp_zoningresolution is "AppendixF"
+    const zrAppendixF = projectActions.filter(action =>
```
godfreyyeung, line 148: is `zrAppendixF` at most only one element?

godfreyyeung: If so, I think we can use something like `projectActions.find()`, or maybe (less appropriate) `projectActions.some()` 

I got a bit confused at first because the `filter()` function makes me expect `zrAppendixF` to be an array of multiple elements.

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -143,6 +143,27 @@ export class RwcdsFormService {
...
+      action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'] === 'ZR'
+      && action.dcp_ZoningResolution.dcp_zoningresolution === 'AppendixF'
+    );
+
+    const includeZoningTextAmendment = zrAppendixF.length > 0;
+
+    if (includeZoningTextAmendment) {
+      await this.crmService.update('dcp_rwcdsforms', dcp_rwcdsformid, {
+        // set to "Yes"
+        dcp_includezoningtextamendment: 717170000,
```
godfreyyeung, line 158: I know it seems tedious, but can we create an optionset for this here?
Like how we have the `PACKAGE_TYPE_OPTIONSET` in package.server.ts?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -143,6 +143,27 @@ export class RwcdsFormService {
       &$expand=dcp_ZoningResolution
     `);
 
+    // dcp_includezoningtextamendment is a field on the rwcds-form that should be "Yes" (717170000) if...
+    // there exists a "ZR" action and that action's dcp_ZoningResolution.dcp_zoningresolution is "AppendixF"
+    const zrAppendixF = projectActions.filter(action =>
+      action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'] === 'ZR'
+      && action.dcp_ZoningResolution.dcp_zoningresolution === 'AppendixF'
```
godfreyyeung, line 150: not ideal to have static values here, but it does seem this is super one-off and unlikely to surface elsewhere (I hope). Otherwise I would also vouch for transforming `ZONING_RESOLUTION_TYPES` into an optionset and using it here for `ZR`. I don't know if it would be worth it to create one for AppendixF.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -67,14 +67,13 @@
           </zrForm.Field>
         </Ui::Question>
       {{/if}}
-
-      {{#if this.zrAppendixF}}
+    
+    {{else if (eq this.zrTypeLabel 'ZR')}}
```
godfreyyeung, line 71: I guess this is okay since we rely so heavily on the action type labels anyway in `lookup-action-type` helper.  But in general I think it would be less brittle if throughout the app we used codes as much as possible for business logic and comaprisons, instead of text labels.

trbmcginnis: you mean use the lookup-action-type helper in the template here?

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: okay good point, I can update

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: Can you explain how using an optionset be any different than using the static value? We'd still have to include 'ZR' somewhere to denote that that's the action we want, whether we're using an optionset or not.

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: For sure!

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: In my mind it seems more proper to use codes like 71710000 to do business logic comparisons, since that's the language in which computers and CRM speaks. In this specific case I guess it should be fine since action codes are unique and not very messy ('ZR' vs 'ZA' vs 'UK', etc), but sometimes when using human readable labels it gets messy. E.g. maybe it would be 'ZR - Zoning Resolution` or 'ZA - Zoning Action' where there's more potential for whitespace, capitalization or pluralization errors. Also coding static values means those errors would be silent. If someone typed 'ZRS' instead of 'ZR', it would fail silently.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
godfreyyeung: Mm, I mean just compare against `this.args.zrForm.data.dcpZoningresolutiontype`. Or maybe introduce a getter to alias that property, called `zrTypeCode`, and compare against that instead. 
`(eq this.zrTypeCode (optionset 'affectedZoningResolution' 'actions' 'code' 'ZR'))`

trbmcginnis: So we're using the optionset to determine `zrTypeLabel` in the javascript for this component. 

What I'm gathering is that you're saying that we should avoid using optionsets in computed properties and then employing them in templates? And instead we should always just try to use the optionset directly in the template instead? 

I can try to update other places if you think that's best, but could you also explain why using directly in the template is better

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: I guess my confusion comes from that 'ZR' is how the data is actually stored in CRM. `action['_dcp_action_value@OData.Community.Display.V1.FormattedValue']` is stored as text strings, e.g. `ZR`, `ZC`, `ZA` -- so why would I use a code?

allthesignals: Using `optionset` centralizes the use of static lookup values like this. This sort of thing should at least be kept in a named lookup so we avoid littering our code with lookup strings (that might need to change later).

I could see a nice lookup table, even at the top of the file (if not defined and then exported in the optionset helper) looking like this:
```javascript
const ACTION_LOOKUPS = {
  ZONING_RESOLUTION: 'ZR',
  // others ...
};
```

Two reasons for this:

1. clearer, self-documenting code
2. if values changes for some reason, we go to one place
3. if values change for some reason, we don't have to edit those values in multiple places

allthesignals: I mean you do the same thing with `INCLUDE_ZONING_TEXT_AMENDMENT_OPTIONSET` which is awesome. Why not keep doing it for these other things? The way you handle `INCLUDE_ZONING_TEXT_AMENDMENT_OPTIONSET` is perfect.

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
```diff
@@ -19,7 +19,7 @@
       {{/if}}
     </Ui::Legend>
 
-    {{#if (or this.hasModifiedZrSectionNumberQuestion this.hasZrSectionTitleQuestion)}}
+    {{#if (or this.hasModifiedZrSectionNumberQuestion this.hasZrSectionTitleQuestion (eq this.zrTypeLabel 'ZR'))}}
```
allthesignals, line 22: Could this be made into a nicely-named computed property in the component class? That way we know what the logic here is trying to achieve in friendlier terms.

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -43,6 +43,21 @@ const ZONING_RESOLUTION_TYPES = [
...
+  },
+  NO: {
+    code: 717170001,
+    label: 'No',
+  },
+  DONT_KNOW: {
+    code: 717170002,
+    label: 'Don\u2019t Know',
+  },
+}
```
allthesignals, line 59: Great great great

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: Okay I'll turn ZONING_RESOLUTION_TYPES lookup into optionset and then just use it as `ZONING_RESOLUTION_TYPE['ZONING_TEXT_AMENDMENT'].label` (which would equal "ZR")

godfreyyeung: > I guess my confusion comes from the fact that 'ZR' is how the data is actually stored in CRM. `action['_dcp_action_value@OData.Community.Display.V1.FormattedValue']` is stored as text strings, e.g. `ZR`, `ZC`, `ZA` -- so why would I use a code?

@trbmcginnis  Oh, when I saw the optionset I assumed in CRM it is stored as codes:
```
![image](https://user-images.githubusercontent.com/3311663/88964860-d3c6c780-d25e-11ea-9bd7-dfbe4b5d1bc1.png)
```
I think maybe when you specifically access the `...Display.V1.FormattedValue` version of the property, CRM nicely provides the humna-readable label version for you, but it's not the true underlying value

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
godfreyyeung: Think of there being two different languages: machine codes (717000) and human-speak labels ("ZR").
My recommendation is work with machine codes as much as possible throughout the code, especially for any business logic like comparisons (regardless of in template or a computed). We would only use the human-speak labels when printing them for **display** in the templates.

server/src/packages/rwcds-form/rwcds-form.service.ts
allthesignals: Ah, yeah — it's tricky. Basically this is a way to get around having to do another expand. CRM Web API provides a general "formatted value" for a related entity. In this case, the related entity is `dcp_action`, and named formatted value is derived from the action type value.

There are other places we do that in the code.

trbmcginnis: So that lookup is for the dcp_affectedzoningresolution entity, the 'ZR' that I'm using is for the dcp_projectaction entity. Basically the same attribute but on two different entities and with two different entity names.

That's a good point about the formatted value version of the attribute, I didn't notice that. Going to try to look into that and see if it uses codes similar to dcp_affectedzoningresolution

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
trbmcginnis: Okay so what I'm taking from this is that using optionsets in computed properties and then employing those in templates is a no-go in the future. And that if we can, we should always be using optionsets DIRECTLY in the template when possible

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -143,6 +158,25 @@ export class RwcdsFormService {
       &$expand=dcp_ZoningResolution
     `);
 
+    // dcp_includezoningtextamendment is a field on the rwcds-form that should be "Yes" (717170000) if...
+    // there exists a "ZR" action and that action's dcp_ZoningResolution.dcp_zoningresolution is "AppendixF"
+    const includeZoningTextAmendmentAction = projectActions.find(action =>
+      action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'] === 'ZR'
+      && action.dcp_ZoningResolution.dcp_zoningresolution === 'AppendixF'
```
allthesignals, line 165: I'm sorry but could we move this into another method?

It could be a private method and `syncActions` calls it.

Splitting things out into new methods gives you a way to name that chunk of logic in a way that makes it clearer (and self-documents). It makes it easier to debug. It also keeps the method shorter.

Please think about the method as a whole and what it's doing — is the method concise? Does it do too much? Should we keep adding on to the amount it's doing? Etc.

trbmcginnis: Yep, I'll move

client/app/components/packages/rwcds-form/proposed-project-actions-editor.hbs
trbmcginnis: Sounds good

PR #570, opened 2020-07-30
Kill all cookies

allthesignals:
This PR addresses NYCPlanning/labs-zap-search#1211 (wrong repo?) by using token-based auth, eliminating the need for cross-site cookies (and the sameSite=none attribute)

### DIFF
--- a/client/app/authenticators/zap-api-authenticator.js
+++ b/client/app/authenticators/zap-api-authenticator.js
@@ -1,69 +1,33 @@
-import BaseAuthenticator from 'ember-simple-auth/authenticators/base';
 import fetch from 'fetch';
-import window from 'ember-window-mock';
-import { inject as service } from '@ember/service';
+import OAuth2ImplicitGrantAuthenticator from 'ember-simple-auth/authenticators/oauth2-implicit-grant';
 import { InvalidError } from '@ember-data/adapter/error';
 import ENV from '../config/environment';
 
-// lifted from https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/oauth2-implicit-grant-callback-route-mixin.js#L6
-// parses a window hash and grabs the access token
-function _parseResponse(locationHash) {
-  const params = {};
-  const query = locationHash.substring(locationHash.indexOf('?'));
-  const regex = /([^#?&=]+)=([^&]*)/g;
-  let match;
+export default class ZAPAuthenticator extends OAuth2ImplicitGrantAuthenticator {
+  async authenticate(...args) {
+    let accessToken;
 
-  // decode all parameter pairs
-  while ((match = regex.exec(query)) !== null) { // eslint-disable-line
-    params[decodeURIComponent(match[1])] = decodeURIComponent(match[2]);
-  }
+    try {
+      const { access_token } = await super.authenticate(...args);
 
-  return params;
-}
-
-export default class ZAPAuthenticator extends BaseAuthenticator {
-  @service
-  store
-
-  // Authentication is implicit with an httponly cookie holding
-  // a ZAP Token. This request will fail if that cookie doesn't exist
-  // https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md#using-a-dedicated-endpoint
-  async _fetchContact() {
-    const { id, emailaddress1, contactid } = await this.store.queryRecord('contact', { me: true });
-
-    return { id, emailaddress1, contactid };
-  }
-
-  async authenticate() {
-    // Extract NYCID Token from URL
-    const { access_token } = _parseResponse(window.location.hash);
-
-    if (!access_token) {
-      throw new InvalidError([{
-        response: {
-          code: 'NO_TOKEN_PRESENT',
-        },
-      }]);
+      accessToken = access_token;
+    } catch (e) {
+      throw new InvalidError([{ detail: e.toString() }]);
     }
 
     // Pass the NYCIDToken to backend /login endpoint.
-    // The backend's response will contain a Cookie with a ZAP token.
-    // This Cookie is automatically stored away and used in future requests to the
-    // backend.
-    const response = await fetch(`${ENV.host}/login?accessToken=${access_token}`, {
-      mode: 'same-origin',
-      credentials: 'include',
-    });
+    // Server should provide an access_token used for signing
+    // requests using the Authorization header
+    const response = await fetch(`${ENV.host}/login?accessToken=${accessToken}`);
+    const body = await response.json();
 
-    if (!response.ok) throw await response.json();
+    if (!response.ok) throw body;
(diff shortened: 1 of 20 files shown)


server/src/app.controller.ts
```diff
@@ -7,20 +7,27 @@ import {
...
     private readonly authService: AuthService,
+    private readonly contactService: ContactService,
   ) {}
 
   @Get('/login')
   async login(@Res() res: Response, @Query('accessToken') NYCIDToken: string) {
     try {
       const ZAPToken = await this.authService.generateNewToken(NYCIDToken);
+      const { contactId } = await this.authService.validateCurrentToken(ZAPToken);
+      const { contactid, emailaddress1 } = await this.contactService.findOneById(contactId);
```
allthesignals, line 24: Ack

godfreyyeung:
> This PR addresses [NYCPlanning/labs-zap-search#1211](https://github.com/NYCPlanning/labs-zap-search/issues/1211) (wrong repo?) by using token-based auth, eliminating the need for cross-site cookies (and the sameSite=none attribute)

Can you elaborate a bit more, and give an overview of the new token-based auth solution and how it works?

allthesignals:
> > This PR addresses [NYCPlanning/labs-zap-search#1211](https://github.com/NYCPlanning/labs-zap-search/issues/1211) (wrong repo?) by using token-based auth, eliminating the need for cross-site cookies (and the sameSite=none attribute)
> 
> Can you elaborate a bit more, and give an overview of the new token-based auth solution and how it works?

Originally, our auth worked like this:

Client app sends the NYC.ID accessToken to our server. The server issues an http-only cookie. For subsequent API calls, the client includes the cookie, which authenticates the request. The cookie stores the token, which the backend decrypts for contactid information.

But, because our client and server live at separate domains, our app begins to get into cross-site cookies territory. That complicates things because of new security restrictions that Google — and some corporate security settings block. 

This PR changes the authentication strategy by removing server-issued cookies entirely and relying on the encryption technology provided by JWT.

Instead of shipping a cookie with a request, it simply stores the JWT (that comes back as JSON value from the API) in every HTTP request header, specifically the `Authentication` header.

This PR leverages ember-simple-auth to handle "remember me" behavior, which, under the hood, stores the JWT in local storage (or browser cookies, as a fallback) so it can be used in future sessions.

godfreyyeung:
> > > This PR addresses [NYCPlanning/labs-zap-search#1211](https://github.com/NYCPlanning/labs-zap-search/issues/1211) (wrong repo?) by using token-based auth, eliminating the need for cross-site cookies (and the sameSite=none attribute)
> > 
> > 
> > Can you elaborate a bit more, and give an overview of the new token-based auth solution and how it works?
> 
> Originally, our auth worked like this:
> 
> Client app sends the NYC.ID accessToken to our server. The server issues an http-only cookie. For subsequent API calls, the client includes the cookie, which authenticates the request. The cookie stores the token, which the backend decrypts for contactid information.
> 
> But, because our client and server live at separate domains, our app begins to get into cross-site cookies territory. That complicates things because of new security restrictions that Google — and some corporate security settings block.
> 
> This PR changes the authentication strategy by removing server-issued cookies entirely and relying on the encryption technology provided by JWT.
> 
> Instead of shipping a cookie with a request, it simply stores the JWT (that comes back as JSON value from the API) in every HTTP request header, specifically the `Authentication` header.
> 
> This PR leverages ember-simple-auth to handle "remember me" behavior, which, under the hood, stores the JWT in local storage (or browser cookies, as a fallback) so it can be used in future sessions.

than you!

client/app/authenticators/zap-api-authenticator.js
```diff
@@ -1,79 +1,32 @@
...
 
-    if (!response.ok) throw await response.json();
+    if (!response.ok) throw body;
 
-    // Since all requests to the API are now authenticated,
-    // formally request a User through Ember Data and return it.
-    // Returning the user data here places it inside
-    // the session data's "authenticated" property and marks
+    // Since all requests to the API are now authenticated.
+    // Returning the session data's "authenticated" property and marks
```
godfreyyeung, line 27: I don't understand this comment here. How are you returning the session data's authenticated property here?

allthesignals: ah yeah i'm referring to this: 
image

http://ember-simple-auth.com/api/classes/SessionService.html#method_authenticate

godfreyyeung: Oh I see, here it's returning _to_ the session data's authenticated property

client/app/controllers/login-error.js
```diff
@@ -6,6 +6,6 @@ export default class LoginErrorController extends Controller {
     *                     about Contact not existing in CRM.
     */
   get contactNotAssigned() {
-    return this.model.errors.some((error) => error.response.code === 'NO_CONTACT_FOUND');
```
godfreyyeung, line 9: Why the move away from `error.response`?
The login error page is now reporting [Object object] once more:
image

client/app/templates/login-error.hbs
```diff
@@ -20,7 +20,7 @@
           {{#if error.detail}}
             <code class="callout display-block text-left red-muted" data-test-error-message={{idx}}>{{error.detail}}</code>
           {{else}}
-            {{#each-in error.response as |key value|}}
```
godfreyyeung, line 23: same comment as above

client/app/controllers/login-error.js
allthesignals: Again, because the `response` property is not part of the JSON:API spec. It doesn't belong in there.

server/src/app.controller.ts
```diff
@@ -7,20 +7,27 @@ import {
...
 
-      res.cookie('token', ZAPToken, { httpOnly: true, sameSite: 'none', secure: true })
-        .send({ message: 'Login successful!' });
+      res.send({
+        access_token: ZAPToken,
+        contactid,
```
godfreyyeung, line 28: Interestingly enough, I don't think the frontend actually needs the contactid 😂  It's not used in the frontend, and the backend looks up the contactid each time using the auth token in the request header.

godfreyyeung: Well, I guess the contactid is already embedded within the token

allthesignals: Oooh. Good catch. I'll clean that up. IDK why I included it anyway.

client/app/controllers/login-error.js
godfreyyeung: Okay, you're right. We should remove the logging of `error.response` in all our templates. I (or one of us) will do the housekeeping in future PRs.

allthesignals: I can do it. I just didn't realize there were other places

PR #591, opened 2020-08-04
Extract user badge to new component; add edit link

allthesignals:
According to NYC.ID terms of use, we must link out to the user's profile on the NYC.ID website. This allows the user to edige their profile: https://www1.nyc.gov/assets/nyc4d/html/services-nycid/account-profile.shtml.

This change takes some of the existing markup for displaying the user email and "log out" link, extracts it into its own component, and hyperlinks that email out to the required NYC.Id profile edit page.

The complexity in here is due to the requirement of NYC.ID to encode the redirect (which occurs afte ruser "saves" their changes to their profile page) uri back to the app. That must be encoded as base64.

### DIFF
--- a/client/app/components/auth/user-badge.hbs
+++ b/client/app/components/auth/user-badge.hbs
@@ -1,0 +1,16 @@
+<FaIcon @icon="user-circle" @prefix="fas" @size="2x" class="auth--icon" />
+<div>
+  <span class="auth--name">
+    <a href={{this.userProfileURI}} rel="noopener noreferrer">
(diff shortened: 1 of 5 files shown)


allthesignals:
> If this PR doesn't have an issue that it links out to, would you mind explaining the problem you're trying to fix in the PR description?

@trbmcginnis Yes, my bad. I'll do better. Updated desc

allthesignals:
@andycochran  need inp ut

allthesignals:
@andycochran going to merge but let's revisit markup in a later PR....

PR #594, opened 2020-08-05
526 Flatten nested response objects and mold to JSON API format 

godfreyyeung:
Rework the HTTP Exception Filter to mold Exceptions to JSON API
compliant error objects before returning them to the client.

Also, response objects from HttpExceptions may be recursively
nested to reflect a stack trace (each level is a layer in the stack).
This flattens them out so that each response
is a top-level item in the errors array.

### DIFF
--- a/server/src/http-exception.filter.ts
+++ b/server/src/http-exception.filter.ts
@@ -1,5 +1,27 @@
 import { ExceptionFilter, Catch, ArgumentsHost, HttpException } from '@nestjs/common';
-import { Request, Response } from 'express';
+import { Response } from 'express';
+
+// response objects from HttpExceptions may be recursively
+// nested to reflect a stack trace (each level is a layer
+// in the stack). This flattens them out so that each response
+// is a top-level item in the errors array.
+export function unfoldedStackTrace(response, status) {
+  const firstError = {
+    code:   response.code,
+    title:  response.title,
+    detail: response.detail,
+    status,
+  };
+
+  if (!response.response) {
+    return [firstError];
+  }
+
+  return [
+    firstError,
+    ...unfoldedStackTrace(response.response, status)
+  ];
+}
 
 // We implement this to reshape the way http exception errors are 
 // provided back to the client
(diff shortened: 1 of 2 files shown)


server/src/http-exception.filter.ts
```diff
@@ -1,6 +1,25 @@
...
 import { Request, Response } from 'express';
 
+
+function unfoldedStackTrace(response, status) {
+  if (!response.response) {
+    return [{
+      code:   response.code,
+      title:  response.title,
+      detail: response.detail,
+      status,
```
allthesignals, line 11: Maybe we could turn this into an object:

```
const errorObject = {
      code:   response.code,
      title:  response.title,
      detail: response.detail,
      status,
};
```

Just so that the shape of the error object is in one spot.

server/src/http-exception.filter.ts
```diff
@@ -1,6 +1,25 @@
 import { ExceptionFilter, Catch, ArgumentsHost, HttpException } from '@nestjs/common';
 import { Request, Response } from 'express';
 
+
+function unfoldedStackTrace(response, status) {
```
allthesignals, line 5: `export function unfoldedStackTrace` and unit test in the http-exception.filter test or somewhere

allthesignals: Also, this is a really elegant and clear use of recursion. Really awesome job. Thank u for solving this!

server/src/http-exception.filter.ts
godfreyyeung: way ahead of me! yep!

server/src/http-exception.filter.ts
godfreyyeung: Thank you Matt!!

server/src/http-exception.filter.ts
```diff
@@ -1,5 +1,27 @@
 import { ExceptionFilter, Catch, ArgumentsHost, HttpException } from '@nestjs/common';
-import { Request, Response } from 'express';
+import { Response } from 'express';
+
+// response objects from HttpExceptions may be recursively
+// nested to reflect a stack trace (each level is a layer
+// in the stack). This flattens them out so that each response
+// is a top-level item in the errors array.
+export function unfoldedStackTrace(response, status) {
+  const firstError = {
```
allthesignals, line 9: Teensy nit, but I don't know if "firstError" was the right name — when it recurses, it'll be the second, third, however deep the nesting goes.

Issue #565, opened 2020-07-30
Update the copying over of dcp_projectactions to dcp_affectedzoningresolution to incorporate possible edits to actions


trbmcginnis:
- Originally when a user would visit the rwcds-form page, data from the `dcp_projectaction` entity would be POSTed to the `dcp_affectedzoningresolution` entity
- This would only ever happen one time. Then for future visits to the rwcds-form page, we would check whether the zoning resolution type existed, and if it did exist we would skip POSTing it to `dcp_affectedzoningresolution`
- we recently learned that planners can go into CRM and update attributes stored on dcp_projectaction, which would mean that dcp_affectedzoningresolution would be out-of-date. 

We need to update this so that if a planner does change the attributes of a dcp_projectaction in CRM, dcp_affectedzoningresolution will also be updated.

trbmcginnis:
**When will planners update actions?** It's always possible to for a planner to update  an action, they can do so any time throughout the process, but DCP usually tries to get them to finalize early. After an applicant submits the RWCDS, there might need to be a couple tweaks -- a new method of doing this is to reactive the package (with the same ID as before) so that a user can edit. Reactivating a package will not trigger the POSTing to dcp_affectedzoningresolution that's in place now, because the rwcds-form visited will not have a new ID. 

Although a planner CAN update the actions AFTER an rwcds form has been submitted, we don't really need the data in the dcp_affectedzoningresolution to be updated after that. The planner will usually prompt the user to either fill out a version of the rwcds-form OR will reactivate the latest form for the user. 

So the POSTing to dcp_affectedzonginresolution will change in these ways:

- The POST will still be triggered by visiting the rwcds-form page. BUT we will not only be checking for dcp_zoningresolutiontype as the attribute that denotes if an action has already been copied over. INSTEAD we will check whether EVERY attribute is the same/different and based on that, we will either POST or pass over that specific action.

PR #595, opened 2020-08-05
565: update syncActions to PATCH updates to dcp_affectedzoningresolution

trbmcginnis:
For cases where the specific dcp_projectaction row has already been POSTed to dcp_affectedzoningresolution entity...if a planner updates dcp_projectaction entity in CRM afterwards, then syncActions function will check for these changes and PATCH the changes to dcp_affectedzoningresolution. 

addresses #565

### DIFF
--- a/server/src/packages/rwcds-form/rwcds-form.service.ts
+++ b/server/src/packages/rwcds-form/rwcds-form.service.ts
@@ -152,7 +152,12 @@ export class RwcdsFormService {
   // The syncActions method queries for project actions (dcp_projectaction) associated
   // with the rwcds form's related project. 
   // Then, it posts each of the project actions to dcp_affectedzoningresolution.
-  // This occurs every time that the rwcds-form endpoint is hit
+  // This occurs every time that the rwcds-form endpoint is hit.
+
+  // dcp_projectaction entity --> dcp_affectedzoningresolution entity attribute mapping:
+  // `_dcp_action_value@OData.Community.Display.V1.FormattedValue`  --> `dcp_zoningresolutiontype`
+  // `ZoningResolution.dcp_zoningresolution` --> `dcp_zrsectionnumeber`
+  // `dcp_zrmodifyingzrtxt` --> `dcp_modifiedzrsectionnumber`
   async syncActions(rwcdsForm) {
     const {
       dcp_rwcdsform_dcp_affectedzoningresolution_rwcdsform: affectedZoningResolutions,
@@ -171,14 +176,30 @@ export class RwcdsFormService {
       const projectActionLabel = action['_dcp_action_value@OData.Community.Display.V1.FormattedValue'];
 
       // Lookup: dcp_affectedzoningresolutions asks for a ZR type,
-      const { label, code } = ZONING_RESOLUTION_TYPES.filter((zr) => zr.label === projectActionLabel)[0] || { label: '', code: null };
+      const { label, code } = ZONING_RESOLUTION_TYPES.find((zr) => zr.label === projectActionLabel) || { label: '', code: null };
+      const matchingZr = affectedZoningResolutions.find(zr => zr.dcp_zoningresolutiontype === code);
+      // action attribute that matches dcp_zrsectionnumber on dcp_affectedzoningresolution
+      const currentActionZoningResolution = action.dcp_ZoningResolution ? action.dcp_ZoningResolution.dcp_zoningresolution : null
 
       if (!label) console.log(`Could not find Affected ZR Type for ${projectActionLabel}`);
 
-      if (!zrTypes.includes(label) && label) {
+      // IF the action.ZoningResolution.dcp_zoningresolution value has changed since the initial POST,
+      // then PATCH the new value to dcp_zrsectionnumber on dcp_affectedzoningresolution entity
+      // NOTE: dcp_modifiedzrsectionnumber not included here because it's updated by applicant on rwcds-form
+      if (matchingZr && matchingZr.dcp_zrsectionnumber !== currentActionZoningResolution) {
+        return this.crmService.update('dcp_affectedzoningresolutions', matchingZr.dcp_affectedzoningresolutionid, {
+          dcp_zrsectionnumber: currentActionZoningResolution,
+        });
+      }
+
+      // IF the zr type does not exist yet, then POST the action
+      // NOTE: some actions that exist in dcp_projectaction SHOULD NOT be copied over
+      // NOTE cont. -- The ZONING_RESOLUTION_TYPES lookup also functions as a list of actions that SHOULD...
+      // NOTE cont. -- be copied over from dcp_projectaction, which is why we check that `label` exists here.
+      if (!matchingZr && label) {
         return this.crmService.create(`dcp_affectedzoningresolutions`, {
-          'dcp_zoningresolutiontype': code, // this is a coded value
-          'dcp_zrsectionnumber': action.dcp_ZoningResolution ? action.dcp_ZoningResolution.dcp_zoningresolution : null,
+          'dcp_zoningresolutiontype': code, // coded value
+          'dcp_zrsectionnumber': currentActionZoningResolution,
           'dcp_modifiedzrsectionnumber': action.dcp_zrmodifyingzrtxt,
           'dcp_rwcdsform@odata.bind': `/dcp_pasforms(${dcp_rwcdsformid})`,
         });
--- a/server/src/packages/rwcds-form/rwcds-form.service.ts
+++ b/server/src/packages/rwcds-form/rwcds-form.service.ts
@@ -193,10 +193,10 @@ export class RwcdsFormService {
       }
 
       // IF the zr type does not exist yet, then POST the action
-      // NOTE: some actions that exist in dcp_projectaction SHOULD NOT be copied over
-      // NOTE cont. -- The ZONING_RESOLUTION_TYPES lookup also functions as a list of actions that SHOULD...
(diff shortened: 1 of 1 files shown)


godfreyyeung:
This makes me wish we had a Nest ORM to interface with CRM! Kinda like how in Ember we have Ember Data

godfreyyeung:
> Thanks for this Taylor! It looks like it's getting complicated... Wonder if there's any way we can simplify the code.
> 
> This makes me wish we had a Nest ORM to interface with CRM! Kinda like how in Ember we have Ember Data

Wondering if we could walk through this together at engie standup tomorrow just to make sure I understand it

trbmcginnis:
> Wondering if we could walk through this together at engie standup tomorrow just to make sure I understand it

Definitely!

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -172,13 +172,26 @@ export class RwcdsFormService {
 
       // Lookup: dcp_affectedzoningresolutions asks for a ZR type,
       const { label, code } = ZONING_RESOLUTION_TYPES.filter((zr) => zr.label === projectActionLabel)[0] || { label: '', code: null };
```
godfreyyeung, line 174: if there can only be one  zoning resolution that matches the project action we should use a find instead of a filter. That way we don't need to use `[0]`

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -172,13 +172,26 @@ export class RwcdsFormService {
...
+      const matchingZr = affectedZoningResolutions.find(zr => zr.dcp_zoningresolutiontype === code);
```
godfreyyeung, line 180: interesting... so there is no way to find the `matchingZr` by directly comparing `zr.dcp_zoningresolutiontype` to an attribute on `projectAction`? We have to use the `code` property in our own ZONING_RESOLUTION_TYPES lookup?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -172,13 +177,26 @@ export class RwcdsFormService {
...
+      // action attribute that matches dcp_zrsectionnumber on dcp_affectedzoningresolution
+      const currentActionZoningResolution = action.dcp_ZoningResolution ? action.dcp_ZoningResolution.dcp_zoningresolution : null
 
       if (!label) console.log(`Could not find Affected ZR Type for ${projectActionLabel}`);
 
+      // if the action.ZoningResolution.dcp_zoningresolution value has changed since the initial POST,
+      // then PATCH the new value to dcp_zrsectionnumber on dcp_affectedzoningresolution entity
+      // NOTE: dcp_modifiedzrsectionnumber not included here because it's updated by applicant on rwcds-form
+      if (zrTypes.includes(label) && label && matchingZr.dcp_zrsectionnumber !== currentActionZoningResolution) {
```
godfreyyeung, line 189: Here, can't we just check that `matchingZr` exists?
```
if(matchingZr && matchingZr.dcp_zrsectionnumber !== currentActionZoningResolution) {...}
```
Doesn't the existence of `matchingZr` imply that `label` is within `zrTypes`?

godfreyyeung: If there is a matching rwcds affected zoning resolution for the current project action, then the current project action's label should naturally be in the list of all rwcds affected zoning resolutions

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: good point!

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: I don't think dcp_projectaction entity has a field that has the corresponding codes, only a field that has labels (`ZC`, `ZA`)

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: Yes that's a good point! The use of `zrTypes` and `label` is left over from old logic/code, and just hasn't been changed yet to match new updates. Thanks for catching that :)

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: ahh ok!

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: np! Thanks for confirming. I think this PR looks good after this fix. 🚀

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -171,14 +176,30 @@ export class RwcdsFormService {
...
+      // action attribute that matches dcp_zrsectionnumber on dcp_affectedzoningresolution
+      const currentActionZoningResolution = action.dcp_ZoningResolution ? action.dcp_ZoningResolution.dcp_zoningresolution : null
 
       if (!label) console.log(`Could not find Affected ZR Type for ${projectActionLabel}`);
 
-      if (!zrTypes.includes(label) && label) {
+      // IF the action.ZoningResolution.dcp_zoningresolution value has changed since the initial POST,
+      // then PATCH the new value to dcp_zrsectionnumber on dcp_affectedzoningresolution entity
+      // NOTE: dcp_modifiedzrsectionnumber not included here because it's updated by applicant on rwcds-form
+      if (matchingZr && label && matchingZr.dcp_zrsectionnumber !== currentActionZoningResolution) {
```
godfreyyeung, line 189: Do you think there is ever a time when label will be undefined when `matchingZr` is undefined? 
It seems like that would only be the case if someone set up ZONING_RESOLUTION_TYPES incorrectly

trbmcginnis: You're right, that should never happen, so I guess this was just defensive coding / trying to have two layers of checks. But no I don't think it's actually necessary. I can remove

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -171,14 +176,30 @@ export class RwcdsFormService {
...
+        return this.crmService.update('dcp_affectedzoningresolutions', matchingZr.dcp_affectedzoningresolutionid, {
+          dcp_zrsectionnumber: currentActionZoningResolution,
+        });
+      }
+
+      // IF the zr type does not exist yet, then POST the action
+      // NOTE: some actions that exist in dcp_projectaction SHOULD NOT be copied over
+      // NOTE cont. -- The ZONING_RESOLUTION_TYPES lookup also functions as a list of actions that SHOULD...
+      // NOTE cont. -- be copied over from dcp_projectaction, which is why we check that `label` exists here.
+      if (!matchingZr && label) {
```
godfreyyeung, line 199: If matchingZr doesn't exist, why does it matter that label exist? 
It seems like more important there should be a check that `code` exists

trbmcginnis: It doesn't matter whether it's `code` vs `label`. What DOES matter is that the action exists in the ZONING_RESOLUTION_TYPES. This lookup also functions as a list of actions that SHOULD be copied over. For example, every rwcds-form requires an `EAS` action but this action should never be copied over to the dcp_affectedzoningresolution entity. So when looping through the dcp_projectactions entity, the EAS action would produce an undefined label/code. When we don't have that extra label/code check here, the function will POST an extra row with a null `dcpZoningresolutiontype` to dcp_affectedzoningresolution (because it would be POSTing this EAS action). In order to keep this from happening, we need to make sure that code/label exists

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -171,14 +176,30 @@ export class RwcdsFormService {
...
+      // NOTE: dcp_modifiedzrsectionnumber not included here because it's updated by applicant on rwcds-form
+      if (matchingZr && matchingZr.dcp_zrsectionnumber !== currentActionZoningResolution) {
+        return this.crmService.update('dcp_affectedzoningresolutions', matchingZr.dcp_affectedzoningresolutionid, {
+          dcp_zrsectionnumber: currentActionZoningResolution,
+        });
+      }
+
+      // IF the zr type does not exist yet, then POST the action
+      // NOTE: some actions that exist in dcp_projectaction SHOULD NOT be copied over
+      // NOTE cont. -- The ZONING_RESOLUTION_TYPES lookup also functions as a list of actions that SHOULD...
```
godfreyyeung, line 197: tiny style nit: Not sure about these `NOTE cont.` comment prefixes 
I think it's fine to just continue the sentence on the subsequent line

server/src/packages/rwcds-form/rwcds-form.service.ts
godfreyyeung: Ah gotcha, that makes sense, thank you for clarifying. Sorry that's my bad for missing it in your comment above. I think `code` will be more clear here, since ultimately that is posted to the new affected zoning resolution.

trbmcginnis: You're all good! Sounds good, I'll change it to `code` :)

server/src/packages/rwcds-form/rwcds-form.service.ts
trbmcginnis: lol sounds good :)

PR #596, opened 2020-08-05
526 Display JSONAPI errors in frontend

godfreyyeung:
This PR unifies the frontend *-error templates
in how they render JSONAPI errors from the backend.
It creates a component, which the templates share,
that iterates through the incoming `errors` array
and renders the keys in each error object.

complements #562

### DIFF
--- a/client/app/templates/projects-error.hbs
+++ b/client/app/templates/projects-error.hbs
@@ -5,20 +5,9 @@
 <div class="text-center xlarge-padding-top xlarge-padding-bottom xlarge-margin-top xlarge-margin-bottom">
   <span>Hmm, something went wrong:</span>
 
-  {{#each this.model.errors as |error|}}
-    <p class="font-mono text-red-dark bg-red-white small-padding text-small" data-test-error-message>
-      Message: {{error.message}} <br>
-      Status: {{error.status}} <br>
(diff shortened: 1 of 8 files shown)


client/app/templates/projects-error.hbs
```diff
@@ -5,20 +5,9 @@
...
-          </li>
-        {{/each-in}}
-      </ul>
-    </p>
-  {{/each}}
+  <Errors::List
+    @errors={{this.model.errors}}
+  />
 
   {{!-- If there's an error, try to logout --}}
```
trbmcginnis, line 12: What does this comment mean? Will this happen automatically?

godfreyyeung: The logout will happen automatically. `<Auth::DoLogout />` will automatically try to load an iframe that hits an NYCID logout endpoint. Note that the logout code isn't part of this PR, though.

godfreyyeung:
> Cool! So this probably makes #515 obsolete, since you've simplified these error templates so much. If you feel like it does, then maybe you can link out to this one as well in the PR description and then close it?

Actually, was about to jump on #515 next. It's still relevant, and can be quickly worked in to these series of frontend error display improvement.

Issue #577, opened 2020-07-31
RWCDS packages are appearing on the applicant portal that do not have applicant only/general public visibility 


ksinghprod:
## Browser:
Chrome,IE, Firefox
## Project:
Sprint N testing 

RWCDS packages are appearing on the applicant portal that do not have applicant only/general public visibility 

![image](https://user-images.githubusercontent.com/57362229/89058684-bad91780-d32d-11ea-9b18-b9185d86678d.png)

![image](https://user-images.githubusercontent.com/57362229/89058564-82393e00-d32d-11ea-8278-3b934dc558e2.png)

![image](https://user-images.githubusercontent.com/57362229/89058633-a268fd00-d32d-11ea-8fce-6ea7365524c2.png)

nnxka:
Solution:
- Review backend logic

trbmcginnis:
@ksinghprod Do you know if there is a way to manually update the Package Status Reason (`statuscode` field) on a package in CRM? I know that submitting the rwcds-form will update the package status reason to `Submitted` but I'm trying to manually update this field on a package to one of these values: `Mistake`/`Certified`/`Final Approval`/`Withdrawn`. This is for testing purposes, so that I can make sure projects with these statuses do NOT show up on the applicant portal.

ksinghprod:
Hi Taylor, I replied to your comment in the ZAP channel 

You can click on the deactivate button (when the package is still active) on the package record and you will see a list of status reasons

PR #602, opened 2020-08-10
577: filter packges on project page for visibility and statuscode

trbmcginnis:
Add package filters to GET for single project based on package `dcp_visibility` and `statuscode`

Addresses #577

### DIFF
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -11,6 +11,18 @@ const PROJECT_ACTIVE_STATE_CODE = 0;
 const PROJECT_VISIBILITY_APPLICANT_ONLY = 717170002;
 const PROJECT_VISIBILITY_GENERAL_PUBLIC = 717170003;
 
+const PACKAGE_VISIBILITY = {
+  APPLICANT_ONLY: 717170002,
+  GENERAL_PUBLIC: 717170003,
+}
+const PACKAGE_STATUSCODE = {
+  PACKAGE_PREPARATION: 1,
+  SUBMITTED: 717170012,
+  UNDER_REVIEW: 717170013,
+  REVIEWED_NO_REVISIONS_REQUIRED: 717170009,
+  REVIEWED_REVISION_REQUIRED: 717170010,
+}
+
 @Injectable()
 export class ProjectsService {
   constructor(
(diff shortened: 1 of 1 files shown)


godfreyyeung:
Cool! Previously we sent all packages to the frontend and filtered on the frontend, but I think it is probably a little better to do so in the backend.

trbmcginnis:
> Cool! Previously we sent all packages to the frontend and filtered on the frontend, but I think it is probably a little better to do so in the backend.

Were we originally filtering for `statuscode` and `dcp_visibility` in the frontend or just on other fields? I know `package-list-item.hbs` has a lot of display logic based on `statuscode`, but was there also filtering at some point?

godfreyyeung:
> > Cool! Previously we sent all packages to the frontend and filtered on the frontend, but I think it is probably a little better to do so in the backend.
> 
> Were we originally filtering for `statuscode` and `dcp_visibility` in the frontend or just on other fields? I know `package-list-item.hbs` has a lot of display logic based on `statuscode`, but was there also filtering at some point?

Yup, previously in `models/project.js` (This is before we introduced the projects/:project_id` page) , we had the `rwcdsPackages` and `pasPackages` computed properties:
https://github.com/NYCPlanning/labs-applicant-portal/blob/4023e3658deeffd63b96e00a3ca990d98374002e/client/app/models/project.js#L30

trbmcginnis:
> > > Cool! Previously we sent all packages to the frontend and filtered on the frontend, but I think it is probably a little better to do so in the backend.
> > 
> > 
> > Were we originally filtering for `statuscode` and `dcp_visibility` in the frontend or just on other fields? I know `package-list-item.hbs` has a lot of display logic based on `statuscode`, but was there also filtering at some point?
> 
> Yup, previously in `models/project.js` (This is before we introduced the projects/:project_id`page) , we had the`rwcdsPackages`and`pasPackages` computed properties:
> https://github.com/NYCPlanning/labs-applicant-portal/blob/4023e3658deeffd63b96e00a3ca990d98374002e/client/app/models/project.js#L30

We still have those computed properties in the project model but they don't include filtering for `dcpVisibility` or `statuscode`, they just have filtering for `dcpPackagetype`. Are you saying that the original plan was to include any extra filters (e.g. `dcpVisibility`) in these computed properties?

godfreyyeung:
> > > > Cool! Previously we sent all packages to the frontend and filtered on the frontend, but I think it is probably a little better to do so in the backend.
> > > 
> > > 
> > > Were we originally filtering for `statuscode` and `dcp_visibility` in the frontend or just on other fields? I know `package-list-item.hbs` has a lot of display logic based on `statuscode`, but was there also filtering at some point?
> > 
> > 
> > Yup, previously in `models/project.js` (This is before we introduced the projects/:project_id`page) , we had the`rwcdsPackages`and`pasPackages` computed properties:
> > https://github.com/NYCPlanning/labs-applicant-portal/blob/4023e3658deeffd63b96e00a3ca990d98374002e/client/app/models/project.js#L30
> 
> We still have those computed properties in the project model but they don't include filtering for `dcpVisibility` or `statuscode`, they just have filtering for `dcpPackagetype`. Are you saying that the original plan was to include any extra filters (e.g. `dcpVisibility`) in these computed properties?

Oh you're right. Nevermind, Ignore my comment.

server/src/projects/projects.service.ts
```diff
@@ -10,6 +10,14 @@ const APPLICANT_ACTIVE_STATUS_CODE = 1;
 const PROJECT_ACTIVE_STATE_CODE = 0;
 const PROJECT_VISIBILITY_APPLICANT_ONLY = 717170002;
 const PROJECT_VISIBILITY_GENERAL_PUBLIC = 717170003;
+const PACKAGE_VISIBILITY_APPLICANT_ONLY = 717170002;
```
godfreyyeung, line 13: Since this list is getting long, one nice thing you can do to organize the code is placing these under an object. like:
```js
 PACKAGE_VISIBILITY = {
    APPLICANT_ONLY: 71717003,
    GENERAL_PUBLIC: 71717004,
    ...
}
```

PR #609, opened 2020-08-12
Standardize exception handling for login

godfreyyeung:
Will properly display error for the /login route, like so:
image

Also document the need for logging out errors

### DIFF
--- a/server/EXCEPTIONS.md
+++ b/server/EXCEPTIONS.md
@@ -1,7 +1,23 @@
 # Exception handling in our backend
 
+## Quickstart example
+```ts
+// services/my-project.service.ts
+public async getProject(){
+  try {
+    return myProject = this.get('dcp_project', ...);
+  } catch (e) {
+    const error = {
+      code: 'NO_PROJECT_FOR_ID',
+      title: 'No project for id',
+      detail: `Failed to find a project for given id. ${e.message}`,
+    };
+    console.log(error); // for backend log records
+    throw new HttpException(error, HttpStatus.NOT_FOUND);
+  }
+}
+```
 ## Quickstart
-
 To generate errors in the backend that will be passed to the
 frontend, throw a [Nest-defined HttpException](https://docs.nestjs.com/exception-filters#throwing-standard-exceptions), <u>_generally within a service_</u>:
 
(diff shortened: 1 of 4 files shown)


server/src/app.controller.ts
```diff
@@ -29,11 +30,15 @@ export class AppController {
...
-        res.status(401).send({ errors: [e] });
+        throw e;
       } else {
-        console.log(e);
-
-        res.status(500).send({ errors: [e] });
+        const error = {
+          code: "LOGIN_ERROR",
+          title: "Failed to login",
+          detail: "An unknown error occured while authenticating user",
```
trbmcginnis, line 38: So this is like the "catch all" net that catches anything that slipped through the service correct?

server/src/contact/contact.service.ts
```diff
@@ -35,9 +35,13 @@ export class ContactService {
 
       return firstRecord;
     } catch(e) {
-      const errorMessage = `Error finding contact by ID. ${e.message}`;
+      const errorMessage = `Error finding contact by ID, possibly due to missing or bad ID. ${e.message}`;
       console.log(errorMessage);
-      throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+      throw new HttpException({
```
trbmcginnis, line 40: Why is this formatted differently from the ones in `auth.service.ts`?

server/src/app.controller.ts
godfreyyeung: yep, exactly what I was thinking!

server/src/contact/contact.service.ts
godfreyyeung: Ah, good catch. I should format it the same as auth.service.ts!

godfreyyeung: Fixed! Thanks!

server/EXCEPTIONS.md
```diff
@@ -1,7 +1,23 @@
...
+public async getProject(){
+  try {
+    return myProject = this.get('dcp_project', ...);
+  } catch (e) {
+    const error = {
+      code: 'NO_PROJECT_FOR_ID',
+      title: 'No project for id',
+      detail: `Failed to find a project for given id. ${e.message}`,
+    };
+    console.log(error); // for backend log records
```
godfreyyeung, line 15: Can we use out of the box HttpException logging features?

PR #611, opened 2020-08-12
Use YES_NO optionsets in forms

godfreyyeung:
This helps futureproof the forms in case CRM optionset values
are modified.


godfreyyeung:
> Yay to the optionset helper!
😁 
Thanks Taylor!

PR #613, opened 2020-08-13
🐞 Package Lists Improvements

andycochran:
This PR fixes [AB#12123](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12123) by improving the package lists on the project view: 
- Conditionally show headers 
- Put all items in one list (they were previously each in their own `<ol>`, resulting in a one-item list for each item)

### DIFF
--- a/client/app/components/project/package-list-item.hbs
+++ b/client/app/components/project/package-list-item.hbs
@@ -1,35 +1,33 @@
-<ol class="no-bullet">
-  <li>
-    {{#if (eq @package.statuscode (optionset 'package' 'status' 'code' 'REVIEWED_REVISIONS_REQUIRED'))}}
-      <FaIcon @icon="exclamation-triangle" class="text-gold" />
-    {{else}}
-      <FaIcon
-        @icon={{if (eq @package.statuscode (optionset 'package' 'status' 'code' 'PACKAGE_PREPARATION')) 'edit' 'check'}}
-        class={{if (eq @package.statuscode (optionset 'package' 'status' 'code' 'PACKAGE_PREPARATION')) '' 'text-green-dark'}}
-      />
-    {{/if}}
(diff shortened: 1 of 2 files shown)


client/app/components/project/package-list-item.hbs
```diff
@@ -1,35 +1,33 @@
...
-        <strong>Version {{@package.dcpPackageversion}}</strong>
-      </LinkTo>
-    {{/if}}
-    <span class="text-gray">| {{if (eq @package.statuscode (optionset 'package' 'status' 'code' 'PACKAGE_PREPARATION')) 'Editable since ' 'Submitted '}}
-      <Ui::DateDisplay @date={{@package.dcpStatusdate}} @outputFormat="MM/D/YYYY" @errorMessage="Date Unknown"/>
-    </span>
-  </li>
-</ol>
+<li>
+  {{#if (eq @package.statuscode (optionset 'package' 'status' 'code' 'REVIEWED_REVISIONS_REQUIRED'))}}
```
godfreyyeung, line 2: Yay, optionsets! :) Just noting here that in my incoming PR, I am changing the `status` key to `statuscode`, since `statuscode` is actually the field name on the Package model. Just thought it would be more predictable, and developers wouldn't need to learn the idiosyncratic field-key mappings in the optionset helper.

godfreyyeung: Either one of these two PRs will have to be updated depending on which is merged first.

godfreyyeung: Oh wait, these are just indentation changes !

andycochran: Yeah, just whitespace. Hopefully git can sort that out.

PR #615, opened 2020-08-13
🐞 fix typos

andycochran:
Fixes [AB#12124](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12124).


andycochran:
Um… why would codecov fail on `do-logout.js` when I've only changed a typo. 🤔

Issue #580, opened 2020-08-03
When the User clicks on the RWCDS form/package, the RWCDS form continues to load infinitely 


ksinghprod:
## Browser 
Chrome

## Project Name
Sprint N Revision Loop Test

When the User clicks on the RWCDS form/package, the RWCDS form continues to load infinitely 

![image](https://user-images.githubusercontent.com/57362229/89197500-cd479100-d579-11ea-92ac-0f7220879c43.png)

![image](https://user-images.githubusercontent.com/57362229/89197456-bacd5780-d579-11ea-8bf0-c40c01c8b4f7.png)

nnxka:
- This bug occured with Sprint N revision loop project

allthesignals:
@godfreyyeung just adding a note here for investigating why our error handling didn't capture this

allthesignals:
This also _might_ be related to the token expiration issue, might not

ksinghprod:
This is happening for all new projects I am creating (following the correct logic)

Another Project: Sprint N End to End to test

godfreyyeung:
@ksinghprod I'm pretty sure the error and infinite load is because that the project Sprint N Revision Loop Test has a RWCDs package **but not** a PAS package. The RWCDS package requires an existing PAS package because it copies some initial valuers from the PAS.

![image.png](https://images.zenhubusercontent.com/5ced7833554ce77a5db49edf/10f14084-b17d-465d-9e3b-0a0b28be3e60)

However, @allthesignals  is right in that we need to make sure the error is properly propogated to the frontend.

hannahkates:
@bean75 How often do you think there will be a project that has RWCDS package and not a PAS package?

CC @nnxka

bean75:
Hard to say. Not too frequently.
________________________________

godfreyyeung:
Okay, we will accommodate those edge cases. Thank you Aline.

Also, will transition this conversation to Azure now.

PR #616, opened 2020-08-13
12127 Display original RWCDS description fields if no PAS

godfreyyeung:
If there is no PAS form on a project, then while loading the RWCDS form, don't copy PAS "project description"  field values into the RWCDS. Just display the usually blank RWCDS form values for those fields.

Addresses #580

### DIFF
--- a/server/src/packages/pas-form/pas-form.service.ts
+++ b/server/src/packages/pas-form/pas-form.service.ts
@@ -43,16 +43,24 @@ export class PasFormService {
           dcp_pasform
       `);
 
-      // if pasA is of a higher version than pasB, it should come first
-      const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
-        return pasB.dcp_packageversion - pasA.dcp_packageversion;
-      });
+      if (pasPackages.length > 0) {
+        // if pasA is of a higher version than pasB, it should come first
+        const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
+          return pasB.dcp_packageversion - pasA.dcp_packageversion;
+        });
 
-      return latestPasForm;
+        return latestPasForm;
+      }
+
+      return null;
     } catch (e) {
-      const errorMessage = `Error finding a PAS Form on given Project. ${e.message}`;
-      console.log(errorMessage);
-      throw new HttpException(errorMessage, HttpStatus.UNAUTHORIZED);
+      const error = {
+        code: `LATEST_PAS_ERROR`,
+        title: `Error finding PAS form`,
+        detail: `Error finding the latest PAS Form on given Project. ${e.message}`,
+      };
+      console.log(error);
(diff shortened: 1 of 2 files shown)


server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -43,16 +43,24 @@ export class PasFormService {
...
+        // if pasA is of a higher version than pasB, it should come first
+        const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
+          return pasB.dcp_packageversion - pasA.dcp_packageversion;
+        });
 
-      return latestPasForm;
+        return latestPasForm;
+      }
+
+      return null;
```
godfreyyeung, line 55: @NYCPlanning/engineering I thought about returning a blank object `{}` here instead of a null. Or maybe an object with null values for the expected fields. e.g.
```
return {
  dcp_projectdescriptionproposedarea: null, 
 dcp_projectdescriptionproposeddevelopment: null,
}
```
But ultimately think `null` most clearly conveys to consumers of this function that no results were found. Thoughts?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -132,17 +132,24 @@ export class RwcdsFormService {
...
-
-      if (rwcdsForm.dcp_projectsitedescription === null) {
-        rwcdsForm.dcp_projectsitedescription = dcp_projectdescriptionproposedarea;
-      }
-
-      if (rwcdsForm.dcp_proposedprojectdevelopmentdescription === null) {
-        rwcdsForm.dcp_proposedprojectdevelopmentdescription = dcp_projectdescriptionproposeddevelopment;
+     const latestPasForm = await this.pasFormService.getLatestPasForm(_dcp_projectid_value);
+
+     if (latestPasForm) {
```
godfreyyeung, line 137: Here, the consuming function of `getLatestPasForm` checks for if the return value is `null`

server/src/packages/pas-form/pas-form.service.ts
andycochran: I think your instincts are right. It make sense that asking for PAS values would return `null` because a PAS doesn't exist.

server/src/packages/pas-form/pas-form.service.ts
```diff
@@ -43,16 +43,24 @@ export class PasFormService {
           dcp_pasform
       `);
 
-      // if pasA is of a higher version than pasB, it should come first
-      const [{ dcp_pasform: latestPasForm }] = pasPackages.sort((pasA, pasB) => {
-        return pasB.dcp_packageversion - pasA.dcp_packageversion;
-      });
+      if (!pasPackages) {
```
godfreyyeung, line 46: `if(pasPackages.length > 0)`
!!([]) actually evaluates to true

sidenote: handlebars makes it convenient for you to do this

godfreyyeung: @NYCPlanning/engineering 
Updated and resolved!

PR #619, opened 2020-08-18
12157: add project editor list to project page

trbmcginnis:
List of project editors displays on individual project page.

Fixes [AB#12157](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12157)

### DIFF
--- a/client/app/models/project-applicant.js
+++ b/client/app/models/project-applicant.js
@@ -1,0 +1,12 @@
+import Model, { attr, belongsTo } from '@ember-data/model';
+
+export default class ProjectApplicantModel extends Model {
+  @attr dcpName;
+
+  @attr emailaddress;
+
+  @attr dcpApplicantrole;
+
+  @belongsTo('project')
+  project;
+}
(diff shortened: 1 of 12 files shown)


client/app/models/project-applicant.js
```diff
@@ -0,0 +1,12 @@
+import Model, { attr, belongsTo } from '@ember-data/model';
```
godfreyyeung, line 1: Ah, so in CRM, pas applicant entities are different from project applicant entities! CRM is weird.

server/src/projects/project-applicants/project-applicants.attrs.ts
```diff
@@ -0,0 +1,6 @@
+export const PROJECTAPPLICANT_ATTRS = [
+  'dcp_name',
+  'emailaddress',
+  'dcp_projectapplicantid',
+  'dcp_applicantrole',
+];
```
godfreyyeung, line 6: nit: line break here

godfreyyeung:
Nice Taylor! Looks awesome!

server/src/projects/project-applicants/project-applicants.attrs.ts
andycochran: We should all set up our text editors to trim trailing whitespace and ensure one trailing new line upon save. Atom's "whitespace" package does this and I often get lots of whitespace diffs when making small changes to files others have worked on.

PR #621, opened 2020-08-18
12129 require upload

allthesignals:
# Summary:

Fixes [AB#12129](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12129)

This PR requires at least 1 document upload before submission.

However, this is more of a stopgap: it's still possible to mark a file for upload, skip "save", and still submit.

The "correct" solution for this requires an overhaul of the attachments component and the 'file manager' class. These classes cause a design problem because they maintain their own state. That means that the source of truth is neither on the changeset or the model. The model must itself update the related file-manager class that a change has occurred. This makes it difficult to validate changes within the curent system.

### Upload Validation ([ADO](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_sprints/taskboard/ZAP%20Portals%20Team/ZAP%20Portals/Sprint%20Q?workitem=12129)):
 - Wrap pasForm in saveableForm context for package.
 - Update the package changeset’s documents on marking/removing for upload
 - Make Save/Submit button state depends on the top-level package validity

### Refactor:
 - Move documents lookup to the module for documents — originally it was in the packages service.
 - Use a component for the error message. Centralizes the markup
 - Refactor refreshExistingDocuments to private synchronizeDocuments. This removes calls to this method directly from the pas-form components. This reduces the amount of public surface required to know about how to actually save a package.
 - Move model reload back to the model. Similar to the point above, this puts the call to the ED method "reload" into the model itself. This allows outside consumers to just know to call "save".

### Some cleanup tasks:
 - Remove unused unit spec files  and disable generating unit test spec in NestJS. These were seldom used. I kept the 1 or 2 that were used


client/app/components/packages/pas-form/edit.hbs
```diff
@@ -1,97 +1,103 @@
 <SaveableForm
-  @model={{this.pasForm}}
-  @validators={{array this.validations.SaveablePasFormValidations this.validations.SubmittablePasFormValidations}}
-  as |saveableForm|
+  @model={{@package}}
+  @validators={{array hash this.validations.SubmittablePackageFormValidations}}
```
godfreyyeung: why hash here?

allthesignals: the component requires 2 validators and we don't have any saveability validations for packages

godfreyyeung:
I don't understand the motivation for the Package Validations  and this item:
> Wrap pasForm in saveableForm context for package.

Why can't the documents validation happen on the PAS form? 
Maybe we can go over it in standup tomorrow?

client/app/components/packages/pas-form/edit.hbs
```diff
@@ -1,97 +1,103 @@
...
+        <Packages::PasForm::ProposedLandUseActions @form={{saveablePasForm}} />
+
+        <Packages::PasForm::ProjectArea @form={{saveablePasForm}} />
+
+        <Packages::PasForm::ProposedDevelopmentSite @form={{saveablePasForm}} />
+
+        <Packages::PasForm::ProjectDescription @form={{saveablePasForm}} />
+
+        <Packages::PasForm::AttachedDocuments
+          @form={{saveablePasForm}}
```
godfreyyeung: should this be `@form={{saveablePackageForm}}`?

allthesignals:
> I don't understand the motivation for the Package Validations and this item:
> 
> > Wrap pasForm in saveableForm context for package.
> 
> Why can't the documents validation happen on the PAS form?
> Maybe we can go over it in standup tomorrow?

Because documents aren't on the pasForm model. They're on the package model.

client/app/components/packages/pas-form/edit.hbs
allthesignals: Weird, let me see

godfreyyeung: Oh, you changed it in a later commit, i see

allthesignals: ooh

client/app/components/saveable-form/field-validation-message.hbs
```diff
@@ -0,0 +1,6 @@
+  <span
```
godfreyyeung, line 1: I had this component at one point and we got rid of it, haha 😂 🤷‍♂️ 
It used to be `validation-message`. My poor baby 🚼 😂 😂

client/app/models/package.js
```diff
@@ -71,13 +67,20 @@ export default class PackageModel extends Model {
 
   async save() {
     await this.fileManager.save();
+<<<<<<< HEAD
```
godfreyyeung: 😮

allthesignals: It's an outdated commit lol...

client/app/models/package.js
```diff
@@ -67,21 +67,14 @@ export default class PackageModel extends Model {
 
   async save() {
     await this.fileManager.save();
-<<<<<<< HEAD
     if (this.dcpPackagetype === DCPPACKAGETYPE.PAS_PACKAGE.code) {
-=======
```
godfreyyeung: nit:  should move this deletion to an earlier commit, just for ease of commit review and history cleanliness

client/app/models/package.js
```diff
@@ -67,21 +67,14 @@ export default class PackageModel extends Model {
...
-    if (this.dcpPackagetype === PACKAGE_TYPE.PAS_PACKAGE.code) {
->>>>>>> Refactor refreshExistingDocuments to private synchronizeDocuments
       await this.pasForm.save();
     }
     if (this.dcpPackagetype === DCPPACKAGETYPE.RWCDS.code) {
       await this.rwcdsForm.save();
     }
     await super.save();
 
-    await this.reload();
```
godfreyyeung: nit: same here

allthesignals:
> Looks good from what I can tell! I haven't tested it but trust it works. Can you either squash the commits, or rebase and clean the commits of the merge conflict markers (bonus points for lint errors).?
> I think removing merge conflict markers from our commit history will make sure someone doesn't get tripped up in the future if they have to deal with one of these commits
> 
> Sorry to nit ~ in memory of @bfreeds excellent commit hygiene 😂

yeah, sure, definitely. however, in general, i'm looking for high-level feedback on my approach and at the very least, testing out my feature. i'm glad you flag the messy commits — those things should definitely not be in there and I will fix them — but at some point they come across to me as [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality), in the absence of other more substantial feedback.

allthesignals:
> A lot of great clean-up stuff, thanks Matt! Also agree with Godfrey about the commit history clean-up, but other than that this looks great

Yep, thanks, but I already addressed those commit history issues.

PR #624, opened 2020-08-18
Natural Areas Districts copy fix

godfreyyeung:


allthesignals:
Sorry, I don't want master getting out of sync with develop. Please don't merge. Why is this urgent that we need to forgo our normal process?

What I'm trying to avoid is the issues we had with CEQR were master/develop were terribly out of sync.

PR #626, opened 2020-08-19
12149 creeper mode fix

allthesignals:
Fixes [AB#12149](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12149)

## Bug Fix ([ADO](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_sprints/taskboard/ZAP%20Portals%20Team/ZAP%20Portals/Sprint%20Q?workitem=12149)):
 - Re-enables creeper mode for all routes, regardless of method.

The changes here move the email-param-bypass behavior into the auth middleware. This middleware then looks up the passed query param (email), generates a new token, and fully simulates the authentication for the "creeped" contact.

Instead of sending the query param explicitly, the API infers the QP from the "referer" header. This seems good enough for an internal-facing debugging feature. Additionally, it removes "creeper mode" behavior from multiple parts of the application

## Refactor:
 - Removes the IMPOSTER_ID stuff which wasn't being used (and shouldn't be used).
 - Removes authorization logic from the project service — for some reason the project service was also verifying that the current session contactId was associated with the given project. This should be moved into a guard or handled by something like the package-access guard.
 - Remove creeper mode logic from routes and centralize to auth middleware.

### DIFF
--- a/server/src/auth.middleware.ts
+++ b/server/src/auth.middleware.ts
@@ -20,25 +20,30 @@ export class AuthMiddleware implements NestMiddleware {
 
     req.session = false;
 
-    const { authorization = '' } = req.headers;
+    const { authorization = '', referer } = req.headers;
     const token = authorization.split(' ')[1];
 
     try {
       // this promise will throw if invalid
       let validatedToken = await this.authService.validateCurrentToken(token);
 
-      const { email } = req.query; // the query param, email, sent from the client. this is who the "Creeper" wants to be.
-      const { mail } = validatedToken; // the "creepers" actual email, for verification.
+      // the referer is the domain issuing the request. in this case, it'll be the client domain,
+      if (referer) {
+        // which includes the query params
+        const refererParams = new URL(referer);
+        const email = refererParams.searchParams.get('email'); // the query param, email, sent from the client. this is who the "Creeper" wants to be.
+        const { mail } = validatedToken; // the "creepers" actual email, for verification.
 
-      // REDO: env variables.
-      // if an e-mail is provided, implicitly it means force creeper mode. then verify creeper mode
-      // with some criteria.
-      if (email && (mail === 'dcpcreeper@gmail.com' || mail.includes('@planning.nyc.gov'))) {
-        validatedToken = await this._spoofToken(validatedToken, email);
+        // REDO: env variables.
+        // if an e-mail is provided, implicitly it means force creeper mode. then verify creeper mode
+        // with some criteria.
+        if (email && (mail === 'dcpcreeper@gmail.com' || mail.includes('@planning.nyc.gov'))) {
+          validatedToken = await this._spoofToken(validatedToken, email);
+        }
       }
 
       req.session = validatedToken;
-       
+
       next();
     } catch (e) {
       next();
--- a/server/src/auth.middleware.ts
+++ b/server/src/auth.middleware.ts
@@ -1,9 +1,14 @@
 import { Injectable, NestMiddleware } from '@nestjs/common';
+import * as jwt from 'jsonwebtoken';
 import { AuthService } from './auth/auth.service';
+import { ConfigService } from './config/config.service';
 
 @Injectable()
 export class AuthMiddleware implements NestMiddleware {
-  constructor(private readonly authService: AuthService) { }
+  constructor(
+    private readonly authService: AuthService,
+    private readonly config: ConfigService,
+  ) { }
 
   async use(req: any, res: any, next: () => void) {
     // skip for the login route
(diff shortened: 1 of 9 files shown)


server/src/auth.middleware.ts
```diff
@@ -15,15 +20,52 @@ export class AuthMiddleware implements NestMiddleware {
...
-      req.session = await this.authService.validateCurrentToken(token);
+      // this promise will throw if invalid
+      let validatedToken = await this.authService.validateCurrentToken(token);
+
+      // the referer is the domain issuing the request. in this case, it'll be the client domain,
+      if (referer) {
+        // which includes the query params
+        const refererParams = new URL(referer);
+        const email = refererParams.searchParams.get('email'); // the query param, email, sent from the client. this is who the "Creeper" wants to be.
+        const { mail } = validatedToken; // the "creepers" actual email, for verification.
```
godfreyyeung, line 35: can we rename this to something like "authedEmail"

allthesignals: @godfreyyeung not really but we can alias it.

allthesignals: it's named as such because it's provided by NYC ID

server/src/auth.middleware.ts
```diff
@@ -15,15 +20,52 @@ export class AuthMiddleware implements NestMiddleware {
...
+
+        // REDO: env variables.
+        // if an e-mail is provided, implicitly it means force creeper mode. then verify creeper mode
+        // with some criteria.
+        if (email && (mail === 'dcpcreeper@gmail.com' || mail.includes('@planning.nyc.gov'))) {
+          validatedToken = await this._spoofToken(validatedToken, email);
```
godfreyyeung, line 41: this is cool

server/src/auth.middleware.ts
```diff
@@ -15,15 +20,52 @@ export class AuthMiddleware implements NestMiddleware {
...
+  async _spoofToken(validatedToken, creeperEmail) {
+    const NYCID_TOKEN_SECRET = this.config.get('NYCID_TOKEN_SECRET');
+
+    // these simulate the flow of authentication for the app
+    const spoofedNycIdToken = jwt.sign({
+      ...validatedToken,
+      mail: creeperEmail,
+    }, NYCID_TOKEN_SECRET);
+    const spoofedZapToken = await this.authService.generateNewToken(spoofedNycIdToken);
+    const validatedSpoofedToken = await this.authService.validateCurrentToken(spoofedZapToken);
```
godfreyyeung, line 65: I always wondered -- why do we need to validate a token we just signed?

server/src/packages/package-access.guard.ts
```diff
@@ -32,9 +32,15 @@ export class PackageAccessGuard implements CanActivate {
       },
       session: {
         contactId,
+        isCreeper = false,
       }
     } = context.switchToHttp().getRequest();
 
+    // because creeper mode allows authorization on all resources, skip this layer
+    if (isCreeper) {
```
godfreyyeung, line 40: <3 yess!!!! LOVE this

server/src/auth.middleware.ts
godfreyyeung: ohh gotcha

server/src/projects/projects.service.ts
```diff
@@ -66,14 +66,13 @@ export class ProjectsService {
...
    * @return     {any}     { ...Project Attributes }
    */
-  public async getProject(projectId: string, contactId: string) {
+  public async getProject(projectId: string) {
     try {
       const { records } = await this.crmService.get('dcp_projects', `
         $filter=
           dcp_dcp_project_dcp_projectapplicant_Project/
             any(o:
-              o/_dcp_applicant_customer_value eq '${contactId}'
```
godfreyyeung, line 75: Wait, why remove the filtering here? 
Doesn't this method return projects assigned to the current user? 
without this filter, it seems it would return all projects with active applicants

server/src/projects/projects.service.ts
```diff
@@ -102,6 +101,11 @@ export class ProjectsService {
       `);
 
       const [ project ] = this.overwriteCodesWithLabels(records);
+
+      if (!project) {
+        throw 'No projects found.';
```
godfreyyeung, line 106: Can we throw this with code, title, detail, status? 
For consistency for now. If we redesign the exception handling, it'll be easier to find and replace.

server/src/projects/projects.controller.ts
```diff
@@ -93,19 +87,9 @@ export class ProjectsController {
   }
 
   @Get('/projects/:id')
-  async projectById(@Session() session, @Param('id') id, @Query('email') email) {
-    let { contactId } = session;
```
godfreyyeung, line 97: why remove `let { contactId } = session;`?

server/src/auth.middleware.ts
allthesignals: @godfreyyeung yeah, good question. I think it's not quite the best name for the method. but the jwt method that's invoked there both validates and _parses/decrypts_ the token into an object.

server/src/projects/projects.service.ts
godfreyyeung: Oh nevermind, i see. this is for a single project. And we will set up a project access guard

server/src/projects/projects.controller.ts
godfreyyeung: nevermind!

server/src/projects/projects.service.ts
allthesignals: It would return a single project based on ID. If we want to enforce something, we should have a guard explicitly throw "False" which would generate an error.

server/src/projects/projects.service.ts
allthesignals: okay!

server/src/projects/projects.service.ts
allthesignals: yeah we don't really even need any of the other filters there except for the id. we could also use the correct rest convention `dcp_projects/{:id}`

server/src/auth.middleware.ts
godfreyyeung: I think `{ mail: authedMail }` would be great, just for code clarity

server/src/projects/projects.service.ts
godfreyyeung: Wait I don't get what you mean. except for which id? Also our endpoint does look like that: currently it is `projects/{:id}` ?

allthesignals:
@godfreyyeung thanks for the feedback. I've addressed your comments in the latest commit. I'm going to go ahead and merge.

PR #629, opened 2020-08-22
12019: replace projectApplicants with contacts in project editor list

trbmcginnis:
This PR employs a double `$expand` (two network requests) in order to get access to contacts that are associated with `dcp_projectapplicants` on the project entity. 

The purpose of this was to replace `dcp_projectapplicants` with `contacts` as list items on the project editor list on the project page.

### DIFF
--- a/server/src/contact/contacts.attrs.ts
+++ b/server/src/contact/contacts.attrs.ts
@@ -1,0 +1,7 @@
+export const CONTACT_ATTRS = [
+  'firstname',
+  'lastname',
+  'emailaddress1',
+  'contactid',
+];
+  
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -103,6 +103,13 @@ export class ProjectsService {
           )
       `);
 
+      const projectApplicantsWithContacts = await this.crmService.get('dcp_projectapplicants', `
+        $filter=
+          _dcp_project_value eq ${projectId}
+        &$expand=
+          dcp_applicant_customer_contact
+      `);
+
       const [ project ] = this.overwriteCodesWithLabels(records);
 
       if (!project) {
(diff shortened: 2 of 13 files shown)


server/src/contact/contacts.attrs.ts
```diff
@@ -0,0 +1,7 @@
+export const CONTACT_ATTRS = [
+  'firstname',
+  'lastname',
+  'emailaddress1',
+  'contactid',
+];
+
```
allthesignals, line 7: Newline newline newline

server/src/projects/projects.service.ts
```diff
@@ -116,7 +123,12 @@ export class ProjectsService {
         }, HttpStatus.NOT_FOUND);
       }
 
-      return project;
+      const projectWithContacts = {
+        ...project,
+        contacts: projectApplicantsWithContacts.records.map(applicant => applicant.dcp_applicant_customer_contact),
+      };
```
allthesignals, line 129: I know we discussed this at length, and I think this is okay for now, but I thought it'd be better to have our model relationships closely match the entity relationships in CRM. So, if CRM has projects -> project_applicant -> contact, then our app should do the same. 

I do think it's probably okay here but could get confusing upon return. Or maybe not. Something to keep an eye on, though, in terms of design...

For example, if we need to associate a contact with a different project, we would need the dcp_projectapplicant entity to do that. Contacts don't have a direct association with projects.

trbmcginnis: Yes that's a good point -- I originally tried to do it as project.projectApplicant.contact but for some reason I was having trouble getting access to the contacts in the frontend through the projectApplicants. Not sure what I was doing wrong, but I ended up using this shortcut as a fix, which probably isn't the most sustainable strategy.

PR #635, opened 2020-08-29
12020 deactivate projectapplicant

godfreyyeung:
Ready for handoff to @allthesignals  and @andycochran . 

I've added a basic confirmation modal for delete. However it still needs to match
the style/layout in the wireframes. I.e. The "Remove Editor" button should sit to the right of "Cancel".  Ideally, it also has the same behavior as the form "Save" button --
using tasks to reflect the save state (but this can be done in another PR).
image

Todos for other PRs:
- I think we can improve the underlying architecture of the project, project applicant classes.
For example, it's not great that we have "first name" and "last name" properties on the project
controller. 

- We can also rename the "toggle" parameter name on confirmation modal to "onClose" to better reflect its purpose. 

- We can abstract the save-button component to handle all generic async task based actions

### DIFF
--- a/client/app/components/project/project-editor-list-item.js
+++ b/client/app/components/project/project-editor-list-item.js
@@ -1,0 +1,34 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { optionset } from '../../helpers/optionset';
+
+export default class ProjectEditorListComponent extends Component {
+  get disableDelete() {
+    return this.args.disableDelete ?? false;
+  }
+
+  @tracked removeEditorModalOpen = false;
+
+  get isPrimaryApplicantOrContact() {
+    const { role } = this.args;
+    return role === optionset(['projectApplicant', 'applicantrole', 'code', 'PRIMARY_CONTACT'])
+      || role === optionset(['projectApplicant', 'applicantrole', 'code', 'PRIMARY_APPLICANT']);
+  }
+
+  @action
+  tryRemoveEditor() {
+    this.removeEditorModalOpen = true;
+  }
+
+  @action
+  async confirmRemoveEditor() {
+    await this.args.onDelete();
+    this.removeEditorModalOpen = false;
+  }
+
+  @action
+  cancelRemoveEditor() {
+    this.removeEditorModalOpen = false;
+  }
+}
--- a/client/app/controllers/project.js
+++ b/client/app/controllers/project.js
@@ -1,0 +1,45 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { optionset } from '../helpers/optionset';
+
+export default class ProjectController extends Controller {
+  @tracked modalOpen;
+
+  @tracked emailAddress;
+
+  @tracked firstName;
+
+  @tracked lastName;
+
+  @service
+  store;
+
+  get matchingCurrentApplicant() {
+    const currentApplicants = this.project.projectApplicants;
+    return currentApplicants.find((applicant) => applicant.emailaddress === this.emailAddress);
+  }
+
+  @action
+  addEditor() {
+    this.modalOpen = true;
+  }
+
+  @action
+  async saveEditor() {
+    this.modalOpen = false;
+    if (!this.matchingCurrentApplicant) {
+      const newApplicant = await this.store.createRecord('project-applicant', {
+        dcpName: `${this.firstName} ${this.lastName}`,
+        emailaddress: this.emailAddress,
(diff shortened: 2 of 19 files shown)


server/src/projects/project-applicants/project-applicant.controller.ts
```diff
@@ -7,7 +7,11 @@ import { JsonApiDeserializePipe } from '../../json-api-deserialize.pipe';
 import { PROJECTAPPLICANT_ATTRS } from './project-applicants.attrs';
 import { CONTACT_ATTRS } from '../../contact/contacts.attrs';
 
-const ACTIVE_CODE = 1;
+const ACTIVE_STATUSCODE = 1;
+const INACTIVE_STATUSCODE = 2;
+
+const ACTIVE_STATECODE = 0;
+const INACTIVE_STATECODE = 1;
```
trbmcginnis, line 14: just a warning, this might cause some conflicts -- I created some variables for these codes as well and named them slightly differently

server/src/projects/project-applicants/project-applicant.controller.ts
```diff
@@ -41,7 +45,7 @@ export class ProjectApplicantController {
     const { records } = await this.crmService.get('contacts', `
       $select=${CONTACT_ATTRS.join(',')}
       &$filter=startswith(emailaddress1, '${email}')
-        and statuscode eq ${ACTIVE_CODE}
+        and statuscode eq ${ACTIVE_STATUSCODE}
```
trbmcginnis, line 48: Also another note -- this was removed in another PR that's now merged into develop. We end up grab all contacts regardless of statuscode

server/src/projects/project-applicants/project-applicant.controller.ts
```diff
@@ -82,4 +86,25 @@ export class ProjectApplicantController {
...
     }
   }
+
+  @Delete('/:id')
+  async deactivate(@Param('id') id) {
+    try {
+      await this.crmService.update('dcp_projectapplicants', id, {
+        statuscode: INACTIVE_STATUSCODE,
+        statecode: INACTIVE_STATECODE,
+      });
```
trbmcginnis, line 96: ooh okay so we're not actually deleting the record, just deactivating it

client/app/components/project/project-editor-list-item.js
```diff
@@ -0,0 +1,34 @@
...
+  get isPrimaryApplicantOrContact() {
+    const { role } = this.args;
+    return role === optionset(['projectApplicant', 'applicantrole', 'code', 'PRIMARY_CONTACT'])
+      || role === optionset(['projectApplicant', 'applicantrole', 'code', 'PRIMARY_APPLICANT']);
+  }
+
+  @action
+  tryRemoveEditor() {
+    this.removeEditorModalOpen = true;
+  }
```
trbmcginnis, line 22: okay so this action just opens the initial modal before the user confirms? "try" feels a bit confusing to me, not sure what would be a better name though, maybe `openRemoveEditorModal` or something?

client/app/controllers/project.js
```diff
@@ -26,6 +26,11 @@ export default class ProjectController extends Controller {
     this.modalOpen = true;
   }
 
+  @action
+  removeEditor(applicant) {
+    applicant.destroyRecord();
+  }
```
trbmcginnis, line 32: So are you actually going to delete the record or just deactivate it? I know that we don't show deactivated contact records in the project editor list based on new changes that are in develop, so maybe we could do the same thing for `projectApplicant`?

client/app/components/project/project-editor-list-item.js
godfreyyeung: @trbmcginnis  Yeah, I also had a hard time naming this. `openRemoveEditorModal` sounds good!

client/app/controllers/project.js
godfreyyeung: @trbmcginnis deactivate it in CRM. However, to the frontend it is deleting it. 

Calling `destroyRecord()` will send a DELETE request to the backend, however the backend DELETE endpoint actually just updates the contact record to be inactive. 

Wait... isn't the project editor list the same as a list of project applicants?

trbmcginnis: We call it a "project editor" in the frontend but the entity that it's referring to in the backend is `dcp_projectapplicants`

allthesignals: @trbmcginnis is right here. It should not be a destroyRecord, it should be a regular patch request to update the field.

server/src/projects/project-applicants/project-applicant.controller.ts
allthesignals: yeah, but it shouldn't work like this — the http verbs should match what they're actually doing.

PR #642, opened 2020-09-01
💅 update auth templates/layouts

andycochran:
So pretty.


allthesignals:
THank uuu

allthesignals:
image

@andycochran only thing I don't understand is this part but that's it

PR #645, opened 2020-09-02
12018: New Auth Workflow

allthesignals:
Addresses [AB#12018](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12018)

Creates a new service, nycid.service, which is responsible for issuing authenticated requests to the NYC.ID platform, and return a structured object containing standard information about the state of the inferred/associated contact record's NYC.ID account.

Adjusts contact.service to use nycid.service to include that information in its own contact data that it returns. Finally, consuming client and other services (auth.service) now react to information from that by performing lookups, creating new records, and syncing critical data between the systems.

Frontend changes here include new routes for auth which work by directing user to certain instruction pages depending on the state of the contact object.

Future improvements:

 - Make NYCID it's own first-class resource with a front-end model. There's too much going on in contact service, and a lot of that responsibility is handled by auth, which needs to know way too much about contact.service.
 - Make the sign-in-button step route-driven, instead of passing a closure action. This would be a cleaner design and better for testing. Also, less prone to "lies" about the user (as it's all based on the QPs, not actual data).
 - Auth should be expanded to require email validation for a user to be truly authenticated.


godfreyyeung:
Should this merge into either `develop` or the `12018-new-auth` branch?

client/app/components/auth/sign-in.js
```diff
@@ -0,0 +1,35 @@
...
+      isNycidEmailRegistered,
+      isCityEmployee,
+    } = await this.args.searchContacts(loginEmail);
+
+    if (isNycidEmailRegistered) {
+      // we can only know email validation after they logged in once.
+      // if it's null, it means we don't know because they've never logged in before (null).
+      if (isNycidValidated === true || isNycidValidated === null) {
+        this.router.transitionTo('auth.login', { queryParams: { loginEmail, isCityEmployee } });
+      } else if (isNycidValidated === false) {
```
andycochran, line 28: Is it necessary to check `isNycidValidated` here? Since we can't know if an email is validated until after the first successful login, perhaps `startLogin()` doesn't even need to check for it. We could instead check if the email is validated upon login, and if it's not validated `doLogout()` and redirect the user to the `auth.validate` route.

allthesignals: Yeah! I think that makes sense.

PR #657, opened 2020-09-03
Return "is-registered" flag for non-email accounts

allthesignals:
Fixes minor bug which allowed non-email types to return a "is-registered" 

Bumps test coverage

### DIFF
--- a/server/src/contact/nycid/nycid.service.ts
+++ b/server/src/contact/nycid/nycid.service.ts
@@ -57,8 +57,17 @@ export class NycidService {
       errors = e.response.body.ERRORS;
     }
 
+    const hasUnknownEmail = errors.hasOwnProperty('cpui.unknownEmail');
+
+    // this implies that an e-mail exists in the system. however, because NYC.ID
+    // does not enable this feature for users who are authenticated through OAUTH,
+    // it returns an "unauthorized" error: The search is unauthorized. This means that
+    // the user exists in the system, but searching for them isn't allowed.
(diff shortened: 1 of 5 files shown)


server/src/contact/nycid/nycid.service.ts
```diff
@@ -57,8 +57,17 @@ export class NycidService {
       errors = e.response.body.ERRORS;
     }
 
+    const hasUnknownEmail = errors.hasOwnProperty('cpui.unknownEmail');
```
godfreyyeung, line 60: what does unknownEmail represent?

server/src/contact/nycid/nycid.service.ts
```diff
@@ -57,8 +57,17 @@ export class NycidService {
...
+
+    // this implies that an e-mail exists in the system. however, because NYC.ID
+    // does not enable this feature for users who are authenticated through OAUTH,
+    // it returns an "unauthorized" error: The search is unauthorized. This means that
+    // the user exists in the system, but searching for them isn't allowed.
+    const hasUnauthorizedSearch = errors.hasOwnProperty('cpui.unauthorized');
+    const is_nycid_email_registered = !hasUnknownEmail && hasUnauthorizedSearch;
```
godfreyyeung, line 67: What does it mean if there is a "known" email but is searchable?

allthesignals: @godfreyyeung searchable means they have logged in to our application through the non-existent SAML implementation.

PR #661, opened 2020-09-03
12256 Land Use form Tax Lots selection

godfreyyeung:
Set up the backend and frontend for Tax Lot selection on the Land Use form.

In the backend, the bbl controller and service is relocated to the /package level so that
it can be abstracted and reused for Land Use.


server/src/packages/bbls/bbls.controller.ts
```diff
@@ -34,17 +34,22 @@ export class BblsController {
   create(@Body() body) {
     const allowedAttrs = pick(body, BBL_ATTRS);
 
-    if (!body.pas_form || !body.project) throw new HttpException(
-      'Missing pas_form or project ids',
-      HttpStatus.UNPROCESSABLE_ENTITY,
-    );
+    if (
+      !(body.pas_form || body.landuse_form)
+      || !body.project
```
allthesignals, line 39: Hmmm, this is interesting.

So, we have many different types of forms, and will have more. It seems like this is making sure its dependencies — various relationships — are included. This is sort of a validation step.

I don't think it's worth doing right now but maybe there is another pipe of some sort, before the deserializer, that validates the input, based on some validation schema.

server/src/packages/packages.controller.ts
```diff
@@ -174,6 +181,7 @@ import { APPLICANT_ATTRS } from './pas-form/applicants/applicants.attrs';
                 dcp_applicantinformationid: `representative-${applicant.dcp_applicantrepresentativeinformationid}`
               }}),
             ],
+            bbls: landuseForm.dcp_dcp_projectbbl_dcp_landuse,
```
allthesignals, line 184: What are some ways to make this branching logic more organized? Maybe each form's service owns how it constructs the object?

allthesignals (approved):

server/src/packages/packages.controller.ts
godfreyyeung: Interesting idea. Is it possible to activate a Serializer on a controller that isn't the endpoint getting called? I can look into it...

server/src/packages/bbls/bbls.controller.ts
godfreyyeung: Thanks for this, I think it's a good idea! Will create a task for it

PR #662, opened 2020-09-04
12259: create related action component for landuse form

trbmcginnis:
Create related actions component for landuse form.

**Frontend behavior:**
- users can add new related actions by clicking on the "add related action" button -- this will display a new related actions form each time the user clicks the button
- users are able to add/edit/remove these related actions
- there is a `related-action-fieldset.hbs` to handle the related action form, a `related-action.hbs` to handle the add button and list out a form per related action, and this `related-action.hbs` component is rendered on the `landuse-form/edit.hbs`
- ember tests were added to test adding/removing applicants and adding/removing related-actions on the landuse form

**Backend behavior:**
- a `related-actions.controller.ts` was created to handle PATCH, POST, and DELETE. The `related-actions` entity GET is on an `expand` in the `landuse-form.service.ts` 

Closes [AB#12259](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12259)


client/tests/acceptance/user-can-click-landuse-form-edit-test.js
```diff
@@ -0,0 +1,132 @@
...
+    const { landuseForm } = project.packages.models[0];
+    // create an applicant model
+    const serverSideApplicant = this.server.create('applicant', 'organizationApplicant', { landuseForm });
+    // get the reference to the model instance
+    const applicant = await this.owner.lookup('service:store').findRecord('applicant', serverSideApplicant.id);
+
+    await visit('/landuse-form/1/edit');
+
+    await assert.equal(applicant.hasDirtyAttributes, false);
+    await assert.equal(applicant.isDeleted, false);
```
allthesignals, line 86: 1. No need to await these assertions — they are synchronous and don't return any promises.
2. Why do we need to assert this model state? Is there some reason we can't trust that the ember-data API might not work?

server/src/json-api-deserialize.pipe.ts
```diff
@@ -33,6 +33,9 @@ export class JsonApiDeserializePipe implements PipeTransform {
         contacts: {
           valueForRelationship: relationship => relationship.id,
         },
+        'related-actions': {
+          valueForRelationship: relationship => relationship.id,
+        },
       }).deserialize(value);
```
allthesignals, line 39: I hope this wasn't as big a headache for as it was last time

client/app/components/packages/related-action-fieldset.hbs
```diff
@@ -0,0 +1,79 @@
...
+      class="delete-fieldset"
+      {{on "click" (fn @removeRelatedAction @relatedAction)}}
+      data-test-remove-related-action-button
+    >
+      Remove Related Action
+      <FaIcon @icon="times" @prefix="fas" @fixedWidth={{true}} />
+    </button>
+
+  </Ui::Question>
+{{/let}}
```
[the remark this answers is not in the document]
trbmcginnis: will do!

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
```diff
@@ -0,0 +1,132 @@
...
+
+    await visit('/landuse-form/1/edit');
+
+    await click('[data-test-add-applicant-button]');
+    await fillIn('[data-test-input="dcpFirstname"]', 'Tess');
+    await fillIn('[data-test-input="dcpLastname"]', 'Ter');
+    await fillIn('[data-test-input="dcpEmail"]', 'tesster@planning.nyc.gov');
+    await click('[data-test-save-button]');
+
+    assert.equal(this.server.db.applicants.firstObject.dcpFirstname, 'Tess');
```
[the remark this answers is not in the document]
trbmcginnis: I guess we were asserting against the fake db to check if the data had actually saved -- but it might be better to transition to the `landuse-form.show` route and then assert against the dom there. I don't think the `landuse-form.show` route is built out yet, but maybe part of building that out will be updating these tests to check for the presence of the correct data in the dom.

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
trbmcginnis: These tests were copied over from how we were doing them in the pas-form, might be outdated code. I don't have any problem removing them

server/src/json-api-deserialize.pipe.ts
trbmcginnis: not at all :)

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
```diff
@@ -0,0 +1,132 @@
...
+    await click('[data-test-remove-applicant-button');
+
+    // should trigger dirty state, be queued for deletion when user saves
+    await assert.equal(applicant.hasDirtyAttributes, true);
+    await assert.equal(applicant.isDeleted, true);
+
+    // FIXME: user shouldn't see the fieldset
+    assert.dom('[data-test-applicant-fieldset="0"]').doesNotExist();
+
+    await saveForm();
```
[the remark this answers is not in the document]
trbmcginnis: It was already created :) -- we have a lot of awesome testing code that can help us build tests faster in the future, loving it

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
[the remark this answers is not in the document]
allthesignals: Sounds good! Yeah, we can trust that Ember Data will issue the correct request when saving.

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
allthesignals: Oh okay! Yeah, just isn't necessary. Up to you.

PR #667, opened 2020-09-05
12254 Set up Land Use Project Area section

godfreyyeung:
### Summary
This PR sets up the Project Area section in the Land Use Form
Fixes [AB#12254](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12254)

### Major feature
Land Use Form -- Project Area

#### Nested radio group questions
Due to the series of cascading questions and the current pattern/design of our form and fields, this component does get rather gnarly. In order to reference the value of a previous radio group question, currently our form pattern calls for the previous radio value to be yielded within the block-form rendering of a radio-group component. The subsequent radio group must sit within the previous group's block so that it can be conditionally displayed based on the yielded value. Since there is a chain of radio groups, each depending on the value of the previous group, there is heavy nesting in this component. 

Aesthetically, the Project Area component is made a bit more gnarly by the use of the Ui::Question component to help with section and label formatting 

#### Clearing radio-group descendant/nested form fields 
The PR also introduces a pattern for clearing all fields downstream to (nested within) a radio group when the value of the radio group changed.  Each radio group is passed a simple action which clears a user-specified list of fields. JavaScript list spreading is used to sort of extend lists of dependencies. 

A potentially better solution we can explore later is to not clear the nested values from the templates, but conditionally patch form values during save/submit.

### DIFF
--- a/client/app/components/packages/landuse-form/project-area.hbs
+++ b/client/app/components/packages/landuse-form/project-area.hbs
@@ -18,6 +18,7 @@
       >
         <DcpWholeCityRadioGroup
           @options={{optionset 'landuseForm' 'dcpWholecity' 'list'}}
+          @onChange={{fn this.resetDependentFields this.dcpWholecity}}
         as |dcpWholecity|>
           {{#if (eq
             dcpWholecity
@@ -37,6 +38,7 @@
               as |DcpEntiretyboroughsRadioGroup|>
                 <DcpEntiretyboroughsRadioGroup
                   @options={{optionset 'landuseForm' 'dcpEntiretyboroughs' 'list'}}
+                  @onChange={{fn this.resetDependentFields this.dcpEntiretyboroughs}}
                 as |dcpEntiretyboroughs|>
                   {{#if (eq
                     dcpEntiretyboroughs
@@ -74,6 +76,7 @@
                       as |DcpEntiretycommunityRadioGroup|>
                         <DcpEntiretycommunityRadioGroup
                           @options={{optionset 'landuseForm' 'dcpEntiretycommunity' 'list'}}
+                          @onChange={{fn this.resetDependentFields this.dcpEntiretycommunity}}
                         as |dcpEntiretycommunity|>
                           {{#if (eq
                             dcpEntiretycommunity
@@ -112,6 +115,7 @@
                               as |DcpNotaxblockRadioGroup|>
                                 <DcpNotaxblockRadioGroup
                                   @options={{optionset 'landuseForm' 'dcpNotaxblock' 'list'}}
+                                  @onChange={{fn this.resetDependentFields this.dcpNotaxblock}}
                                 as |dcpNotaxblock|>
                                   {{#if (eq
                                     dcpNotaxblock
--- a/client/app/components/packages/landuse-form/project-area.hbs
+++ b/client/app/components/packages/landuse-form/project-area.hbs
@@ -1,0 +1,152 @@
+{{#let @form as |form|}}
+  <form.Section @title="Proposed Project Area">
+    <p>
+      The <b>Project Area</b> is the entirety of all land parcels
+      that are affected by all actions being sought.
+    </p>
+
+    <Ui::Question
+    as |dcpWholecityQ|>
+      <dcpWholecityQ.Label>
+        Do <em>all</em> Actions being sought apply to the whole City?
+      </dcpWholecityQ.Label>
+
+      <form.Field
+        @attribute="dcpWholecity"
+        @type="radio-group"
+        as |DcpWholeCityRadioGroup|
+      >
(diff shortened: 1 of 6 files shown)


client/app/components/packages/landuse-form/project-area.hbs
```diff
@@ -0,0 +1,159 @@
...
+            |dcpEntiretyboroughsQ|>
+              <dcpEntiretyboroughsQ.Label>
+                Do <em>all</em> Actions being sought apply to the entirety of one or
+                more Boroughs?
+              </dcpEntiretyboroughsQ.Label>
+
+              <form.Field
+                @attribute="dcpEntiretyboroughs"
+                @type="radio-group"
+                as |DcpEntiretyboroughsRadioGroup|
```
allthesignals, line 38: The naming here is awkward. What about `IsAllBoroughsRadioGroup`?

client/app/components/packages/landuse-form/project-area.hbs
```diff
@@ -0,0 +1,159 @@
...
+              >
+                <DcpEntiretyboroughsRadioGroup
+                  @options={{optionset 'landuseForm' 'dcpEntiretyboroughs' 'list'}}
+                  @onChange={{fn this.resetDependentFields this.dcpEntiretyboroughs}}
+                as |dcpEntiretyboroughs|>
```
allthesignals, line 43: I think making this formatting consistent would make it much easier to read this.

client/app/components/packages/landuse-form/project-area.hbs
```diff
@@ -0,0 +1,159 @@
...
+                        id={{dcpBoroughsQ.questionId}}
+                      />
+                    </Ui::Question>
+                  {{/if}}
+                  {{#if (eq
+                    dcpEntiretyboroughs
+                    (optionset 'landuseForm' 'dcpEntiretyboroughs' 'code' 'NO')
+                  )}}
+                    <Ui::Question as
+                    |dcpEntiretycommunityQ|>
```
allthesignals, line 68: Looks like you format this is a bunch of different ways. Why?

client/app/components/packages/landuse-form/project-area.js
```diff
@@ -0,0 +1,41 @@
...
+  dcpEntiretyboroughs = [
+    ...this.dcpEntiretycommunity,
+    'dcpEntiretycommunity',
+    'dcpBoroughs',
+  ];
+
+  dcpWholecity = [
+    ...this.dcpEntiretyboroughs,
+    'dcpEntiretyboroughs',
+  ];
```
allthesignals, line 29: Interesting!

client/app/components/packages/landuse-form/project-area.js
```diff
@@ -0,0 +1,41 @@
...
+
+  get landuseForm() {
+    return this.args.form.data;
+  }
+
+  @action
+  resetDependentFields(fields) {
+    fields.forEach((field) => {
+      this.landuseForm[field] = null;
```
allthesignals, line 38: Is there a way to reset the changeset? I'm guessing we don't expose access to changeset methods, but this might be relevant:

https://github.com/poteto/ember-changeset#rollbackproperty

allthesignals (approved):
Looks good, had some ideas on dealing with resetting the dependent fields.

What ideas do you have on cleaning up the complexity in project-area.hbs? What are some easy wins we can do?

client/app/components/packages/landuse-form/project-area.hbs
godfreyyeung: Hmm, I wanted to stick with the format of `<field_name>RadioGroup`, instead of creating a new alias, so that the component names are predictable for developers.

client/app/components/packages/landuse-form/project-area.hbs
godfreyyeung: you're right; I will reformat and make sure it's consistent!

client/app/components/packages/landuse-form/project-area.hbs
godfreyyeung: will reformat!

client/app/components/packages/landuse-form/project-area.js
godfreyyeung: oh yeah, I honestly forgot that `form.data`  is a changeset, maybe because since it's so well abstracted.
Great idea to use the changeset api. I think using `.set()` here would be good. 
Looks like "rollback" will just roll it back to the previous value instead of clearing the value. 
That will have unintended consequences. 

For example,
  1. saved value of an input field is "Douglas street"
  2. user changes it to "Court street"
  3. user toggles a parent radio group
  4. the input field changes back to "Douglas street"

PR #668, opened 2020-09-06
12257: create environmental review component on landuse-form

trbmcginnis:
**Big Picture Summary**
User can edit Environmental Review information on the landuse form.

The landuse form

This PR adds a new component that allows users to enter information on the landuse form for the Environmental Review section. The fields in this section are on the `dcp_landuse` entity (the `landuse-form` model)

Closes [AB#12257](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12257)

**NOTE**: This implementation differs from the wireframes, two input boxes were left out
- `dcp_leadagency` was originally assumed to be a direct string field on the `dcp_landuse` entity, but it requires an `expand` and includes an entire object of data. In CRM the input for `dcp_leadagency` on the landuse form is a lookup, not a text input as originally noted in the wireframe. This will be handled in a separate PR since it will require updating wireframes and an updated technical approach ([AB#12382](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12382))
- We currently do not have a standardized component for inputting date information. The date field, `dcp_determinationdate` was left out of this PR and will be implemented after a standardized date component has been created. ([AB#12383](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12383))

### DIFF
--- a/client/app/optionsets/landuse-form.js
+++ b/client/app/optionsets/landuse-form.js
@@ -1,0 +1,14 @@
+export const CEQR_TYPE = {
+  TYPE_I: {
+    code: 717170000,
+    label: 'Type I',
+  },
+  TYPE_II: {
+    code: 717170001,
+    label: 'Type II',
+  },
+  UNLISTED: {
+    code: 717170002,
+    label: 'Unlisted',
+  },
+};
--- a/client/app/components/packages/landuse-form/edit.hbs
+++ b/client/app/components/packages/landuse-form/edit.hbs
@@ -37,6 +37,10 @@
         @form={{saveableForm}}
       />
 
+      <Packages::LanduseForm::EnvironmentalReview
+        @form={{saveableForm}}
+      />
+
       <Packages::RelatedAction
         @form={{saveableForm}}
         @addRelatedAction={{this.addRelatedAction}}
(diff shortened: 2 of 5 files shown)


client/app/optionsets/landuse-form.js
```diff
@@ -0,0 +1,14 @@
+export const CEQR_TYPE = {
```
godfreyyeung, line 1: hmm, small nitpick but.. conceptually from a CRM/business process perspective, CEQR_TYPE may not be scoped to the form. In CRM we set CEQR_TYPE on the project, (through the actions tab)  right?

godfreyyeung (approved):

client/app/optionsets/landuse-form.js
trbmcginnis: I just based it off of the `dcp_landuse` entity field `dcp_ceqrtype`. I want to leave this here since so far it's the only time we're using a ceqr type field in our forms. Maybe in the future if we end up incorporating another ceqr type field we can rearrange these optionsets? I don't want to get too far ahead of ourselves / try to implement something that's currently outside the scope of what we're doing.

PR #669, opened 2020-09-07
12249: user can edit project name on landuse form

trbmcginnis:
**Big Picture Summary**
User can edit the associated project name on the landuse form.

Land Use Form

A new component was created `packages/landuse-form/project-information` which is rendered on the landuse form. This component handles renaming a project and PATCHes to the `dcp_project` entity (the `project` model). 

Closes [AB#12249](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12249)

### DIFF
--- a/client/app/models/landuse-form.js
+++ b/client/app/models/landuse-form.js
@@ -84,6 +84,7 @@ export default class LanduseFormModel extends Model {
     await this.saveDirtyRelatedActions();
     await this.saveDirtyApplicants();
     await this.saveDirtyBbls();
+    await this.saveDirtyProject();
     await super.save();
   }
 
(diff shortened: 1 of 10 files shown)


client/app/models/landuse-form.js
```diff
@@ -111,6 +112,12 @@ export default class LanduseFormModel extends Model {
     );
   }
 
+  async saveDirtyProject() {
+    if (this.isProjectDirty) {
+      this.package.project.save();
+    }
+  }
```
allthesignals, line 119: for later: we have a lot of shared behavior across form models. maybe we can make an abstract form model, and these forms just extend from it?

`class LandUseForm extends Form {}`

trbmcginnis: yes, good point!

godfreyyeung (approved):
Looks okay to me!

PR #676, opened 2020-09-09
Merge develop into master

allthesignals:
This is a placeholder PR for merging work in develop into master branch.

Likely, we only want a subset of develop merged into master. We need to confirm how much we can put into production.


allthesignals:
Should not go straight from develop to master

PR #683, opened 2020-09-09
12255 proposed development site

godfreyyeung:
### Summary
This sets up the Proposed Development Site section for the Land Use form

Land use - proposed development site
Fixes [AB#12255](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12255)

### Technical Details
The component design is similar to that for Project Area. 
We use component nesting to implement conditional/cascading questions.

### DIFF
--- a/client/app/components/packages/landuse-form/proposed-development-site.hbs
+++ b/client/app/components/packages/landuse-form/proposed-development-site.hbs
@@ -1,0 +1,87 @@
+{{#let @form as |form|}}
+  <form.Section @title="Proposed Development Site">
+    <p>
+      The <b>Development Site</b> is the specific parcel(s) that the Applicant is seeking to develop.
+      A Project Area and Development Site can be the same parcels of land or different,
+      depending on the Actions being sought. For instance, a Special District may be mapped
+      over a portion of a neighborhood (Project Area), but only certain parcels within it may
+      be proposed for immediate development by the Applicant (Development Site).
+    </p>
+
+    <Ui::Question
+    as |dcp500kpluszoneQ|>
+      <dcp500kpluszoneQ.Label>
+        Does the Application result in the development of 500,000+ ZSF of floor area?
+      </dcp500kpluszoneQ.Label>
+
+      <form.Field
+        @attribute="dcp500kpluszone"
+        @type="radio-group"
+      as |Dcp500kpluszoneRadioGroup|>
+        <Dcp500kpluszoneRadioGroup
+          @options={{optionset 'landuseForm' 'dcp500kpluszone' 'list'}}
+          @onChange={{fn (mut @form.data.dcpDevsize) null}}
+        as |dcp500kpluszone|>
+          {{#if (eq
+            dcp500kpluszone
+            (optionset 'landuseForm' 'dcp500kpluszone' 'code' 'YES')
+          )}}
+            <Ui::Question
+              @required={{true}}
+            as |dcpDevsizeQ|>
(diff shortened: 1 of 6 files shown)


client/app/components/packages/landuse-form/proposed-development-site.hbs
```diff
@@ -0,0 +1,87 @@
...
+                @attribute="dcpDevsize"
+                @type="radio-group"
+              as |DcpDevsizeRadioGroup|>
+                <DcpDevsizeRadioGroup
+                  @options={{optionset 'landuseForm' 'dcpDevsize' 'list'}}
+                />
+              </form.Field>
+            </Ui::Question>
+          {{/if}}
+        </Dcp500kpluszoneRadioGroup>
```
allthesignals, line 46: Interesting! Keep an eye out for ways to improve this API. I'm not sure I'm a fan of the weird contextual component naming (`<Dcp500kpluszoneRadioGroup />`), but can't think of an alternative for now.

Do we need to name it as such? Or it's the linting errors you were seeing, about "shadow"s? I imagined this should just be `as |radioGroup|`. But I can see that you have another nested radio group: `DcpDevsizeRadioGroup`. What about some naming in between generic and specific?

```handlebars
<RadioGroup> 
  <ConditionalRadioGroup>
  </ConditionalRadioGroup>
</RadioGroup>
```

Or, maybe this needs to be an API change — having `<Field>` component yield out a full component as its block doesn't allow for specific naming. Maybe it should be:

```handlebars
<form.Field
  @attribute="dcp500kpluszone"
  @type="radio-group" as |dcp500kpluszoneField|
>
  <dcp500kpluszoneField.RadioGroup>
  </dcp500kpluszoneField.RadioGroup>
</form.Field>
```

This gets around the shadowing issue. It's more verbose, but it's also clearer that it's dealing with a specific field as well as a specific named component.

What do you think?

allthesignals (approved):

client/app/components/packages/landuse-form/proposed-development-site.hbs
godfreyyeung: The shadowing issue I actually took care of just by making sure that I capitalized component calls 
(`<Dcp500kpluszoneRadioGroup />` instead of `<dcp500kpluszoneRadioGroup />`)

The primary reason I thought of using these specific names is that I hoped to make it easier to distinguish between the multitude of radio groups on these forms components. Some may be nested within one another, some may sit adjacent to others, but ultimately there's a lot to keep track of. 

I think especially for radio groups which are nested, it would be hard to track which RadioGroup is being referred to without a bit more specific naming:
```
<form.Field as |radioGroup|>
          <form.Field as |radioGroup|>
                       <form.Field as |radioGroup|>
                                 // at this point, hard to track which field radioGroup references 
```
or
```
<RadioGroup> 
  <ConditionalRadioGroup>
        <ConditionalRadioGroup>
                 <ConditionalRadioGroup>
                        // might be hard to track which field conditionalRadioGroup references? 
                </ConditionalRadioGroup>
         <ConditionalRadioGroup>
  </ConditionalRadioGroup>
</RadioGroup>
```

>  <dcp500kpluszoneField.RadioGroup>
This idea is interesting.  One benefit is it would standardize our calls to `form.Field`. Right now only radio groups use the block form/yield, but if we redesign it this way, then all input types would require the block form/yield as well.
I think it's worth a shot -- I can create a task and file it in the backlog?

PR #684, opened 2020-09-09
12354: fix bug where deleted records were not being saved

trbmcginnis:
**Big Picture Summary**
Previously, users had to click save twice in order to properly DELETE `relatedActions` and `applicants`. 

PAS form and Land Use Form -- the applicants and related actions sections

A while back, we switched from `destroyRecord` to `deleteRecord`. Since then, when we looped through the records with dirty attributes and only saved those records, we were not taking into account records that had been deleted -- these records did not show up in the array and were therefore not saved. This PR passes the records that are currently being "removed" to the package model to be saved. 

Closes [AB#12354](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12354)

### DIFF
--- a/client/app/components/packages/landuse-form/edit.js
+++ b/client/app/components/packages/landuse-form/edit.js
@@ -37,10 +37,15 @@ export default class LandUseFormComponent extends Component {
     return this.package.landuseForm || {};
   }
 
+  get packageIsDirtyOrRecordsDeleted() {
+    return this.package.isDirty || this.recordsToDelete.length > 0;
+  }
+
   @action
   async savePackage() {
     try {
       await this.args.package.save(this.recordsToDelete);
+      this.recordsToDelete = [];
     } catch (error) {
       console.log('Save Landuse package error:', error);
     }
@@ -88,5 +93,12 @@ export default class LandUseFormComponent extends Component {
     this.recordsToDelete.push(relatedAction);
 
     relatedAction.deleteRecord();
+  }
+
+  @action
+  removeBbl(bbl) {
+    bbl.deleteRecord();
+    this.recordsToDelete.push(bbl);
+    this.args.package.landuseForm.bbls.removeObject(bbl);
   }
 }
--- a/client/app/components/packages/landuse-form/edit.js
+++ b/client/app/components/packages/landuse-form/edit.js
@@ -1,6 +1,7 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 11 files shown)


client/app/components/packages/landuse-form/edit.js
```diff
@@ -80,6 +85,8 @@ export default class LandUseFormComponent extends Component {
   removeRelatedAction(relatedAction, changeset) {
     removeFromHasMany(changeset, 'relatedActions', relatedAction);
 
+    this.recordsToDelete.push(relatedAction);
+
     relatedAction.deleteRecord();
   }
 }
```
godfreyyeung, line 92: Wondering.... Do we also need to push removed BBL/taxlots to recordsToDelete?
Like within the "removeBbl" action within the ProjectGeography component?

image

godfreyyeung:
> **Technical Explanation — How does it work?**
> A while back, we switched from `destroyRecord` to `deleteRecord`. Since then, when we looped through the records with dirty attributes and only saved those records, we were not taking into account records that had been deleted -- these records did not show up in the array and were therefore not saved. This PR passes the records that are currently being "removed" to the package model to be saved.
> 

Great explanation! Super helpful. Thanks so much! 
Thanks for all the sleuthing too!

client/app/components/packages/landuse-form/edit.js
trbmcginnis: > Wondering.... Do we also need to push removed BBL/taxlots to recordsToDelete?
> Like within the "removeBbl" action within the ProjectGeography component?
> 

Good point! I just missed that, going to check on that/implement that now

client/app/components/packages/project-geography.hbs
```diff
@@ -84,7 +84,7 @@
         <button
           type="button"
           class="delete-fieldset"
-          {{on "click" (fn this.removeBbl bbl)}}
+          {{on "click" (fn @removeBbl bbl)}}
```
godfreyyeung, line 87: ❤️

client/app/components/packages/landuse-form/edit.js
```diff
@@ -89,4 +94,11 @@ export default class LandUseFormComponent extends Component {
 
     relatedAction.deleteRecord();
   }
+
+  @action
+  removeBbl(bbl) {
+    bbl.deleteRecord();
```
godfreyyeung, line 100: ❤️

godfreyyeung (approved):
Looks great to me!

PR #687, opened 2020-09-10
Add team member model; extract data from CRM; expands on proejct service

allthesignals:
# Big Picture Summary
This PR shows the DCP Team in the project page contact management section

Contact Management

It wasn't displaying the DCP team members and this PR displays them in their full glory according to the REQS

It makes another request to the CRM API to add the information about DCP team members alongside the project applicants (think of these entities as siblings). It then does some idiosyncratic munging of that data (sorry, but some of the data is on a separate related entity).

This PR is "pretty bad" in terms of our code standards. There isn't any additional test coverage for this part of contact mgmt (although AFAICT there wasn't any for that feature to begin with), and this makes several assumptions about the data that's being pulled (it's not clear if we're using the correct phone number and e-mail information as the dcp_dcpteammember entity does not have an associated contact record — instead it has a "dcp_user", which is mentioned in the docs)

Closes [AB#12206](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12206)

### DIFF
--- a/client/app/templates/project.hbs
+++ b/client/app/templates/project.hbs
@@ -159,6 +159,31 @@
 
     <Messages::Assistance class="large-margin-top" />
 
+    <h3>
+      City Planning Staff
+    </h3>
+    <ul class="no-bullet">
+      {{#each this.project.teamMembers as |teamMember|}}
+        <li class="grid-x medium-margin-bottom">
+          <div class="cell shrink small-padding-right">
+            <FaIcon @icon='user' @fixedWidth={{true}} @transform='down-1' class="text-gray" />
+          </div>
+          <div class="cell auto">
+            <h5 class="no-margin">{{teamMember.name}}</h5>
+            <p class="text-small tiny-margin-bottom">
+              <i>{{teamMember.role}}</i>
+            </p>
+            <p class="text-small tiny-margin-bottom">
+              <a href="mailto:{{teamMember.email}}">{{teamMember.email}}</a>
+            </p>
+            <p class="text-small tiny-margin-bottom">
+              {{teamMember.phone}}
+            </p>
+          </div>
+        </li>
+      {{/each}}
+    </ul>
+
   </div>
 </div>
 
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -23,6 +23,11 @@ const PACKAGE_STATUSCODE = {
   REVIEWED_REVISION_REQUIRED: 717170010,
 }
 
+const DCP_PROJECTROLES = {
+  LEAD_PLANNER: 717170026,
+  BOROUGH_TEAM_LEADER: 717170000,
+};
+
 @Injectable()
 export class ProjectsService {
   constructor(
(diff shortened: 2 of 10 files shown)


server/src/projects/projects.service.ts
```diff
@@ -126,10 +148,19 @@ export class ProjectsService {
       const projectWithContacts = {
         ...project,
         'project-applicants': projectApplicantsWithContacts,
+        'team-members': projectTeamMembers.map(member => ({
+          dcp_dcpprojectteamid: member.dcp_dcpprojectteamid,
+          name: member.dcp_name,
```
trbmcginnis, line 153: How do we feel about renaming like this? Just because I know we like to keep field names as close to the names in CRM as possible -- maybe we should keep it as `dcp_name`?

trbmcginnis (approved):
Looks good!

client/app/templates/project.hbs
```diff
@@ -159,6 +159,39 @@
 
     <Messages::Assistance class="large-margin-top" />
 
+    <h3>
+      City Planning Staff
+    </h3>
+    <ul class="no-bullet">
+      {{#each this.project.teamMembers as |teamMember|}}
+        <li class="grid-x medium-margin-bottom">
```
trbmcginnis, line 167: Should we put this in its own component?

server/src/projects/projects.service.ts
godfreyyeung: Really good point and goot catch here, taylor. I think dcp_name make sense here

client/app/templates/project.hbs
```diff
@@ -159,6 +159,39 @@
...
+              <a href="mailto:{{teamMember.email}}">{{teamMember.email}}</a>
+            </p>
+            <p class="text-small tiny-margin-bottom">
+              {{teamMember.phone}}
+            </p>
+          </div>
+        
+          {{#if @canDelete}}
+          <div class="cell shrink medium-padding-left">
+            <button type="button" class="text-red-dim" {{on "click" this.tryRemoveEditor}}>
```
trbmcginnis, line 186: I don't think `tryRemoveEditor` is an action on the project.js controller

godfreyyeung (approved):
Actually, to me the code quality looks good and the PR looks pretty straightforward/following our current patterns. Maybe the CRM architecture of associating team members to users is messy, but that's not indicative of this PR

client/app/templates/project.hbs
godfreyyeung: 👏  Another great catch. Same for `@ canDelete`

server/src/projects/projects.service.ts
allthesignals: Totally. Renaming dcp_dcpprojectteam to team-members is actually okay as far as how we handle other entity relationships. Agree with the renaming. I'm going to get this in then refactor afterward. Thanks for pointing that out.

client/app/templates/project.hbs
allthesignals: Yes!

client/app/templates/project.hbs
allthesignals: Whoaaaa my b

PR #690, opened 2020-09-10
12413 Conditionally show LU dev site and tax lots sections

godfreyyeung:
### Summary 
Part of the Land use form is shown only when the Project applies to a partial area, and not to entire the city/borough/community or lot (i.e. the four respective LU fields relating to project area all evaluate to "false")

image

Fixes [AB#12413](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12413)

Form inputs are updating the changeset held at `saveableForm.data`, which is made available through the `saveableForm` yield in the `landuse-form/edit` component. Thus to access the responses to the city/borough/community/lot radio questions held in `saveableForm.data`, the conditional business logic for this PR is implemented in `landuse-form/edit.hbs` through template truth helpers (In contrast to through a controller or model file)

**- A subsequent PR will make sure that the last three fields in the Project Area section are similarly conditionally displayed**

### DIFF
--- a/client/app/components/packages/landuse-form/edit.hbs
+++ b/client/app/components/packages/landuse-form/edit.hbs
@@ -26,14 +26,6 @@
         @form={{saveableForm}}
       />
 
-      <Packages::LanduseForm::ProjectArea
-        @form={{saveableForm}}
-      />
-
-      <Packages::LanduseForm::ProposedDevelopmentSite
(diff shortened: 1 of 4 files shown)


allthesignals (approved):
Nice!

PR #698, opened 2020-09-11
Rename method

allthesignals:
This wraps the lookup for whether a user is part of a package project applicant team into a named, declarative method.

Thoughts?


godfreyyeung (approved):
Works for me! Self-documenting code ❤️  :) 
Yeah previously there were no comments, so someone might be confused what exactly is the point of the web query

PR #703, opened 2020-09-14
Auth Middleware should throw unauthenticated

allthesignals:
@godfreyyeung IDK how much I Like this.

This change makes the auth middleware throw a 401 if the validation step fails, which makes sense.

The problem was that the packages guard was throwing a 403 when the session didn't have a contactId. This is the wrong error code. It should have been throwing a 401 earlier in the lifeceycle (either middleware or a pipe).


godfreyyeung (approved):
Ohh, good catch. Then should we also change to 401 for the HttpException in the package guard as well? Or just get rid of it altogether? Although if we get rid of it, we have to assume that the Auth Middleware always precedes the guard?

PR #705, opened 2020-09-14
💅 Improve sign in email detector

andycochran:
This PR…
- Use an action on a form element to `startLogin()`
- Deletes the actions on the input and button
- Changes the button to a submit (this way a button click or an input enter will trigger the action)


allthesignals (approved):
nice! look at that native DOM event being useddddd

PR #710, opened 2020-09-16
Display applicant team phone number

allthesignals:
Addresses [AB#12453](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12453)

### DIFF
--- a/client/app/helpers/format-phone.js
+++ b/client/app/helpers/format-phone.js
@@ -1,0 +1,13 @@
+import { helper } from '@ember/component/helper';
+
+export function formatPhone(params) {
+  const [phone] = params;
+
+  if (!phone) {
+    return phone;
+  }
+
+  return String(phone).replace(/(\d{3})(\d{3})(\d{4})/, '($1) $2-$3');
+}
+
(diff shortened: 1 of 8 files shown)


client/app/helpers/format-phone.js
```diff
@@ -0,0 +1,13 @@
+import { helper } from '@ember/component/helper';
+
+export function formatPhone(params) {
+  const [phone] = params;
+
+  if (!phone) {
+    return phone;
```
godfreyyeung, line 7: Wait, why return phone if it's undefined or null?

godfreyyeung (approved):

client/app/helpers/format-phone.js
godfreyyeung: Oh, is this just to prevent undefined/null from going through the String replace function?

allthesignals: falsey-ness isn't just undefined null — a blank string is also falsey.

PR #719, opened 2020-09-17
12404 Land Use Subject Sites section 

godfreyyeung:
### Summary 
Sets up creating, deleting, and loading Subject Sites

Fixes [AB#12404](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12404)

Technical details:
- Each subject site is a record on a new entity called dcp_sitedatahform
- There is a hasMany relationship from dcp_landuse to this new entity
- Mostly follows the same patterns as for previous hasMany 
- Some differences include:
   - The navigation property from dcp_sitedatahform to dcp_landuse uses the term "landuseform" instead of just "landuse"
   - We needed to introduce multiple get requests to load the a land use form, since the land use form now has over five expands. Five is the limit of expands for a single get request.


client/app/routes/landuse-form.js
```diff
@@ -4,10 +4,19 @@ import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-rout
...
+    'landuse-form.related-actions',
+    'landuse-form.landuse-actions',
+    'landuse-form.sitedatah-forms',
+  ];
+
   async model(params) {
     const landuseFormPackage = await this.store.findRecord('package', params.id, {
       reload: true,
-      include: 'landuse-form.bbls,landuse-form.applicants,project,landuse-form.related-actions,landuse-form.landuse-actions',
+      include: this.packageSideloads.join(),
```
godfreyyeung, line 19: The line was getting too long and unwieldy, so wanted to break it out vertically.

client/app/components/packages/landuse-form/edit.js
```diff
@@ -100,6 +102,24 @@ export default class LandUseFormComponent extends Component {
...
+    });
+
+    addToHasMany(changeset, 'sitedatahForms', newSitedatahForm);
+  }
+
+  @action
+  removeSitedatahForm(sitedatahForm, changeset) {
+    removeFromHasMany(changeset, 'sitedatahForms', sitedatahForm);
+
+    this.recordsToDelete.push(sitedatahForm);
```
allthesignals, line 118: This feels wrong. It seems like it's done to trigger an observer somewhere. Not a big deal because we probably do it elsewhere but should find a better way soon.

client/app/routes/landuse-form.js
allthesignals: cool! maybe just in-line the array.

allthesignals (approved):
Looks good! Thanks for putting this together. 

"sitedatah-form" is such an unfortunate name!

godfreyyeung:
> Looks good! Thanks for putting this together.
> 
> "sitedatah-form" is such an unfortunate name!

Tell me about it 🤦‍♂️  

> maybe just in-line the array.
Oh right! Didn't think of that...

client/app/components/packages/landuse-form/edit.js
godfreyyeung: I agree. One reason for this design currently is to "queue" up the records for actual deletion in the backend. 
However I think it may also be contributing to a current bug with the state of the save button (the save button isn't being appropriately disabled when sitedatahForm are removed)

PR #720, opened 2020-09-17
Payment link gen prototype

allthesignals:
New CityPay module — creates a new service, which is used in the packages controller.

The service generates a citypay link for a given package id.


server/src/citypay/citypay.service.ts
```diff
@@ -0,0 +1,112 @@
...
+      if (retries) {
+        return this.generateCartKey(packageId, retries - 1);
+      } else {
+        throw new Error(`Could not generate the key for some reason: ${e.toString}`);
+      }
+    }
+  }
+
+  // generates cookies by headless login to MS Applicant Portal
+  private async stealCookies(retries = MAX_RETRIES) {
```
godfreyyeung, line 72: stealCookies isn't the clearest name... Maybe something like "getMsPortalCookies"

server/src/citypay/citypay.service.ts
```diff
@@ -0,0 +1,112 @@
+import { Injectable } from '@nestjs/common';
+import * as puppeteer from 'puppeteer';
+import * as superagent from 'superagent';
+import { ConfigService } from '../config/config.service';
+
+const MAX_RETRIES = 5;
+const MS_APPLICANT_PORTAL = {
+  COOKIE_NAME: '.AspNet.ApplicationCookie',
+  SIGN_IN_PAGE: 'https://nycdcppfs.dynamics365portals.us/SignIn',
```
godfreyyeung, line 9: We should place these in environment variables

server/src/citypay/citypay.service.ts
```diff
@@ -0,0 +1,112 @@
...
+    console.log('getting cart key...');
+
+    try {
+      const cookies = await this.stealCookies();
+      const microsoftPortalCookie = cookies.find(cookie => cookie.name === MS_APPLICANT_PORTAL.COOKIE_NAME)
+
+      // fetch the JWT to sign the next req
+      const MSPortalJWTRequest = await superagent
+        .get(MS_APPLICANT_PORTAL.TOKEN_API)
+        .query({ client_id: '29f7ab7e-e9bb-4b20-9b8f-4760a8313b28' })
```
godfreyyeung, line 46: Looks like we should environment variable this, too

godfreyyeung (approved):

PR #721, opened 2020-09-18
Use NYC.ID oauth user endpoint; create synchronize method

allthesignals:
Addresses [AB#12365](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12365) 

This PR makes sure that the first and last names of CRM contacts are synced to the respective values in NYCID, after a user has edited their profile on NYCID. 

Includes a new client auth route, "sync", which is the redirect location after a user updates the profile on NYCID. 
The route infers that the user has updated their profile if the browser "referrer" is the NYC.ID user profile edit page. If so, it triggers new backend methods to request the updated user first and last name from NYCID, and then send the info to CRM.

The sync route does so by triggering the save() method on the contact model. The save() method hits a new backend Contact PATCH endpoint, which eventually calls a new backend method `getNycidOAuthUser`.  That method calling the OAuth-based "user" endpoint, which returns an updated user profile. It requires the user's accessToken to authenticate the request. So, we include this with the session information, which itself is encrypted.

After acquiring the updated user profile, the info is sent to CRM via the CRM Web API.

### DIFF
--- a/client/app/authenticators/zap-api-authenticator.js
+++ b/client/app/authenticators/zap-api-authenticator.js
@@ -17,15 +17,12 @@ export default class ZAPAuthenticator extends OAuth2ImplicitGrantAuthenticator {
     }
 
     let attemptedLoginEmail = '';
+    const NYCIDUser = jwtDecode(accessToken);
 
     // attempt to decode the token to include email address for debugging
-    try {
-      const { mail } = jwtDecode(accessToken);
+    const { mail } = NYCIDUser;
 
-      attemptedLoginEmail = mail;
-    } catch (e) {
-      console.log('Failed to decode the JWT', e);
-    }
+    attemptedLoginEmail = mail;
 
     // Pass the NYCIDToken to backend /login endpoint.
     // Server should provide an access_token used for signing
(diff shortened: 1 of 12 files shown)


client/app/authenticators/zap-api-authenticator.js
```diff
@@ -17,15 +17,12 @@ export default class ZAPAuthenticator extends OAuth2ImplicitGrantAuthenticator {
     }
 
     let attemptedLoginEmail = '';
+    const NYCIDUser = jwtDecode(accessToken);
 
     // attempt to decode the token to include email address for debugging
-    try {
-      const { mail } = jwtDecode(accessToken);
+    const { mail } = NYCIDUser;
```
godfreyyeung, line 23: Why remove the try/catch?

godfreyyeung (approved):
I updated the description based on my understanding; see if it makes sense. Also it looks like there's a lot of extra refactoring -- would be nice if they were broken out into a separate PR for clarity.

client/app/authenticators/zap-api-authenticator.js
allthesignals: i had it there because it was for debugging purposes and didn't want it to break anything. now we need it

PR #724, opened 2020-09-21
Contact MGMT improvements

andycochran:
This PR… 

- Adds a config var to enable the Contact MGMT features (`true` be default, `false` in production environment) 
- Improves the UI of the Contact MGMT section by 
  - Associating the "Manage/Cancel" button w/ the section header (it seemed like part of the "Add Editor" form)
  - Replaces the user icon with an X if `@canDelete` (so it's more direct association w/ user instead of aligned to the far right)
  - Adds tooltips that show when you enable editing to highlight what's editable
  - Condenses the "Add Editor" form 
  - Uses the `<Ui::GenericModal>` instead of the confirmation modal
- Improves the spacing and dividing rules of sections of the page (e.g. horizontal line between packages, appropriate spacing between sidebar sections)

### DIFF
--- a/client/app/components/project/project-editor-list-item.hbs
+++ b/client/app/components/project/project-editor-list-item.hbs
@@ -1,6 +1,17 @@
 <li class="grid-x medium-margin-bottom">
   <div class="cell shrink small-padding-right">
-    <FaIcon @icon='user' @fixedWidth={{true}} @transform='down-1' class="text-gray" />
+    {{#if @canDelete}}
+      <button
+        type="button"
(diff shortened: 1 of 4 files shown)


godfreyyeung (approved):
Thanks for the great writeup! The UI changes look okay to me!

In terms of the environment variable, it sounds like we'll have to re-enable it later when it's "ready". 
Like you mentioned earlier, I guess the right way would have been to only merge this when it was ready (at least until we get a tagging system in place).

What are the downsides to displaying the contact management features to users now?

PR #725, opened 2020-09-22
12391 public facilities

godfreyyeung:
### Summary 
Sets up Public Facilities section
Fixes [AB#12391](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12391)

### Technical Details
- Introduces the YES_NO_INTEGER optionset to accommodate the unfortunate Yes/No options which map to 1 and 0, instead of true and false. 
- Uses native Ember Input component to render checkboxes tied to integer values, since our Checkbox component assumes boolean values.
- Introduces the `data-test-checked` and `aria-checked` attributes to Radio components so that we can assert their checked state in tests
- Re-use the pattern for clearing dependent fields introduced in the housing-plans section

### DIFF
--- a/client/app/components/packages/landuse-form/public-facilities.hbs
+++ b/client/app/components/packages/landuse-form/public-facilities.hbs
@@ -1,0 +1,340 @@
+{{#let @form as |form|}}
+  {{#let (intersect
+    (map-by "dcpActioncode" form.data.data.landuseActions)
+    (array 'PC' 'PQ' 'PS' 'PX')
+  ) as |publicFacilitiesActions|}}
+    {{#if (gt publicFacilitiesActions.length 0)}}
+      <form.Section @title="Public Facilities">
+        <p>
+          The following questions relate to Office Space Lease,
+          Public Facility Site Selection or Acquisition and the
+          following Actions: PC, PQ, PS, PX. These Actions require
+          additional Attachments as noted in the Attachments section.
+        </p>
+
+        <Ui::Question
+        as |proposedActionsQ|>
+          <proposedActionsQ.Label>
+            Which action(s) is/are being proposed? (select all that apply)
+          </proposedActionsQ.Label>
+
+          <Input
+            type="checkbox"
+            id="dcpOfficespaceleaseopt"
+            data-test-checkbox="dcpOfficespaceleaseopt"
+            checked={{if
+              (eq
+                form.data.dcpOfficespaceleaseopt
+                (optionset 'landuseForm' 'dcpOfficespaceleaseopt' 'code' 'Yes')
+              ) "checked"}}
+            onClick={{fn (mut form.data.dcpOfficespaceleaseopt)
+              (if form.data.dcpOfficespaceleaseopt
+                (optionset 'landuseForm' 'dcpOfficespaceleaseopt' 'code' 'No')
+                (optionset 'landuseForm' 'dcpOfficespaceleaseopt' 'code' 'Yes')
+              )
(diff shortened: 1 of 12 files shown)


andycochran (approved):

client/app/components/packages/landuse-form/public-facilities.hbs
```diff
@@ -0,0 +1,340 @@
+{{#let @form as |form|}}
+  {{#let (intersect
+    (map-by "dcpActioncode" form.data.data.landuseActions)
+    (array 'PC' 'PQ' 'PS' 'PX')
```
trbmcginnis, line 4: Oh nifty -- So does this only display this content if `form.data.data.landuseActions` includes at least one of the four options in this array of action codes?

client/app/components/packages/landuse-form/public-facilities.hbs
```diff
@@ -0,0 +1,340 @@
...
+            Will the existing facility remain?
+          </dcpExistingfacilityproposedtoremainoptQ.Label>
+
+          <form.Field
+            @attribute="dcpExistingfacilityproposedtoremainopt"
+            @type="radio-group"
+          as |DcpExistingfacilityproposedtoremainoptRadioGroup|>
+            <DcpExistingfacilityproposedtoremainoptRadioGroup
+              @options={{optionset 'landuseForm' 'dcpExistingfacilityproposedtoremainopt' 'list'}}
+              @onChange={{fn this.resetDependentFields this.dependantsOf.dcpExistingfacilityproposedtoremainopt}}
```
trbmcginnis, line 114: interesting!

client/app/components/packages/landuse-form/public-facilities.hbs
godfreyyeung: Yep, exactly! I think if this was any longer or messier I would definitely move this into the component class (js) file..

client/app/components/packages/landuse-form/public-facilities.hbs
godfreyyeung: Yep! This is to clear values of nested answers when this radio answer changes

client/app/components/packages/landuse-form/public-facilities.hbs
```diff
@@ -0,0 +1,340 @@
...
+
+          <Input
+            type="checkbox"
+            id="dcpOfficespaceleaseopt"
+            data-test-checkbox="dcpOfficespaceleaseopt"
+            checked={{if
+              (eq
+                form.data.dcpOfficespaceleaseopt
+                (optionset 'landuseForm' 'dcpOfficespaceleaseopt' 'code' 'Yes')
+              ) "checked"}}
```
trbmcginnis, line 29: Aw okay cool

client/app/components/packages/landuse-form/public-facilities.hbs
trbmcginnis: Yeah I have a computed property that does the same thing but the list of required actions codes is much longer. This is a cool way to do it though! Didn't know it was possible

trbmcginnis (approved):

client/app/components/packages/landuse-form/public-facilities.hbs
godfreyyeung: It took me a whiiile to think of this 🙈  I tried to overcomplicate at first with a new helper/component

PR #746, opened 2020-09-28
Landuse form -> qa

godfreyyeung:


godfreyyeung:
image
After conversation with Matt, closing in favor of
1. merging landuse-form into develop
2. Force pushing develop to qa
(Saving local copy of qa in case ) 

We are returning to Git Flow-like workflow

PR #749, opened 2020-09-29
Adds an extra condition to ensure email exact match

allthesignals:
This adds an extra condition to ensure an “exact” but case-insensitive e-mail match before overwriting the previously-stored GUID.

# Big Picture Summary
Because users will have to re-register, they will have new GUIDs. This change makes sure that the e-mail on the NYC.ID at least also matches before overwriting the GUID (in case something goes wrong or "creeper mode" is enabled).

Auth

### DIFF
--- a/server/src/auth/auth.service.ts
+++ b/server/src/auth/auth.service.ts
@@ -126,8 +126,14 @@ export class AuthService {
       dcp_nycid_guid: GUID,
     });
 
+    const nycIdEmailRegEx = new RegExp(`^${mail}$`, 'gi');
+
     // if their e-mail is validated, associate the NYCID guid
-    if (nycExtEmailValidationFlag && contact.dcp_nycid_guid !== GUID) {
+    if (
+      nycExtEmailValidationFlag
+      && contact.dcp_nycid_guid !== GUID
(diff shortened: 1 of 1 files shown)


server/src/auth/auth.service.ts
```diff
@@ -126,8 +126,14 @@ export class AuthService {
...
     });
 
+    const nycIdEmailRegEx = new RegExp(`^${mail}$`, 'gi');
+
     // if their e-mail is validated, associate the NYCID guid
-    if (nycExtEmailValidationFlag && contact.dcp_nycid_guid !== GUID) {
+    if (
+      nycExtEmailValidationFlag
+      && contact.dcp_nycid_guid !== GUID
+      && contact.emailaddress1.match(nycIdEmailRegEx)
```
godfreyyeung, line 135: Should you add a check for `givenName` and `sn` here? 
Or actually it seems like this first update can forego updating `firstname` and `lastname`, since the update below (lines 147-148) takes care of it. 

I know this came in a previous PR, but having a bit of trouble following the logic of these two conditions (line 132 and 145)

godfreyyeung (approved):
The new regex itself looks good!

PR #752, opened 2020-09-29
12654: fix typo in POST for applicant representative

trbmcginnis:
**What was wrong and how does this fix the problem?**
User was unable to save applicants on the Other Team Member section of the applicants component on the landuse-form. This was due to a typo in the POST on the `applicants.controller.ts`, where the incorrect entity was listed. This PR fixes the typo so that users can save to the correct entity. 

Closes [AB#12654](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12654)

### DIFF
--- a/server/src/packages/pas-form/applicants/applicants.controller.ts
+++ b/server/src/packages/pas-form/applicants/applicants.controller.ts
@@ -124,7 +124,7 @@ export class ApplicantsController {
           },
         );
       } else if (body.landuse_form) {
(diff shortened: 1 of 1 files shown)


allthesignals:
Good catch, thanks... it seems to make sense to me because it's inside the `dcp_applicantrepresentativeinformation` if/else block.

godfreyyeung (approved):
Thank you! Nice catch

PR #753, opened 2020-09-30
Prefers and adds env variable for domains

allthesignals:
When deployed to any environment, Ember's internal setting is always "production."

When I switched the NYC.ID domain to prod, it caused non production environments to break because they were pulling NYC.ID guids from the non-prod environment and using those to query the prod env.

This change provies an env variable to override that domain to non-prod.

### DIFF
--- a/client/config/environment.js
+++ b/client/config/environment.js
@@ -79,7 +79,7 @@ function getHost(environment) {
 }
 
 function getOAuthDomain(environment) {
-  const { HOST } = process.env;
+  const { HOST, NYCID_DOMAIN } = process.env;
 
   // plain local development, no local backing server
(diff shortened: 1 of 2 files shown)


godfreyyeung:
> When deployed to any environment, Ember's internal setting is always "production."
> 
> When I switched the NYC.ID domain to prod, it caused non production environments to break because they were pulling NYC.ID guids from the non-prod environment and using those to query the prod env.
> 
> This change provies an env variable to override that domain to non-prod.

Do you mean pulling NYC.ID GUIDs from prod NYCID and using those to query non-prod CRM env?

godfreyyeung (approved):

PR #760, opened 2020-10-01
12401: lead-agency section on landuse form

trbmcginnis:
**Big Picture Summary**
Users can use a dropdown to search and select a lead agency (e.g. DOE) on the Environmental Review section of the Land Use Form.

Draft Land Use Form

- A **new componen**t was created called `lead-agency` that is rendered on the `environmental-review` component, which is then rendered on the landuse-form
- We have a GET for the `accounts` entity in order to **display a list of all accounts** that match the desired criteria in a dropdown. 
- When a user selects an option from the **dropdown**, the field `chosenLeadAgencyId`, which is just a made-up field and does not exist in CRM, is updated to match the id of the account selected. 
- When the user saves the landuse-form, this `chosenLeadAgencyId` is sent to the backend. And this **account id is then used to bind** the matching account to the `dcp_landuse`'s `dcp_leadagency` field. 
- We never PATCH or POST to this `dcp_leadagency` field. This field is dependent on the "bind" to the matching account. We GET information from `dcp_leadagency` in order to handle display logic on the frontend, but other than that we do not interact with the `lead-agency` model.

Closes [AB#12401](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12401)

**Next steps**: instead of a made-up field on the landuse-form, we should update the lead-agency model and then grab the id from the relationship field of the response body on the landuse-form controller

### DIFF
--- a/client/app/models/landuse-form.js
+++ b/client/app/models/landuse-form.js
@@ -22,6 +22,13 @@ export default class LanduseFormModel extends Model {
   @hasMany('landuse-geography', { async: false })
   landuseGeographies;
 
+  // this is just for GETting dcp_leadagency information
+  // we do not PATCH or POST directly to dcp_leadagency
+  // instead we handle sending related information to the backend
+  // through a made-up field called `chosenLeadAgencyId`
+  @belongsTo('lead-agency', { async: false })
+  leadAgency;
+
   @attr dcpVersion;
 
   // project name
@@ -79,8 +86,6 @@ export default class LanduseFormModel extends Model {
   @attr dcpSitedataidentifylandmark;
 
   // Environmental Review attrs
-  @attr dcpLeadagency;
-
   @attr dcpCeqrnumber;
 
   @attr dcpCeqrtype;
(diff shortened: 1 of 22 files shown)


client/app/controllers/landuse-form/edit.js
```diff
@@ -0,0 +1,4 @@
+import Controller from '@ember/controller';
+
+export default class LanduseFormEditController extends Controller {
+}
```
allthesignals, line 4: I would add an `accounts` property `= []`

client/app/routes/landuse-form.js
```diff
@@ -22,4 +23,9 @@ export default class LanduseFormRoute extends Route.extend(AuthenticatedRouteMix
 
     return landuseFormPackage;
```
allthesignals, line 24: https://api.emberjs.com/ember/release/classes/rsvp/methods

client/app/models/landuse-form.js
```diff
@@ -75,9 +82,6 @@ export default class LanduseFormModel extends Model {
 
   @attr dcpSitedataidentifylandmark;
 
-  // Environmental Review attrs
```
godfreyyeung, line 78: Should leave this comment in the file. I placed these comments here to section off these attributes

trbmcginnis: Whoops that was an accident!

PR #761, opened 2020-10-05
12473 populate section

CeciliaCY:
# Big Picture Summary
User story 12473 - After user submits LU form, they see a summary of their submission. Read only view of LU form

Land Use Form

Need to populate all the answers that were submitted via LU form edit page

1. Use the <Ui::Answer> component
2. Keep the same display logic as edit page
3. Create several new option lists


client/app/templates/ZAPPortal.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
+	"folders": [
+		{
+			"path": "/Users/chengying/Development/labs-applicant-portal"
+		}
+	]
+}
```
allthesignals, line 7: can you add this to gitignore

allthesignals:
Thanks, @CeciliaCY !!!

Just a few things:

- there were a couple of things committed that shouldn't have been: node_modules had been committed for some reason. also, a .code-workspace file was committed.
- I reverted some of the changes to the project-geography component because they seemed to break some of the tests. I think we should figure out what was going on with that in a later ticket. 
- commit messages were a little confusing — I couldn't understand why some changes were made (to optionset and bbl). but that's okay.

otherwise, code looks good! Thank you!!

PR #763, opened 2020-10-05
12394: add zoning text amendment section on landuse-form

trbmcginnis:
**Big Picture Summary**
Add Zoning Text Amendment section to the Land Use Form. This section allows users to "add" a ZR Section which then displays a form (per each ZR Section added) that has two input questions. Users can also remove these "ZR Sections"

Land Use Form

- Two components were created to handle this functionality. One was the `zoning-text-amendment-fieldset` which handles the two input questions and the "Delete ZR Section" button. This `fieldset` component was then rendered on `zoning-text-amendment.hbs` which handles the title and "adding" functionality. This component is then rendered on the landuse form. 
- This component saves fields on the `affected-zoning-resolution` model, so a `POST` and `DELETE` were added to the `affected-zoning-resolutions.controller` in the backend

Closes [AB#12394](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12394)

### DIFF
--- a/client/app/components/packages/landuse-form/zoning-text-amendment-fieldset.hbs
+++ b/client/app/components/packages/landuse-form/zoning-text-amendment-fieldset.hbs
@@ -2,20 +2,27 @@
   <Ui::Question
     class="fieldset relative"
     ...attributes
-    as |Q|
-  >
+  as |Q|>
     <Q.Legend data-test-zr-section-title>
       ZR Section
     </Q.Legend>
 
     <label>
       Zoning Section Number
-      <form.Field @attribute="dcpZrsectionnumber" />
+      <form.Field
+        @attribute="dcpZrsectionnumber"
+        @showCounter={{true}}
+        @maxlength="25"
+      />
     </label>
 
     <label>
       Zoning Section Title
-      <form.Field @attribute="dcpZrsectiontitle" />
+      <form.Field
+        @attribute="dcpZrsectiontitle"
+        @showCounter={{true}}
+        @maxlength="100"
+      />
     </label>
 
     <button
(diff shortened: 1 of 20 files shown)


client/app/components/packages/landuse-form/zoning-text-amendment-fieldset.hbs
```diff
@@ -0,0 +1,32 @@
...
+    ...attributes
+    as |Q|
+  >
+    <Q.Legend data-test-zr-section-title>
+      ZR Section
+    </Q.Legend>
+
+    <label>
+      Zoning Section Number
+      <form.Field @attribute="dcpZrsectionnumber" />
```
godfreyyeung, line 13: We need to toggle `showCounter` and set the maxlength here

client/app/components/packages/landuse-form/zoning-text-amendment-fieldset.hbs
```diff
@@ -0,0 +1,32 @@
...
+    </label>
+
+    <label>
+      Zoning Section Title
+      <form.Field @attribute="dcpZrsectiontitle" />
```
godfreyyeung, line 18: Same comment as above here

godfreyyeung (changes requested):

godfreyyeung (commented):
The rest of this looks great! Thank you Taylor!

client/app/components/packages/landuse-form/zoning-text-amendment.hbs
```diff
@@ -0,0 +1,51 @@
+{{#let @form as |form|}}
```
godfreyyeung, line 1: Oops, we also need to toggle this section based off of if there is a ZR action. I'll add

godfreyyeung (approved):

client/app/components/packages/landuse-form/zoning-text-amendment.hbs
trbmcginnis: thanks!

PR #768, opened 2020-10-08
Downgrade Labs UI

allthesignals:
This change downgrades Labs UI package to a previous version which prefers the correct version of ember-power-select.

We need to fix some of the dependency issues later down the road.


trbmcginnis:
So the upgrade to the new labs-ui version was so that our app had access to Andy's new date picker. If we downgrade this then it'll be a blocker for updating any of the date inputs. Just something to consider.

PR #769, opened 2020-10-08
12395 draft lu zoning map amendment

godfreyyeung:
### Summary
Sets up the Draft Land Use Zoning Map Amendment Section
Fixes [AB#12395](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12395) 

### Technical details (Nice-to-knows, but you don't really need to read all this):
Mostly follow the same patterns as other sections. Some nice-to-knows:
- This PR sets an example for calling *-fieldset components going forward. Notice how the zoning-map-amendment-fieldset component is called without the @zoningMapChange argument, since that argument is redundant and confusing.
- Similarly this PR sets an example for calling the `@ remove<fieldset>` actions. Previously in other sections we passed an extra, unnecessary argument to the `@ remove<fieldset>` action inside the *-fieldset file (at the remove button). 
- We should later assess whether the `ZONING_MAP_CHANGE_OPTIONSET.DCPEXISTINGZONINGDISTRICTVALUE` 
optionset is frequently growing/shrinking. If so, we should pull this optionset down on app load. I created a ticket here: [AB#12865](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12865)
- Another area we are pushing up against the "saveable form" api: the Q.Label component doesn't create working data-test hooks for power selects. Power select need a redundant `<label>` to house working test selectors. 
- In CRM, the singular entity name for a Zoning Map Change is dcp_zoningmapchanges, and plural is dcp_zoningmapchangeses. But we hide most of the app from this detail. It only affects the direct, low-level calls to the CRM Web API.

### DIFF
--- a/client/app/components/packages/landuse-form/zoning-map-amendment-fieldset.hbs
+++ b/client/app/components/packages/landuse-form/zoning-map-amendment-fieldset.hbs
@@ -1,0 +1,71 @@
+{{#let @form as |form|}}
+  <Ui::Question
+    class="fieldset relative"
+    ...attributes
+  as |Q|>
+    <Q.Legend data-test-zoning-map-change-title>
+      Zoning Map Change
+    </Q.Legend>
+
+    <Ui::Question
+    as |dcpZoningsectionmapsnumberQ|>
+      <dcpZoningsectionmapsnumberQ.Label>
+        Zoning Section Map(s) Number
+      </dcpZoningsectionmapsnumberQ.Label>
+
+      <form.Field
+        @attribute="dcpZoningsectionmapsnumber"
+        @maxlength="100"
+        @showCounter={{true}}
+        id={{dcpZoningsectionmapsnumberQ.id}}
+      />
+    </Ui::Question>
+
+
+    <Ui::Question
+    as |dcpExistingzoningdistrictvalueQ|>
+      <dcpExistingzoningdistrictvalueQ.Label
+        data-test-applicant-state-dropdown
+      >
+        Existing Zoning District(s)
+      </dcpExistingzoningdistrictvalueQ.Label>
+      
+      <label data-test-dcpExistingzoningdistrictvalue-dropdown>
+        <PowerSelect
+          supportsDataTestProperties={{true}}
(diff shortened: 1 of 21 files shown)


client/app/components/packages/landuse-form/zoning-map-amendment-fieldset.hbs
```diff
@@ -0,0 +1,71 @@
...
+  <Ui::Question
+    class="fieldset relative"
+    ...attributes
+  as |Q|>
+    <Q.Legend data-test-zoning-map-change-title>
+      Zoning Map Change
+    </Q.Legend>
+
+    <Ui::Question
+    as |dcpZoningsectionmapsnumberQ|>
```
trbmcginnis, line 11: So thourough with these names, love it! :)

client/app/components/packages/landuse-form/zoning-map-amendment.hbs
```diff
@@ -0,0 +1,83 @@
...
+        {{/each}}
+      {{/if}}
+
+      <Ui::Question
+      as |dcpTotalzoningareatoberezonedQ|>
+        <dcpTotalzoningareatoberezonedQ.Label>
+          What is the total area of all Zoning Lots in the area to the be rezoned?
+        </dcpTotalzoningareatoberezonedQ.Label>
+        
+        {{!-- Todo: refactor to avoid double label for sake of PowerSelect --}}
```
trbmcginnis, line 22: Thanks for the note

client/app/templates/landuse-form/edit.hbs
```diff
@@ -1,6 +1,6 @@
 <Ui::Breadcrumbs as |Crumb|>
   <Crumb @text="My Projects" @route='projects' />
-  <Crumb @text={{@model.project.dcpProjectname}} @disabled={{true}} />
+  <Crumb @text={{@model.package.project.dcpProjectname}} @disabled={{true}} />
```
trbmcginnis, line 3: oooh -- thanks for fixing!

PR #771, opened 2020-10-13
12672 Patch de/serialization for dcp_500kpluszone

godfreyyeung:
### Summary 

Fixes [AB#12572](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12572) (answeres to the dcp_500kpluszone question was not getting written to/read from CRM)

It turns out our current Nest de/serialization and CRM is at odds with
Ember Data in regards to whether digits appear
before for after a dash/underscore. Thus we have to implement
custom serialization/deserialization for attributes with
digit tokens.

### Technical Details (Important -- read this!)

When generating JSONAPI request payloads, Ember Data dasherizes model properties by
  1. placing dashes before any capital letters. Note this is only letters but not numbers
  2. lowercasing the capital letters

When generating models from JSONAPI response payloads, Ember Data does the reverse:
  1. capitalizing letters that immediately follow a dash
  2. remove dashes

When deserializeing JSONAPI request payloads, Nest replaces dashes with underscores.
(This appropriately generate respective CRM properties)

When serializing objects into JSONAPI request responses, Nest replaces underscores with dashes.

Thus...
   - It is not possible to rely on the "5" in 500 for proper dasherization, because Ember Data uses capital letters
to indicate placement of a dash during serialization (into a JSONAPI object). Thus when Ember Data serializes dcp500kpluszone, it simply is `dcp500kpluszone`. Contrast this to another property like dcpProjectname, which Ember Data serializes into dcp-projectname. 
  -   In our backend, without the dash the Nest deserializer cannot automatically determine where an underscore should be placed and generate the respective CRM property for dcp500kpluszone. 
  - To solve this problem, in the Deserializer we add manual deserialization (translation) of dcp500kpluszone to dcp_500kpluszone. Similarly, during Serialization we translate dcp_500kpluszone to dcp500kpluszone.
  - For all other attributes, we default to the `snake_case` implementation, which turns out to just be the `underscore()` method from the Inflector library.

### DIFF
--- a/server/src/json-api-deserialize.pipe.ts
+++ b/server/src/json-api-deserialize.pipe.ts
@@ -1,5 +1,6 @@
 import { ArgumentMetadata, Injectable, PipeTransform } from '@nestjs/common';
 import { Deserializer } from 'jsonapi-serializer';
+import { underscore } from 'inflected';
 
 @Injectable()
 export class JsonApiDeserializePipe implements PipeTransform {
@@ -8,7 +9,16 @@ export class JsonApiDeserializePipe implements PipeTransform {
 
     if (type === 'body') {
       return new Deserializer({
-        keyForAttribute: 'snake_case',
+        keyForAttribute: (attribute) => {
+          // One-off deserialization for any properties should happen here.
+          // Add another if statement.
+          // See https://github.com/NYCPlanning/labs-applicant-portal/pull/771
+          // TODO: Extract out and modularize the set of one-offs if it grows
+          // beyond 2.
+          if (attribute === 'dcp500kpluszone')
+            return 'dcp_500kpluszone';
+          return underscore(attribute);
+        },
         packages: {
           valueForRelationship: relationship => relationship.id,
         },
--- a/server/src/json-api-serialize.interceptor.ts
+++ b/server/src/json-api-serialize.interceptor.ts
@@ -16,6 +16,15 @@ export class JsonApiSerializeInterceptor implements NestInterceptor {
       // REDO: we should just make the client support dash prefixes
       // because it'll be tough when we have to deserialize
       keyForAttribute(key) {
+        // One-off serialization for any properties should happen here.
+        // Add another if statement.
+        // See https://github.com/NYCPlanning/labs-applicant-portal/pull/771
+        // TODO: Extract out and modularize the set of one-offs if it grows
+        // beyond 2.
+        if (key === 'dcp_500kpluszone') {
+          return 'dcp500kpluszone';
+        }
+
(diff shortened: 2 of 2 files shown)


server/src/json-api-deserialize.pipe.ts
```diff
@@ -8,7 +9,11 @@ export class JsonApiDeserializePipe implements PipeTransform {
 
     if (type === 'body') {
       return new Deserializer({
-        keyForAttribute: 'snake_case',
+        keyForAttribute: (attribute) => {
+          if (attribute === 'dcp500-kpluszone')
+            return 'dcp_500kpluszone';
+          return underscore(attribute);
```
godfreyyeung, line 15: I looked into the source for jsonapi-serializer to confirm that snake_case defers to this underscore() method

server/src/json-api-deserialize.pipe.ts
```diff
@@ -8,7 +9,11 @@ export class JsonApiDeserializePipe implements PipeTransform {
 
     if (type === 'body') {
       return new Deserializer({
-        keyForAttribute: 'snake_case',
+        keyForAttribute: (attribute) => {
+          if (attribute === 'dcp500kpluszone')
+            return 'dcp_500kpluszone';
```
allthesignals, line 14: Thanks for fixing this! I'm sure this took a lot of head-scratching.

Is there not a more generic way to do this? Using regex or something?

Even a dictionary option would work: you could initialize the serializer/deserializer pipe from the controller with a hash of key-value pairs that behave as a manual translation dictionary for weird, custom one-offs like these.

allthesignals: @godfreyyeung  do you think there could end  up being a lot more of these situations? i would add a comment explaining that additional one-offs should also be setup in the serializer. could even mention that if this grows to more, we should probably look into a more generic way.

allthesignals (approved):
Thanks!!

I left some comments for improvement but it's your call whether to address.

server/src/json-api-deserialize.pipe.ts
godfreyyeung: Thanks for the suggestions! Good ideas about parameterizing the serializer/deserializer with custom dictionary of one-offs, and the comments. Both flashed across my mind but I got so wrapped up in sleuthing I forgot!  

So far it seems there is only this one property that starts with a digit after `dcp`. I think if we have more properties, we should definitely take some time to modularize this. I'll definitely add the comments to this PR to start

PR #774, opened 2020-10-14
12887 Basic set up for Filed Land Use

godfreyyeung:
### Summary
Originally meant to roughly spike an end-to-end setup of Filed Land Use form, but this can probably already satisfy the basic user stories.
This proves that Filed Land Use not only can re-use the same components, but can even re-use the same models, routes and
backend API controllers and services, since a Filed Land Use Package+Form is essentially a Draft Land Use Package+Form but with a different Package "type" value. 

Fixes #774 

### Technical details:
  - Both Filed and Draft Land Use forms make use of the `dcp_landuse` entity to represent the form, so we did not need to set up a new *-form controller or service for a new form type. 
  - The only thing distinguishing a Draft from a Filed Land Use form is the associated Package's type (filed or draft)
  - Upon Draft Land Use Form submission and sign-off by DCP, CRM copies the Draft dcp_landuse record to the new Filed dcp_landuse record of a new Filed Land Use package.
  - Most entities associated with `dcp_landuse`, like Related Actions (`dcp_relatedactions`), Subject Sites (`dcp_sitedatah`) are also generated new records associated with the new `dcp_landuse` record. 
  - Some entites like `dcp_applicantinformation` do not receive new records. Both the draft and filed land use forms modify the same applicant records.

### DIFF
--- a/client/app/models/package.js
+++ b/client/app/models/package.js
@@ -77,7 +77,8 @@ export default class PackageModel extends Model {
     if (this.dcpPackagetype === DCPPACKAGETYPE.RWCDS.code) {
       await this.rwcdsForm.save();
     }
-    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code) {
+    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code
+      || this.dcpPackagetype === DCPPACKAGETYPE.FILED_LU_PACKAGE.code) {
       await this.saveDeletedRecords(recordsToDelete);
       await this.landuseForm.save();
     }
@@ -126,7 +127,8 @@ export default class PackageModel extends Model {
         || this.rwcdsForm.hasDirtyAttributes
         || this.rwcdsForm.isAffectedZoningResolutionsDirty;
     }
-    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code) {
+    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code
+      || this.dcpPackagetype === DCPPACKAGETYPE.FILED_LU_PACKAGE.code) {
       return isPackageDirty
         || this.landuseForm.hasDirtyAttributes
         || this.landuseForm.isBblsDirty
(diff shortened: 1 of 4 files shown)


client/app/models/package.js
```diff
@@ -77,7 +77,8 @@ export default class PackageModel extends Model {
     if (this.dcpPackagetype === DCPPACKAGETYPE.RWCDS.code) {
       await this.rwcdsForm.save();
     }
-    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code) {
+    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_LU_PACKAGE.code
+      || this.dcpPackagetype === DCPPACKAGETYPE.FILED_LU_PACKAGE.code) {
```
allthesignals, line 81: Lots of complexity in this method. It's starting to look like it needs to a [polymorphic model](https://guides.emberjs.com/release/models/relationships/#toc_polymorphism).

Is there a way we could start making these computed properties with human-readable names that describe what the conditional is?

client/app/models/project.js
```diff
@@ -57,11 +57,29 @@ export default class ProjectModel extends Model {
...
   get landusePackages() {
+    const landusePackages = [
+      ...this.filedLandusePackages,
+      ...this.draftLandusePackages,
+    ];
+
+    return landusePackages;
+  }
+
+  get draftLandusePackages() {
```
allthesignals, line 68: good

allthesignals (approved):

client/app/models/package.js
godfreyyeung: ohh interesting idea. With polymorphism I'm guessing we can do away with these conditionals and instead parameterize which form is saved. 
```
// models/package.js
export default class PackageModel extends Model {
  @action
   save(){
       this.form.save();
   }
}
```

```
// models/draft-landuse-package.js
export default class DraftLandUsePackageModel extends Model {
  @belongsTo('landuse-form', { async: false })
  form 
}



```

godfreyyeung: > Is there a way we could start making these computed properties with human-readable names that describe what the conditional is?

Which Computed Properties are you referring to? 
We could further abstract this if statement to be something like `if(isLandusePackage)`

PR #776, opened 2020-10-15
12683: add zoning resolution dropdown to proposed actions component on landuse-form

trbmcginnis:
**Big Picture Summary**
Create a new Zoning Resolution dropdown that is part of the Proposed Actions section on the Land Use Form.

Proposed Actions section of the Land Use Form

- all zoning resolution names and ID's in CRM (around 600 records) are loaded onto the landuse-form. These zoning resolution names are used for the options list in the dropdown.
- when a user selects an option from the dropdown a Zoning Resolution id is sent to the backend and when the `dcp_landuseaction` entity is PATCHed, the zoning resolution id is binded to the landuseAction using the `dcp_zoningresolutionsectionactionispursuantto` field

**NOTE**: for the sake of time I created a fake ID field on the landuse-action model called `chosenZoningResolutionId` which is sent to the backend to bind the zoningResolution to the landuseAction. This can be updated in the future by doing this through updating the zoningResolution model directly and then using the "relationship" between the landuseAction and the zoningResolution (one-to-one) to access the chosen ID for binding. Can do this in a future refactor.

Addresses [AB#12683](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12683)

### DIFF
--- a/client/app/components/packages/landuse-form/zoning-resolution-dropdown.js
+++ b/client/app/components/packages/landuse-form/zoning-resolution-dropdown.js
@@ -1,0 +1,12 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class PackagesLanduseFormZoningResolutionComponent extends Component {
+  @tracked chosenZoningResolution;
+
+  get searchPlaceholder() {
+    if (this.args.landuseActionForm.data.zoningResolution) {
+      return this.args.landuseActionForm.data.zoningResolution.dcpZoningresolution;
+    } return 'Search Zoning Resolution Sections...';
+  }
+}
--- a/client/app/routes/landuse-form.js
+++ b/client/app/routes/landuse-form.js
@@ -18,6 +18,7 @@ export default class LanduseFormRoute extends Route.extend(AuthenticatedRouteMix
         'landuse-form.landuse-geographies',
         'landuse-form.lead-agency',
         'landuse-form.affected-zoning-resolutions',
+        'landuse-form.landuse-actions.zoning-resolution',
       ].join(),
     });
 
@@ -27,6 +28,7 @@ export default class LanduseFormRoute extends Route.extend(AuthenticatedRouteMix
     return RSVP.hash({
       package: landuseFormPackage,
       accounts: await this.store.findAll('account'),
+      zoningResolutions: await this.store.findAll('zoning-resolution'),
     });
   }
 }
--- a/server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
+++ b/server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
@@ -30,11 +30,11 @@ export class LanduseActionsController {
   async update(@Body() body, @Param('id') id) {
     const allowedAttrs = pick(body, LANDUSE_ACTION_ATTRS);
 
-      await this.crmService.update(
-        'dcp_landuseactions',
-        id,
-        allowedAttrs,
-      );
+    await this.crmService.update('dcp_landuseactions', id, {
+      ...allowedAttrs,
+
+      ...(body.chosen_zoning_resolution_id ? { 'dcp_zoningresolutionsectionactionispursuantto@odata.bind': `/dcp_zoningresolutions(${body.chosen_zoning_resolution_id})` } : {}),
+    });
 
     return {
       dcp_landuseactionid: id,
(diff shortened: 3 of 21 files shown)


server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
```diff
@@ -30,11 +30,11 @@ export class LanduseActionsController {
...
 
-      await this.crmService.update(
-        'dcp_landuseactions',
-        id,
-        allowedAttrs,
-      );
+    await this.crmService.update('dcp_landuseactions', id, {
+      ...allowedAttrs,
+
+      ...(body.chosen_zoning_resolution_id ? { 'dcp_zoningresolutionsectionactionispursuantto@odata.bind': `/dcp_zoningresolutions(${body.chosen_zoning_resolution_id})` } : {}),
```
godfreyyeung, line 36: sweet!

server/src/zoning-resolutions/zoning-resolutions.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+  @UseGuards(AuthenticateGuard)
+  @UsePipes(JsonApiDeserializePipe)
+  @Controller('zoning-resolutions')
+  export class ZoningResolutionsController {
+    constructor(
+      private readonly crmService: CrmService,
+    ) {}
+
+    @Get('/')
+    async zoningResolutions() {
```
godfreyyeung, line 52: is this get endpoint ever used directly?

client/app/routes/landuse-form.js
```diff
@@ -27,6 +28,7 @@ export default class LanduseFormRoute extends Route.extend(AuthenticatedRouteMix
     return RSVP.hash({
       package: landuseFormPackage,
       accounts: await this.store.findAll('account'),
+      zoningResolutions: await this.store.findAll('zoning-resolution'),
```
godfreyyeung, line 31: hmm.. it looks like you are both side loading zoning-resolutions along with landuse-form.landuse-actions (line 21) and also querying for them directly here?

client/app/components/packages/landuse-form/zoning-resolution-dropdown.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class PackagesLanduseFormZoningResolutionComponent extends Component {
+  @tracked chosenZoningResolution;
+
+  get searchPlaceholder() {
```
godfreyyeung, line 7: love this

server/src/packages/landuse-form/landuse-form.service.ts
```diff
@@ -38,6 +38,16 @@ export class LanduseFormService {
         dcp_dcp_landuse_dcp_zoningmapchanges_LandUseForm,
     `);
 
+    const { records: landuseActionsWithZr } = await this.crmService.get(`dcp_landuseactions`, `
```
godfreyyeung, line 41: Nice idea to break this out into a separate request to do the second-degree expand

server/src/packages/packages.controller.ts
```diff
@@ -243,7 +252,12 @@ import { CitypayService } from '../citypay/citypay.service';
             ],
             bbls: landuseForm.dcp_dcp_projectbbl_dcp_landuse,
             'related-actions': landuseForm.dcp_dcp_landuse_dcp_relatedactions,
-            'landuse-actions': landuseForm.dcp_dcp_landuse_dcp_landuseaction,
+            'landuse-actions': landuseForm.landuseActions.map(action => {
+              return {
```
godfreyyeung, line 256: nice

godfreyyeung (approved):

client/app/routes/landuse-form.js
trbmcginnis: I know I explained this in the chat but just reposting here for documentation purposes :) 

We need to load zoning resolutions in two different ways. One is for the dropdown that's displayed on the Proposed Actions section. This dropdown needs to list EVERY possible zoning resolution (around 600) that a user could choose from. The second way -- we need to know which zoning resolutions are tied to which landuse-actions so that when the form is reloaded, we can show which zoning resolution was selected and saved previously for that specific landuse-action. If we only got zoning resolutions through the `expand`, we'd have only zoning resolutions associated with the landuse-actions on the page, and therefore wouldn't be able to populate the dropdown.

PR #778, opened 2020-10-16
12730 create invoices route view

dhochbaum-dcp:
Creation of the initial /invoice/ route, model, and template.

### DIFF
--- a/client/app/templates/application.hbs
+++ b/client/app/templates/application.hbs
@@ -1,23 +1,25 @@
-<LabsUi::SiteHeader
-  @responsiveNav={{true}}
-  @responsiveSize="medium"
-  @betaNotice={{true}}
-  as |banner|
->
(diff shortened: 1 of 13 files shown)


client/app/templates/application.hbs
```diff
@@ -1,23 +1,25 @@
...
-  </banner.title>
-  <banner.nav>
-    <ul class="menu vertical medium-horizontal">
-      {{#if this.session.isAuthenticated}}
-        <Auth::UserBadge />
-      {{/if}}
-    </ul>
-  </banner.nav>
-</LabsUi::SiteHeader>
+{{#if this.header}}
```
godfreyyeung, line 1: Is this necessary for this PR? 

It looks like there were no changes to this file other than adding this conditional -- the rest is reformatting, is that right?

PR #783, opened 2020-10-16
12720 list invoices project view

allthesignals:
# Big Picture Summary
Adds the UI for fetching the payment link for invoices. Lists invoices and links to their show view.

Payment

No way to pay — this adds the UI to interact with the payment endpoint.

Just some components. The component hits the altready-built citypay service in the backend, which generates a citypay link. This link is sent to the user client.

Want to get some eyes on this, will fill out test coverage tomorrow.

Branches off some of @dhochbaum-dcp work.

### DIFF
--- a/client/app/components/project/package-section.hbs
+++ b/client/app/components/project/package-section.hbs
@@ -12,11 +12,81 @@
       </ol>
     </div>
     <div class="cell large-6">
-      <h4>Invoices</h4>
+      {{#let (flatten (map-by 'invoices' @packages)) as |invoices|}}
+        {{#if invoices}}
+          <h4>Invoices</h4>
 
-      {{#each (map-by 'invoices' @packages) as |invoice|}}
-        {{log invoice}}
-      {{/each}}
+          {{#each invoices as |invoice|}}
+            <Project::PackageSection::InvoiceInfo
+              @invoice={{invoice}}
+            />
+          {{/each}}
+
+          <br/>
+          <br/>
+          <button
+            type="button"
+            class="secondary button"
+            {{on 'click' (fn this.beginPayment)}}
+          >
+            Pay
+          </button>
+
+          <Ui::GenericModal
+            @show={{this.isPaying}}
+            @closeButtonAction={{fn (mut this.isPaying) false}}
+          >
+            <h3>Invoices to be paid</h3>
+
+            {{#each invoices as |invoice|}}
+              <Project::PackageSection::InvoiceInfo
+                @invoice={{invoice}}
+              />
+            {{/each}}
+
+            <br/>
+            <br/>
+            <h3>Total: ${{reduce (fn this.sum) 0 (map-by 'dcpGrandtotal' invoices)}}</h3>
+
+            <p>
+              The following link will direct you to the CityPay website for payment processing. Before proceeding, be ready with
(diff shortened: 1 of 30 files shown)


client/app/components/project/package-section.hbs
```diff
@@ -0,0 +1,92 @@
...
+      <ol class="no-bullet">
+        {{#each @packages as |package|}}
+          <Project::PackageListItem
+            @package={{package}}
+          />
+        {{/each}}
+      </ol>
+    </div>
+    <div class="cell large-6">
+      {{#let (flatten (map-by 'invoices' @packages)) as |invoices|}}
```
godfreyyeung, line 15: Hmm why flatten? After map by, does(n't) it look like this? `[{ dcpGrandTotal: 323, dcpName: 'name'}, {dcpGrandTotal: 342, dcpName: 'name2'....]`

godfreyyeung, line 15: Hmm. Each package can have multiple invoices, but each invoice is in an array itself?

client/app/components/project/package-section.hbs
```diff
@@ -0,0 +1,92 @@
...
+
+            {{#each invoices as |invoice|}}
+              <Project::PackageSection::InvoiceInfo
+                @invoice={{invoice}}
+              />
+            {{/each}}
+
+            <br/>
+            <br/>
+            <h3>Total: ${{reduce (fn this.sum) 0 (map-by 'dcpGrandtotal' invoices)}}</h3>
```
godfreyyeung, line 49: would this be more clear if it was all done in a getter/cp?

godfreyyeung (approved):

client/app/components/project/package-section.hbs
godfreyyeung: Having a hard time visualizing the before/after of this data structure

allthesignals: no, that's not what this says. each package has many invoices. this is mapping out invoices from package _s_, plural.

client/app/components/project/package-section.hbs
allthesignals: Just for the sake of discussion on handlebars helpers, can you try to make a guess at what this does?

client/app/components/project/package-section.hbs
godfreyyeung: @allthesignals  Hmm, I was under the assumption we should list invoices by package. Just realized we have a bit of discrepancy between our current wireframe design and the wording on acceptance criteria. Maybe it's fine though

client/app/components/project/package-section.hbs
godfreyyeung: It's basically a reduce function and you're supplying the accumulator function (simple add). It just looks a tiny bit wonky at first and could probably be done in a getter 

So it's adding up each dcpGrandtotal value on each invoice

client/app/components/project/package-section.hbs
allthesignals: > list invoices by package

@godfreyyeung hmm, I'm not sure how what I said contradicts this? It does list invoices by package. As far as I understand, there can't be invoices for draft packages.

allthesignals: Let me know if we should huddle on this @godfreyyeung

PR #788, opened 2020-10-20
12730 create invoices route view

dhochbaum-dcp:
Creation of the initial /invoice/ route, model, and template.


godfreyyeung (approved):
Thank you David!

PR #797, opened 2020-10-22
12549 Disable Sign-In Next btn and email input after click

godfreyyeung:
Disable the sign-in Next button while it queries contacts. This dramatically improves the sign-in UX. 

Fixes [AB#12549](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12549)


allthesignals (commented):
Thanks for doing this. Reminds me, I need to wrap the searchContacts promise in a try/catch to reset the `isLoginStarted` in case of network failure.

PR #798, opened 2020-10-22
Disable action to sync develop->landuse-form

godfreyyeung:
Disabled the action to sync develop->landuse-form, since it's no longer needed


allthesignals:
thank you sooo much!

PR #809, opened 2020-10-26
12871 - Two field names were reversed

dhochbaum-dcp:
#### Task/Bug Number
Fixes [AB#12871](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12871)

### Technical Explanation
The question "How long has the existing facility been at this location?" appears twice in this form - once as "dcpTextexistingfacility" and once as "dcpHowlonghasexistingfacilitybeenatthislocat".  They needed to be reversed, as they were each where the other should be.


godfreyyeung:
@dhochbaum-dcp  Can you write an explanation in the PR summary as to why this fixes the Task 12871?

godfreyyeung (approved):

godfreyyeung (approved):
Looks like you just need to fix up some tests and it will be good to go

PR #813, opened 2020-10-26
13054 Setup save button for FEIS

CeciliaCY:
### Summary
13054 Setup save button for FEIS

#### Task/Bug Number
Fix AB [13054](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13054)

### Technical Explanation
Same as Draft EAS Save Button

### DIFF
--- a/client/app/components/packages/feis/edit.hbs
+++ b/client/app/components/packages/feis/edit.hbs
@@ -5,39 +5,39 @@
 >
 
   <div class="grid-x grid-margin-x">
-      <div class="cell large-8">
+    <div class="cell large-8">
 
-        <section class="form-section">
-          <h1 class="header-large">
-            Final Environmental Impact Statement (FEIS)
-            <small class="text-weight-normal">
-              {{if @package.dcpPackageversion (concat '(V' @package.dcpPackageversion ')')}}
-            </small>
-          </h1>
+      <section class="form-section">
+        <h1 class="header-large">
+          Final Environmental Impact Statement (FEIS)
+          <small class="text-weight-normal">
+            {{if @package.dcpPackageversion (concat '(V' @package.dcpPackageversion ')')}}
+          </small>
+        </h1>
 
-          <h2 class="no-margin">
-            {{@package.project.dcpProjectname}}
-            <small class="text-weight-normal">
-              {{if @package.project.dcpName (concat '(' @package.project.dcpName ')')}}
-            </small>
-          </h2>
+        <h2 class="no-margin">
+          {{@package.project.dcpProjectname}}
+          <small class="text-weight-normal">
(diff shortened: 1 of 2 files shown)


client/app/components/packages/feis/edit.hbs
```diff
@@ -1,26 +1,43 @@
...
 
-      <section class="form-section">
-        <h1 class="header-large">
-          Final Environmental Impact Statement (FEIS)
-          <small class="text-weight-normal">
-            {{if @package.dcpPackageversion (concat '(V' @package.dcpPackageversion ')')}}
-          </small>
-        </h1>
+  <div class="grid-x grid-margin-x">
+      <div class="cell large-8">
```
allthesignals, line 8: I know this isn't part of your commit, but it looks like the spacing is messed up here again. 

It shouldn't be 4 spaces, it should only be 2 spaces for tabbing. 

Sorry we don't have a better linter to help with this. Can you fix please? Then, you'll need to re-align the tabbing for the rest of the diff below.

client/app/models/project.js
```diff
@@ -112,8 +112,8 @@ export default class ProjectModel extends Model {
 
   get eisPackages() {
     const eisPackages = [
-      ...this.deisPackages,
       ...this.feisPackages,
+      ...this.deisPackages,
```
godfreyyeung, line 116: Weird, this change should already be in develop

client/app/components/packages/feis/edit.hbs
allthesignals, line 8: I know this isn't part of your commit, but it looks like the spacing is messed up here again.

It shouldn't be 4 spaces, it should only be 2 spaces for tabbing.

Sorry we don't have a better linter to help with this. Can you fix please? Then, you'll need to re-align the tabbing for the rest of the diff below.

allthesignals (commented):
This looks good but where is the acceptance test for FEIS?

client/app/models/project.js
godfreyyeung: Why is this diff still rendering? Very strange...

godfreyyeung (approved):
After the fixes Matt mentioned, this should be good to go.
Also, see if rebasing this branch off develop fixes the strange behavior I mentioned for the `eisPackages` getter.

client/app/models/project.js
CeciliaCY: Please check again

client/app/components/packages/feis/edit.hbs
CeciliaCY: @allthesignals I think this is related to my vscode, I just updated it tab = 2 spaces. Please check again. Thanks!

PR #814, opened 2020-10-26
Re-enables the “email” param

allthesignals:
Takes a simpler approach, removes the dangerous “imposter” capability which re-generated the entire token.

### Summary
Allows the "email" param on the /projects view. Will show assigned projects for that e-mail.

#### Task/Bug Number
[AB#12660](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12660)

### Technical Explanation
Inserts the `email` search param into the session, allowing it to be referenced as-needed. Right now, the implementation is simplistic because it's only really needed in one place (dcp users may still view projects and forms that are not theirs). If we need to have this in other places, we should consider a pipe or guard or something that'll extract the correct "creeped" contcctid and swap teh current session with that id.

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -6,6 +6,7 @@
     <title>NYC Planning Applicants</title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="referrer" content="no-referrer-when-downgrade">
 
     {{content-for "head"}}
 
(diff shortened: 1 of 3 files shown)


godfreyyeung (approved):
Approving with preference that we perform a check for userEmail === dcpCreeper first

PR #835, opened 2020-10-30
12712: update  date fields on landuse-form to include  new date-picker component

trbmcginnis:
**Summary**
Implement date-picker component on the Related Actions and Proposed Actions sections on the Land Use Form.

**Task/Bug Number**
[AB#12712](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/12712)

**Technical Explanation**
- Implement Ember Flatpicker component throughout the landuse-form. All styling was previously addressed by Andy. 
- When the date value is sent to the backend, it shows up in the body formatted as a date-string within an array. Because of this I had to reformat the date in the related-actions.controller and landuse-actions.controller. Otherwise CRM will throw an error claiming that the date is formatted incorrectly.

**Open to suggestions on how to better go about this formatting issue!**

### DIFF
--- a/client/app/components/packages/landuse-form/proposed-action-editor.hbs
+++ b/client/app/components/packages/landuse-form/proposed-action-editor.hbs
@@ -202,27 +202,35 @@
 
     {{#if this.projectHasRequiredActionsAndFollowUpYes}}
 
-      <Ui::Question as |dcpDateofpreviousapprovalQ|>
+      <Ui::Question
+      as |dcpDateofpreviousapprovalQ|>
         <dcpDateofpreviousapprovalQ.Label>
           Date of previous approval
         </dcpDateofpreviousapprovalQ.Label>
 
-        <landuseActionForm.Field
-          @attribute="dcpDateofpreviousapproval"
-          type="date"
-          id={{dcpDateofpreviousapprovalQ.questionId}}
+        <EmberFlatpickr
+          data-test-dcpdateofpreviousapproval
+          @allowInput={{false}}
+          @date={{if landuseActionForm.data.dcpDateofpreviousapproval landuseActionForm.data.dcpDateofpreviousapproval null}}
+          @onChange={{fn (mut landuseActionForm.data.dcpDateofpreviousapproval)}}
+          @enableTime={{false}}
+          placeholder="Choose a date..."
         />
       </Ui::Question>
 
-      <Ui::Question as |dcpLapsedateofpreviousapprovalQ|>
+      <Ui::Question
+      as |dcpLapsedateofpreviousapprovalQ|>
         <dcpLapsedateofpreviousapprovalQ.Label>
           Lapse date of previous approval
         </dcpLapsedateofpreviousapprovalQ.Label>
 
-        <landuseActionForm.Field
-          @attribute="dcpLapsedateofpreviousapproval"
-          type="date"
-          id={{dcpLapsedateofpreviousapprovalQ.questionId}}
+        <EmberFlatpickr
+          data-test-dcplapsedateofpreviousapproval
+          @allowInput={{false}}
+          @date={{if landuseActionForm.data.dcpLapsedateofpreviousapproval landuseActionForm.data.dcpLapsedateofpreviousapproval null}}
+          @onChange={{fn (mut landuseActionForm.data.dcpLapsedateofpreviousapproval)}}
+          @enableTime={{false}}
+          placeholder="Choose a date..."
         />
       </Ui::Question>
 
@@ -291,15 +299,19 @@
         />
       </Ui::Question>
 
-      <Ui::Question as |dcpRecordationdateQ|>
+      <Ui::Question
+      as |dcpRecordationdateQ|>
         <dcpRecordationdateQ.Label>
           Recordation date of Legal Instrument
         </dcpRecordationdateQ.Label>
 
-        <landuseActionForm.Field
-          @attribute="dcpRecordationdate"
-          type="date"
-          id={{dcpRecordationdateQ.questionId}}
+        <EmberFlatpickr
+          data-test-dcprecordationdate
+          @allowInput={{false}}
+          @date={{if landuseActionForm.data.dcpRecordationdate landuseActionForm.data.dcpRecordationdate null}}
+          @onChange={{fn (mut landuseActionForm.data.dcpRecordationdate)}}
+          @enableTime={{false}}
+          placeholder="Choose a date..."
         />
       </Ui::Question>
 
--- a/client/tests/acceptance/user-can-click-landuse-form-edit-test.js
+++ b/client/tests/acceptance/user-can-click-landuse-form-edit-test.js
@@ -11,6 +11,7 @@ import { setupApplicationTest } from 'ember-qunit';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 import { authenticateSession } from 'ember-simple-auth/test-support';
 import { selectChoose } from 'ember-power-select/test-support';
+import { setFlatpickrDate } from 'ember-flatpickr/test-support/helpers';
 import exceedMaximum from '../helpers/exceed-maximum-characters';
 
 const saveForm = async () => {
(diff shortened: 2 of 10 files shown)


server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
```diff
@@ -29,16 +29,35 @@ export class LanduseActionsController {
...
       allowedAttrs,
     };
   }
+
+  // when a user selects a date from the frontend date-picker, this value
+  // is sent to backend as a datestring in an array, so we need to grab
+  // the first item in that array. When the user does not use the date-picker,
+  // and instead the date value comes from CRM (still exists in the body), the
+  // date is just formatted as an isolated datestring and does not need to be reformatted.
+  async setDate(bodyDateField) {
```
godfreyyeung, line 56: It looks like maybe `pickDate` would be a better name? This function isn't setting anything, but retrieving a date field from an array, if available. It looks like it serves a similar purpose to the `pick()` function, but customized to dates. 

Also this doesn't need to be async. I don't see any asynchronous functions within setDate.

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
```diff
@@ -29,16 +29,35 @@ export class LanduseActionsController {
   @Patch('/:id')
   async update(@Body() body, @Param('id') id) {
     const allowedAttrs = pick(body, LANDUSE_ACTION_ATTRS);
+    const dateOfPreviousApproval = await this.setDate(body.dcp_dateofpreviousapproval);
```
godfreyyeung, line 32: Remove `await` here after removing `async` from `setDate()`

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
```diff
@@ -29,16 +29,35 @@ export class LanduseActionsController {
...
+  // when a user selects a date from the frontend date-picker, this value
+  // is sent to backend as a datestring in an array, so we need to grab
+  // the first item in that array. When the user does not use the date-picker,
+  // and instead the date value comes from CRM (still exists in the body), the
+  // date is just formatted as an isolated datestring and does not need to be reformatted.
+  async setDate(bodyDateField) {
+    if (bodyDateField) {
+      return Array.isArray(bodyDateField) ? bodyDateField[0] : bodyDateField;
+    } else {
+      return null;
```
godfreyyeung, line 60: comment [A]
hmm... it looks like if one of the date fields, say `dcp_dateofpreviousapproval`, isn't even in the body, then this function returns null. What if this overwrites a value in CRM? Are we assuming that if a value exists in CRM, then bodyDateField would never be null?

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
```diff
@@ -29,16 +29,35 @@ export class LanduseActionsController {
...
     const allowedAttrs = pick(body, LANDUSE_ACTION_ATTRS);
+    const dateOfPreviousApproval = await this.setDate(body.dcp_dateofpreviousapproval);
+    const lapseDateOfPreviousApproval = await this.setDate(body.dcp_lapsedateofpreviousapproval);
+    const recordationDate = await this.setDate(body.dcp_recordationdate);
 
     await this.crmService.update('dcp_landuseactions', id, {
       ...allowedAttrs,
 
       ...(body.chosen_zoning_resolution_id ? { 'dcp_zoningresolutionsectionactionispursuantto@odata.bind': `/dcp_zoningresolutions(${body.chosen_zoning_resolution_id})` } : {}),
+      dcp_dateofpreviousapproval: dateOfPreviousApproval,
```
godfreyyeung, line 40: Relating to [Comment A] below, I think here these fields should only be set on the update object if the values aren't null.

Here it looks like you would still send an update request with `null` assigned to these date properties.

server/src/packages/landuse-form/related-actions/related-actions.controller.ts
```diff
@@ -31,12 +31,14 @@ import {
...
   
-        await this.crmService.update(
-          'dcp_relatedactionses',
-          id,
-          allowedAttrs,
-        );
+      await this.crmService.update('dcp_relatedactionses', id, {
+        ...allowedAttrs,
+    
+        dcp_applicationdate: applicationDate,
```
godfreyyeung, line 40: same comments as for landuse-actions

server/src/packages/landuse-form/related-actions/related-actions.controller.ts
```diff
@@ -63,5 +70,17 @@ import {
         id,
       };
     }
+
+    // when a user selects a date from the frontend date-picker, this value
+    // is sent to backend as a datestring in an array, so we need to grab
+    // the first item in that array. When the user does not use the date-picker,
+    // and instead the date value comes from CRM (still exists in the body), the
+    // date is just formatted as an isolated datestring and does not need to be reformatted.
+    async setDate(bodyDateField) {
```
godfreyyeung, line 79: same comments as for landuse-actions

client/tests/acceptance/user-can-click-landuse-form-edit-test.js
```diff
@@ -608,6 +613,9 @@ module('Acceptance | user can click landuse form edit', function (hooks) {
     await click('[data-test-radio="dcpApplicantispublicagencyactions"][data-test-action="ZC"][data-test-radio-option="Yes"]');
     await click('[data-test-radio="dcpApplicantispublicagencyactions"][data-test-action="ZA"][data-test-radio-option="No"]');
 
+    const recordationDate = new Date(2020, 0, 1);
+    setFlatpickrDate('[data-test-dcprecordationdate]', recordationDate);
```
godfreyyeung, line 617: cool!

client/app/components/packages/landuse-form/proposed-action-editor.hbs
```diff
@@ -291,15 +299,19 @@
...
+      as |dcpRecordationdateQ|>
         <dcpRecordationdateQ.Label>
           Recordation date of Legal Instrument
         </dcpRecordationdateQ.Label>
 
-        <landuseActionForm.Field
-          @attribute="dcpRecordationdate"
-          type="date"
-          id={{dcpRecordationdateQ.questionId}}
+        <EmberFlatpickr
```
godfreyyeung, line 308: Awesome, excited about this new standard for picking dates

godfreyyeung (changes requested):

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
trbmcginnis: Okay sounds good

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
trbmcginnis: these date fields are always included in the body regardless of whether the user has updated them manually or not. So if the user did not change anything, what's in the body would be what was saved in CRM previously.

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
trbmcginnis: I think this goes back to our conversation about whether we should make just the **field** conditional OR the **entire update object**. I tried to make it as simple as possible by just doing the field, but maybe doing the entire object is safe in the long run

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
godfreyyeung: ahh ok. Got you. Okay if that holds true then sounds good for now.
Just a note: In general I think it's better to program a little defensively just in case this endpoint is used differently. 
For example, if someone sent a custom PATCH request not through Ember, but through Postman and forgot to include `dcp_dateofpreviousapproval`, then the existing value in CRM would be overwritten. Just something to keep in mind!

server/src/packages/landuse-form/landuse-actions/landuse-actions.controller.ts
godfreyyeung: Sounds good. You put that very clearly and it sounds like you're aware of the Pros, Cons, dangers and tradeoffs. So we're making a calculated decision here. Thanks!

godfreyyeung (approved):

allthesignals:
@godfreyyeung @trbmcginnis merge?

PR #846, opened 2020-11-06
13238 Saturate Project Packages with Documents

godfreyyeung:
### Summary
This PR makes sure that when a user visits the individual Project page, the sideloaded Packages will have a `documents` property that is hydrated with associated documents. 
image

This is in preparation to list out documents by package on the individual Project page.

This PR also unifies document saturation+munging code to the document service

#### Task/Bug Number
Fixes [AB#13238](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13238)

### Technical Explanation
  - Notice how `this.documentService.packageWithDocuments(<package>)` is now referenced in both the Packages and Projects services! The documents array creation is shared by both of those services, so it made sense to move it into one function inside the Document service.

### DIFF
--- a/server/src/packages/packages.service.ts
+++ b/server/src/packages/packages.service.ts
@@ -104,7 +104,7 @@ export class PackagesService {
 
     // Get package
     try {
-        const { records: [firstPackage] } = await this.crmService.get('dcp_packages', `
+        let { records: [firstPackage] } = await this.crmService.get('dcp_packages', `
         $select=${PACKAGE_ATTRS.join(',')}
         &$filter=dcp_packageid eq ${packageId}
         &$expand=dcp_project($select=${PROJECT_ATTRS.join(',')})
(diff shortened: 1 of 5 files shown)


server/src/packages/packages.service.ts
```diff
@@ -144,16 +133,16 @@ export class PackagesService {
...
       }
 
+      // re-alias project property
+      const {
+        dcp_project,
+      } = firstPackage;
+
       return {
         ...firstPackage,
         ...formData,
```
godfreyyeung, line 143: Yes, there's some inconsistency in how we inject related entities like invoices, forms or documents into projects or packages. I think in the future we can look into moving the injection code out of the package and services and more into their respective services

godfreyyeung:
Closing for now because realize this needs to be transferred to Zap Search

PR #853, opened 2020-11-10
13318 Fixed show page display

dhochbaum-dcp:
#### Task/Bug Number
Fixes [AB#13318](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13318)

Changed if statement so that when 'If this is a follow-up Action, indicate the previously approved Action Code' is not null, the response to that question and the questions below appear on the show page.

This also fixes [AB#13317](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13317)


client/app/components/packages/landuse-form/show.hbs
```diff
@@ -605,16 +605,15 @@
...
-            <Ui::Answer @beside={{true}} as |A|>
-              <A.Prompt>
-                This is a follow-up Action. The previous approval Action Code:
-              </A.Prompt>
-              <A.Field>
-                {{optionset 'landuseAction' 'actions' 'label' landuseAction.dcpPreviouslyapprovedactioncode}}
-              </A.Field>
-            </Ui::Answer>
-
-            {{#if this.projectHasRequiredActionsAndFollowUpYes}}
```
godfreyyeung, line 617: Perhaps the issue is that this getter `projectHasRequiredActionsAndFollowUpYes` needs to be added to landuse-form/show.js, just like how it is in landuse-form/edit.js?

godfreyyeung (commented):
Can you explain a little more? I'm looking at the Task and it doesn't have much detail, just a title. Having a hard time understanding the problem and how this fixes it.

client/app/components/packages/landuse-form/show.hbs
dhochbaum-dcp: That getter is unnecessary.  If landuseAction.dcpPreviouslyapprovedactioncode is not null, we want to show it and all of the responses that had previously been in the {{#if this.projectHasRequiredActionsAndFollowUpYes}} block.  If landuseAction.dcpPreviouslyapprovedactioncode is null, then we don't want to show them.

dhochbaum-dcp:
> Can you explain a little more? I'm looking at the Task and it doesn't have much detail, just a title. Having a hard time understanding the problem and how this fixes it.

If someone has selected an action code, it should show that response, and the responses to the conditional questions below it.  The page had been showing the question and then not showing the conditional questions below it, regardless of the response.

client/app/components/packages/landuse-form/show.hbs
godfreyyeung: Hmm okay I was concerned because this getting takes into account `dcpFollowuptopreviousaction` , however looking at the action it looks like the value of that field just reflects whether dcpPreviouslyapprovedactioncode is filled.

godfreyyeung (approved):

PR #858, opened 2020-11-12
Correct arguments for contact mgmt

allthesignals:
Pass the save closure action and the correct reference to the project model

### Summary
Before, I forgot to pass the closure action that actually performs the save to the contct mgmt component. This change passes that and also uses the correct reference to the project model.

#### Task/Bug Number
Fixes [AB#13262](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13262)


godfreyyeung (approved):

trbmcginnis (approved):
Thanks!

PR #862, opened 2020-11-12
Use correct attr names for entity; join in labels; iterate line items

allthesignals:
### Summary
Fixes the issue by using the correct, "real" attribute names as they're named on the invoice entity.

Includes the line items of the invoice using an expand. Avoids using a model and simply provides a read-only array of line items. 

Applies the overwriteValuesWithLabels helper to get neater naming for each value.

#### Task/Bug Number
Fixes [AB#13298](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13298)

### DIFF
--- a/client/app/templates/invoice.hbs
+++ b/client/app/templates/invoice.hbs
@@ -14,7 +14,7 @@
   </p>
   <p class="cell medium-6">
     Project: <br class="show-for-medium">
-    <strong class="text-xlarge">{{@model.dcpProjectname}}</strong>
+    <strong class="text-xlarge">{{@model.dcpInvoiceprojectname}}</strong>
   </p>
 </div>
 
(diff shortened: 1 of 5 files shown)


godfreyyeung (approved):
You're on 🔥  today!

client/app/templates/invoice.hbs
```diff
@@ -27,15 +27,14 @@
...
-    <tr>
-      <td>{{dcpApplication.dcpApplicationNumber}}</td>
-      <td>{{dcpApplication.dcpAction}}</td>
-      <td class="text-right font-mono">{{dcpApplication.dcpFee}}</td>
-    </tr>
+    {{#each @model.lineitems as |lineitem|}}
+      <tr>
+        <td>{{lineitem.dcp-applicationnumber}}</td>
+        <td>{{lineitem.dcp-action}}</td>
+        <td class="text-right font-mono">{{lineitem.dcp-fee}}</td>
```
godfreyyeung, line 34: this works?

allthesignals: @godfreyyeung yeah why not? it doesn't camelize bc it's not a real model. the json api serializer does convert it, however, because it does so recursively for all object attr names.

PR #863, opened 2020-11-12
Fix invoice filter logic; hide pay button when no invoices need payment

allthesignals:
Show checkmark icon when paid

### Summary
 - Modifies the filtering logic for invoices a little to allow for paid or approved invoices
 - include the statuscode attribute on invoices for icons and filtering
 - hide pay button when no payable invoices exist

#### Task/Bug Number
Fixes [AB#13299](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13299)

### DIFF
--- a/client/app/components/project/package-section.hbs
+++ b/client/app/components/project/package-section.hbs
@@ -20,21 +20,24 @@
           {{#if invoices}}
             <h4>Invoices</h4>
 
-            {{#each invoices as |invoice|}}
-              <Project::PackageSection::InvoiceInfo
-                @invoice={{invoice}}
-              />
-            {{/each}}
+            <ul>
+              {{#each invoices as |invoice|}}
+                <Project::PackageSection::InvoiceInfo
+                  @invoice={{invoice}}
+                />
+              {{/each}}
+            </ul>
 
-            <br/>
-            <br/>
-            <button
-              type="button"
-              class="secondary button"
-              {{on 'click' (fn this.beginPayment)}}
-            >
-              Pay
-            </button>
+            {{#if this.hasPayableInvoices}}
+              <br/>
+              <button
+                type="button"
+                class="secondary button"
+                {{on 'click' (fn this.beginPayment)}}
+              >
+                Pay
+              </button>
+            {{/if}}
 
             <Ui::GenericModal
               @show={{this.isPaying}}
--- a/client/app/components/project/package-section.js
+++ b/client/app/components/project/package-section.js
@@ -21,6 +21,16 @@ export default class ProjectPackageSectionComponent extends Component {
     return payablePackage;
   }
 
+  get invoices() {
+    return this.args.packages
+      .mapBy('invoices')
+      .reduce((acc, curr) => acc.concat(curr), []);
+  }
+
+  get hasPayableInvoices() {
+    return this.invoices.filter((invoice) => invoice.statuscode === 'Approved').length >= 1;
+  }
+
   @action
   async beginPayment() {
     this.isPaying = true;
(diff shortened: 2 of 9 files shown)


client/app/components/project/package-section.js
```diff
@@ -21,6 +21,16 @@ export default class ProjectPackageSectionComponent extends Component {
     return payablePackage;
   }
 
+  get invoices() {
```
trbmcginnis, line 24: Love this as a computed prop

trbmcginnis, line 24: Ooh okay, so if I'm following correctly -- you're trying to create an array of alll the invoices from every package regardless of package type -- can you explain why the reduce necessary here?

client/app/components/project/package-section.js
```diff
@@ -21,6 +21,16 @@ export default class ProjectPackageSectionComponent extends Component {
...
+    return this.args.packages
+      .mapBy('invoices')
+      .reduce((acc, curr) => acc.concat(curr));
+  }
+
+  get hasPayableInvoices() {
+    return this.invoices.filter((invoice) => invoice.statuscode === 'Approved').length >= 1;
```
trbmcginnis, line 31: So an "Approved" invoice is one that has yet to be paid? Or one that has already been paid?

client/app/components/project/package-section/invoice-info.hbs
```diff
@@ -1,12 +1,19 @@
...
+    @model={{@invoice}}
+    @query={{hash header=false}}
+  >
+    {{@invoice.dcpName}}
+  </LinkTo>
 
-<em>{{moment-format (utc @invoice.dcpInvoicedate) 'YYYY-MM-DD'}}</em>
+  <strong>{{@invoice.dcpGrandtotal}}</strong>
+
+  <em>{{moment-format (utc @invoice.dcpInvoicedate) 'YYYY-MM-DD'}}</em>
```
trbmcginnis, line 18: We have a `dateDisplay` ui component, might be better to use that?

client/app/components/project/package-section.hbs
```diff
@@ -16,41 +16,43 @@
...
+              {{#each this.invoices as |invoice|}}
                 <Project::PackageSection::InvoiceInfo
                   @invoice={{invoice}}
                 />
               {{/each}}
 
               <br/>
               <br/>
-              <h3>Total: ${{reduce (fn this.sum) 0 (map-by 'dcpGrandtotal' invoices)}}</h3>
+              <h3>Total: ${{reduce (fn this.sum) 0 (map-by 'dcpGrandtotal' this.invoices)}}</h3>
```
trbmcginnis, line 55: whoa never seen this used before in a template, super cool!

trbmcginnis (approved):
Just a couple comments, but looks great, thanks!

client/app/components/project/package-section/invoice-info.hbs
allthesignals: Oops, yes! Actually, I changed it to where it provides the CRM-formatted values instead of UTC values.

client/app/components/project/package-section.js
allthesignals: Yep, that's it. It's confusing.

client/app/components/project/package-section.js
allthesignals: The reducer here is a common array flattener — if I map out the invoices of packages, I get an array of arrays. I need to flatten it out so I can treat them as a one dimensional array.

client/app/components/project/package-section.hbs
allthesignals: @trbmcginnis https://github.com/DockYard/ember-composable-helpers

PR #869, opened 2020-11-16
Force reload entire invoice; format raw values

allthesignals:
### Summary
Small changes to force reload of the invoice route. Before, it was only getting a partial fragment of the invoice from a previous route, and using that fragment to render the view. This change forces a reload of the entire entity.

#### Task/Bug Number
Fixes [AB#13298](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13298)

Includes bug fix in which "hasPayableInvoice" computed was not calculating correctly because of the nature of ember data's hasMany relationships

### DIFF
--- a/client/app/components/project/package-section.js
+++ b/client/app/components/project/package-section.js
@@ -23,13 +23,11 @@ export default class ProjectPackageSectionComponent extends Component {
 
   get invoices() {
     return this.args.packages
-      .map(pkg => { return pkg.get('invoices').toArray(); })
+      .map((pkg) => pkg.get('invoices').toArray())
       .flat();
   }
 
   get hasPayableInvoices() {
-    console.log(this.invoices);
-
     return this.invoices.filter((invoice) => invoice.statuscode === 'Approved').length >= 1;
   }
 
--- a/client/app/components/project/package-section.js
+++ b/client/app/components/project/package-section.js
@@ -23,11 +23,13 @@ export default class ProjectPackageSectionComponent extends Component {
 
   get invoices() {
     return this.args.packages
-      .mapBy('invoices')
-      .reduce((acc, curr) => acc.concat(curr), []);
+      .map(pkg => { return pkg.get('invoices').toArray(); })
+      .flat();
   }
 
   get hasPayableInvoices() {
+    console.log(this.invoices);
+
     return this.invoices.filter((invoice) => invoice.statuscode === 'Approved').length >= 1;
   }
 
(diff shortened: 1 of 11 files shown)


trbmcginnis:
Would you mind linking the Azure Board task? I think the PR still has the default

godfreyyeung (approved):
Looks good!

client/app/components/project/package-section.js
```diff
@@ -23,8 +23,8 @@ export default class ProjectPackageSectionComponent extends Component {
 
   get invoices() {
     return this.args.packages
-      .mapBy('invoices')
-      .reduce((acc, curr) => acc.concat(curr), []);
+      .map((pkg) => pkg.get('invoices').toArray())
+      .flat();
```
trbmcginnis, line 27: Okay so this is alllll invoices across all packages in one array, cool

client/app/helpers/to-locale-string.js
```diff
@@ -0,0 +1,5 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(function toLocaleString(number) {
+  return number.toLocaleString();
```
trbmcginnis, line 4: aw okay so this makes it so that it match the language-sensitive? cool! Haven't seen this before

server/src/projects/projects.service.ts
```diff
@@ -190,6 +190,10 @@ export class ProjectsService {
         packages: projectPackages.map(pkg => ({
           ...pkg,
           invoices: joinInvoiceLabels(pkg.dcp_dcp_package_dcp_projectinvoice_package),
+
+          // NOTE: "virtual" property. Maybe not be available in other requests for pkg data!
```
trbmcginnis, line 194: Can u explain this a bit more?

client/app/components/project/package-section.js
allthesignals: yup! I didn't even know about the `flat` method.

I was running into a weird issue with Ember Data hasMany relationships, though, which is why I had to map and `toArray` stuff. Note to self: Ember Data relationships are NOT normal arrays.

server/src/projects/projects.service.ts
allthesignals: It's generated in the API based on other values. It's not a "real" property in CRM, similar to the key names we use for entity relationships.

allthesignals:
> @allthesignals Would you mind linking the Azure Board task? I think the PR still has the default

my b

PR #886, opened 2020-11-19
12656: prepopulate applicant form on landuse-form from pas-form

trbmcginnis:
**Summary**
Pre-populate the applicants section on the Land Use Form from data that was saved on the Pas Form. 

When the user loads the Land Use Form, they should see all the applicants that they previously entered on the Pas Form.

**Task/Bug Number**
Fixes AB#12656

**Technical Explanation**
When a user loads the landuse-form, we grab the `_dcp_project_value` from the current landuse-form. We use this value to find the pas-form that is associated with that project. We check for whether the applicants on the landuse-form have matching emails with those on the pas-form. If these are matching, we do not run `crmService.update`. If they do not match,  then we use `crmService.update` to associate the applicant with the landuse-form using `@odata.bind`. This should make these already-created applicants now show up on the landuse-form when the user loads the page.

### DIFF
--- a/server/src/packages/landuse-form/landuse-form.attrs.ts
+++ b/server/src/packages/landuse-form/landuse-form.attrs.ts
@@ -448,5 +448,6 @@ export const LANDUSE_FORM_ATTRS = [
   'dcp_landuse_dcp_sitedatageneralform_landuseform',
   'dcp_landuse_Annotations',
   'dcp_dcp_landuse_dcp_landuseaction',
-  'dcp_dcp_landuse_dcp_sitedatahform_landuseform'
+  'dcp_dcp_landuse_dcp_sitedatahform_landuseform',
+  '_dcp_project_value'
 ];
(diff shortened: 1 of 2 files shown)


server/src/packages/landuse-form/landuse-form.attrs.ts
```diff
@@ -448,5 +448,6 @@ export const LANDUSE_FORM_ATTRS = [
   'dcp_landuse_dcp_sitedatageneralform_landuseform',
   'dcp_landuse_Annotations',
   'dcp_dcp_landuse_dcp_landuseaction',
-  'dcp_dcp_landuse_dcp_sitedatahform_landuseform'
+  'dcp_dcp_landuse_dcp_sitedatahform_landuseform',
+  '_dcp_project_value'
```
CeciliaCY, line 452: Check this **_**dcp_project_value to see if this should be dcp_dcp_project_value or dcp_project_value

PR #895, opened 2020-12-02
13394 Display the CEQR Questionnaire on the submitted Filed EAS form

CeciliaCY:
### Summary
13394 Display the CEQR Questionnaire on the submitted Filed EAS form

#### Task/Bug Number
Fixes [AB#13394](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13394)

### Technical Explanation
Add CEQR section in show.hbs based on edit.hbs. Also add SaveableForm to enable dynamic nav on the right.

PR #901, opened 2020-12-09
13386 Add CEQR to DSOW show page

CeciliaCY:
### Summary
13386 Display the CEQR Questionnaire on the submitted DSOW form

#### Task/Bug Number
[Fixes AB#13386](https://dcp-paperless.visualstudio.com/ZAP%20Portals/_workitems/edit/13386)

### Technical Explanation
Add CEQR section in show.hbs based on edit.hbs. Also add SaveableForm to enable dynamic nav on the right. Similar to #895

### DIFF
--- a/client/app/components/packages/scope-of-work-draft/show.hbs
+++ b/client/app/components/packages/scope-of-work-draft/show.hbs
@@ -44,10 +44,8 @@
             <saveableForm.Section
               @title="Ceqr Invoice Questionnaire" 
               @elementId="ceqr-invoice-questionnaire"
+              data-test-ceqr-invoice-questionnaire
             >
-              <saveableForm.SaveableForm
-                @model={{ceqrInvoiceQuestionnaire}}
-              >
                 <Ui::Answer @beside={{true}} as |A|>
                   <A.Prompt>
                     Is the sole applicant a government agency (City, State, or Federal)?
@@ -96,7 +94,6 @@
                   </Ui::Answer>
                 {{/if}}
               {{/if}}
-             </saveableForm.SaveableForm>
             </saveableForm.Section>
           {{/let}}
         {{/if}}
--- a/client/app/components/packages/scope-of-work-draft/show.hbs
+++ b/client/app/components/packages/scope-of-work-draft/show.hbs
@@ -42,12 +42,11 @@
       {{#if (gte saveableForm.data.ceqrInvoiceQuestionnaires.length 1)}}
           {{#let saveableForm.data.singleCeqrInvoiceQuestionnaire as |ceqrInvoiceQuestionnaire|}}
             <saveableForm.Section
-            @title="Ceqr Invoice Questionnaire" 
-            @elementId="ceqr-invoice-questionnaire"
+              @title="Ceqr Invoice Questionnaire" 
+              @elementId="ceqr-invoice-questionnaire"
             >
               <saveableForm.SaveableForm
                 @model={{ceqrInvoiceQuestionnaire}}
-                @validators={{array (hash) (hash)}} 
               >
                 <Ui::Answer @beside={{true}} as |A|>
                   <A.Prompt>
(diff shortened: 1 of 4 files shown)


client/app/components/packages/scope-of-work-draft/show.hbs
```diff
@@ -1,71 +1,139 @@
...
-        data-test-package-statuscode
-      >
-        {{optionset 'bbl' 'boroughs' 'label' @package.project.dcpBorough}} |
-        {{optionset 'package' 'statuscode' 'label' @package.statuscode}}
-      </p>
-    </section>
+      {{#if (gte saveableForm.data.ceqrInvoiceQuestionnaires.length 1)}}
+          {{#let saveableForm.data.singleCeqrInvoiceQuestionnaire as |ceqrInvoiceQuestionnaire|}}
+            <saveableForm.Section
+            @title="Ceqr Invoice Questionnaire"
```
godfreyyeung, line 45: tiny nit: indent the attributes here

client/app/components/packages/scope-of-work-draft/show.hbs
```diff
@@ -1,71 +1,139 @@
...
+      {{#if (gte saveableForm.data.ceqrInvoiceQuestionnaires.length 1)}}
+          {{#let saveableForm.data.singleCeqrInvoiceQuestionnaire as |ceqrInvoiceQuestionnaire|}}
+            <saveableForm.Section
+            @title="Ceqr Invoice Questionnaire" 
+            @elementId="ceqr-invoice-questionnaire"
+            >
+              <saveableForm.SaveableForm
+                @model={{ceqrInvoiceQuestionnaire}}
+                @validators={{array (hash) (hash)}} 
+              >
```
godfreyyeung, line 51: Why is this necessary? It doesn't look like it is being used. There is no yield like the first SaveableForm above. I.e. no `as |cequirInvoiceSaveableForm|`

I think this was only necessary on the Edit page, but doesn't look necessary here

client/app/components/packages/scope-of-work-draft/show.hbs
```diff
@@ -1,71 +1,139 @@
...
+              </small>
+            </li>
+          {{/each}}
+        </ul>
+      </saveableForm.Section>
+    </div>
+
+    <div class="cell large-4 sticky-sidebar">
+      <Messages::Assistance class="large-margin-top" />
+      <saveableForm.PageNav>
```
godfreyyeung, line 134: Since there is no content inside this component, I think you can just write
`<saveableForm.PageNav />` 

instead of the opening and closing tags 
```
  <saveableForm.PageNav>
  </saveableForm.PageNav>
```

godfreyyeung (commented):
Thanks! So fast. I wrote a question and a few comments

client/app/components/packages/scope-of-work-draft/show.hbs
CeciliaCY: To bring in ceqrInvoiceQuestionnaire. Maybe you have a better way.

client/app/components/packages/scope-of-work-draft/show.hbs
CeciliaCY: Fixed

client/app/components/packages/scope-of-work-draft/show.hbs
godfreyyeung: It should still work without this. 
When you display a value like on line 57, it is referencing the  `ceqrInvoiceQuestionnaire` block variable from line 43. 

Previously on the Edit page, there is a new `saveableForm.SaveableForm` component in order to render Input fields, etc, but now that's not necessary because the user already inputted the data

client/app/components/packages/scope-of-work-draft/show.hbs
CeciliaCY: Corrected

client/app/components/packages/scope-of-work-draft/show.hbs
CeciliaCY: Thanks @godfreyyeung , corrected and also added a test case.

godfreyyeung (approved):

PR #922, opened 2020-12-23
13527: remove the reformatting of dates into strings for milestones to allow for sorting

trbmcginnis:
### Summary
Make sure that milestones list on the individual project page are sorted by Planned Start Date for "Not Started" milestones and by Actual Start Date for milestones that are "In Progress" or "Completed"

#### Task/Bug Number
Fixes [AB#13527](https://dcp-paperless.visualstudio.com/d36fd830-9029-4b77-b0c4-b0df2012eb98/_workitems/edit/13527)

### Technical Explanation
- Previously we were reformatting all fields from `dcp_projectmilestone` entity. This caused four date fields to be reformatted into strings when sent to the frontend. This PR removes this reformatting so that the dates in the frontend are formatted as dates. We then instead use `date-display` helper to reformat them to strings for display purposes on the template. 
- A new computed property was added called `sortedGroupedMilestones` which sorts the milestones based off of their start date

### DIFF
--- a/server/src/projects/projects.service.ts
+++ b/server/src/projects/projects.service.ts
@@ -7,7 +7,7 @@ import { joinLabels as joinInvoiceLabels } from '../invoices/invoices.service';
 import { NycidService } from '../contact/nycid/nycid.service';
 import { CrmService } from '../crm/crm.service';
 import { overwriteCodesWithLabels } from '../_utils/overwrite-codes-with-labels';
-import { MILESTONE_ATTRS } from './projects.attrs';
+import { MILESTONE_ATTRS, MILESTONE_NON_DATE_ATTRS } from './projects.attrs';
 
 const APPLICANT_ACTIVE_STATUS_CODE = 1;
 const PROJECT_ACTIVE_STATE_CODE = 0;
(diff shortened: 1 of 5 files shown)


trbmcginnis:
### Milestone sorting before

milestone_sorting_before

### Milestone sorting after

milestone_sorting_after

server/src/projects/projects.service.ts
```diff
@@ -229,7 +229,15 @@ export class ProjectsService {
 
           ...projectMilestone,
         }));
-      const formattedProjectMilestones = overwriteCodesWithLabels(projectMilestones, [...MILESTONE_ATTRS]);
+
+      // if we include dates when we reformat below then the dates end up being sent to the frontend as strings,
+      // so we exclude them here
+      const MILESTONE_ATTRS_WITHOUT_DATES = [
```
godfreyyeung, line 235: optional suggestion:
Maybe instead of defining this here, in `projects/projects.attrs.ts` , you can compose MILESTONE_ATTRS using two different subsets, like 
```
export const MILESTONE_DATE_ATTRS = [
  'dcp_actualstartdate',
  'dcp_actualenddate',
  'dcp_plannedcompletiondate',
  'dcp_plannedstartdate',
];

export const MILESTONE_CORE_ATTRS = [ // or MILESTONE_NON_DATE_ATTRS?
 'statuscode',
 '_dcp_milestone_value',
]

export const MILESTONE_ATTRS = [
   ...MILESTONE_DATE_ATTRS,
   ...MILESTONE_CORE_ATTRS,
];
```

Then import the MILESTONE_CORE_ATTRS array in this file.
```
import { MILESTONE_ATTRS, MILESTONE_CORE_ATTRS } from './projects.attrs';
```

godfreyyeung (approved):
Nice! Both the code itself and pr description were very clear! Made one optional suggestion

server/src/projects/projects.service.ts
trbmcginnis: Good idea! I'll make those changes

PR #934, opened 2021-01-08
13578 Improve Save and Submit UX and implementation 

### LINKED ISSUE #2: Create JSON API query that obtains a contact's projects and PAS statuses
Starting with a user's contact ID, try to create the following chained, joined query

- Query the dcp_projectapplicant table to get the contact's assigned project IDs
- From dcp_project table, get project visibility and statuscode=active
- Get the associated PAS packages
- Get the status and visibility of the PAS packages

contact -> projectapplicantteam -> project -> paspackage

Final query: 

```
https://dcppfsuat2.crm9.dynamics.com/api/data/v9.1/dcp_projects?$filter=dcp_dcp_project_dcp_projectapplicant_Project/any(o:o/_dcp_applicant_customer_value eq  'A42C6343-455A-EA11-A9A8-001DD8308047')&$expand=dcp_dcp_project_dcp_package_project($filter=statecode eq 0),dcp_dcp_project_dcp_projectapplicant_Project($filter=statecode eq 0)
```

godfreyyeung:
The Save and Submit functionality faced a couple problems, which needed to be teased apart:

1.  **The attachments component is adding `filesToUpload` files to the `package.documents` array.** This is likely due to the need to validate the number of documents to upload ( >= 1), but can cause problems if a package is saved and the documents property is serialized. When the user selects files for upload, they are represented as `ember-file-upload` `fileQueue` objects. These complex fileQueue objects are being added to the `package.documents` array alongside our custom document objects (simply consisting of `name`, `timeCreated`, and `relativeUrl` properties). When the `fileQueue` objects are serialized, they error out because these objects are recursive.

    The error yielded is `TypeError: Cannot convert circular structure to JSON`

    For now, this. refactor needs to be handled in another ticket/PR. This PR creates the scaffolding to do so. For starters, it prevents the TypeError by fix #2 below.  

2. **The package model is trying to write to the `documents` property upon being saved.**  This should not be the case because the documents property is read-only. `ember-file-upload` handles the uploading of files (performing PATCH requests to the applicant portal API), while Ember Data handles retrieving the most up-to-date list of package documents from the API (sending GET package requests, which include the documents property).  

    - **FIX:** **This PR prevents the troublesome writing to that property by customizing the Package Adapter's `updateRecord` method.** For Package models, the `updateRecord` method will now empty out the `documents` property in the Snapshot of the Package.
  
      A snapshot is like a frozen image of a model, and is used to help avoid issues with async/promises when serializing models. You can read a little more about snapshots here: https://davidtang.io/2016-02-27-what-are-ember-data-snapshots/
      
      I did try completely `delete`-ing the documents property from the snapshot, but some other opaque part of the Ember Data `save` process still tried to reference the documents property, so would yield an error.
      
3. **Errors to file upload were not displayed to the user because the POST operation is not through Ember Data.** Since the POST requests for file upload are generated by ember-file-upload instead of a `Model.save()`,  any returned server errors are not hydrating/injected into the Package Model's `adapterError` property. The `*-error` handlebars components checks the `adapterError` property on Packages and Forms to determine whether to show an error box to the user.
    - **FIX:**  T**his PR adds a `fileUploadErrors` property to Packages, to mimic the adapterError property.** This fileUploadErrors property specifically tracks server errors yielded from the file upload process. The `*-error` handlebar components for all forms were also updated to check and display errors from this property.
  
      (I did try tracking the adapterErrors property, but that yielded errors. Probably because Ember did not intend for it to be directly set).  
    
4. **A file upload error would stop the form and package from saving.** This was a bad user experience because users were not made aware their upload failed (due to problem #3), and would lose all their entered form information upon closing the window. 
    - **FIX:**  **To better assist the user, this PR ensures that file, form, and package save processes will fail independently, and not affect one another.** This is implemented by wrapping each save process in its own try/catch statement. This way, files will still upload if there is a form error, and vice versa. 
  
      Wrapping each save process in a try/catch had another benefit: it prevented errors from being "hidden" by a previous error. It may be that multiple of the save processes would fail, but only the first one is captured by the outermost try/catch in the `@savePackage` action of the form `edit.js` file.
  
      The consequence of `try/catch`-ing each save is that execution continues, so the save process must take ownership of checking for error before proceeding with further operations. More on this in point #5 and #6. Another unfortunate side effect is the need to place a band-aid patch in mirage, described in #7. 
    
5. **The Submit process relied on an uncaught error to stop submission and page transition. With the previous changes, it meant submission continues despite an error.** This problem revealed itself thanks to our test suite.  The Submit process calls Save, and then calls a page transition. However, previously there were zero try/catch statements in entire Save process. Submission would stop if an error occurred anywhere.  Now that we use try/catch statements around parts of the Save process, submission and page transition will continue even if there are errors.

    - **FIX:**  **This PR ensures that the Submit process checks for package, form and file `adapterError`s before transitioning to the Show page.** On error, the user will stay on the page and can inspect the error box for any errors.
    
6. **Similarly, due to the independent failures, the Save process continued to reload the Package model despite errors.  The model reload would reset the Package and Form adapterError.**

    - **FIX:**  **This PR ensures that the Save process checks for package, form and file `adapterError`s before reloading the Package model and synchronizing the package fileManager to the documents property.**
    

7. For some reason, the `updateRecord` method on the Package Adapter is called in Mirage when performing resource GET. Due to the modification of the Snapshot, it's clearing out the `package.documents` property returned from the Mirage server. When a package is loaded from the mmirage server, it doesn't contain any documents.
    - **TEMPORARY FIX:**  **In `mirage/config.js`, for the `get('/packages/:id')` endpoint, I've assigned a hardcoded array of documents to  `package.documents` to satisfy our tests.**  This will simulate a package GET response from CRM. The document names match that of the documents hardcoded in the Package Factory's `withExistingDocuments` trait. This should be revisited and investigate why updateRecord, which sends data to the server, would affect resource retrieval.

### DIFF
--- a/client/app/adapters/application.js
+++ b/client/app/adapters/application.js
@@ -1,6 +1,21 @@
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
 import { inject as service } from '@ember/service';
 import ENV from '../config/environment';
+import PackageModel from '../models/package';
+
+// Taken from
+// https://github.com/emberjs/data/blob/v3.24.0/packages/adapter/addon/-private/utils/serialize-into-hash.js
+function serializeIntoHash(store, modelClass, snapshot, options = { includeId: true }) {
+  const serializer = store.serializerFor(modelClass.modelName);
+
+  if (typeof serializer.serializeIntoHash === 'function') {
+    const data = {};
+    serializer.serializeIntoHash(data, modelClass, snapshot, options);
+    return data;
+  }
+
+  return serializer.serialize(snapshot, options);
+}
 
 export default class ApplicationAdapter extends JSONAPIAdapter {
   @service
@@ -16,5 +31,25 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
     }
 
     return {};
+  }
+
+  // Delete the documents array from Snapshot because
+  // package.documents is read-only.
+  // Discarding it from the update prevents
+  // serialization errors that occur when file upload fails.
+  // When file upload fails, recursive ember-file-upload document objects
+  // are inadvertently placed into the package.documents array.
+  // These recursive objects cannot be serialized, so yield serialization
+  // errors.
+  updateRecord(store, type, snapshot) {
+    if (type === PackageModel) {
+      snapshot.__attributes.documents = [];
+      snapshot._changedAttributes.documents = [];
+    }
+    const data = serializeIntoHash(store, type, snapshot);
+
+    let url = this.buildURL(type.modelName, snapshot.id, snapshot, 'updateRecord');
+
+    return this.ajax(url, 'PATCH', { data });
   }
 }
--- a/client/app/adapters/application.js
+++ b/client/app/adapters/application.js
@@ -1,21 +1,6 @@
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
 import { inject as service } from '@ember/service';
 import ENV from '../config/environment';
-import PackageModel from '../models/package';
-
-// Taken from
-// https://github.com/emberjs/data/blob/v3.24.0/packages/adapter/addon/-private/utils/serialize-into-hash.js
-function serializeIntoHash(store, modelClass, snapshot, options = { includeId: true }) {
-  const serializer = store.serializerFor(modelClass.modelName);
-
-  if (typeof serializer.serializeIntoHash === 'function') {
-    const data = {};
-    serializer.serializeIntoHash(data, modelClass, snapshot, options);
-    return data;
-  }
-
-  return serializer.serialize(snapshot, options);
-}
 
 export default class ApplicationAdapter extends JSONAPIAdapter {
   @service
@@ -31,25 +16,5 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
     }
 
     return {};
-  }
-
-  // Delete the documents array from Snapshot because
-  // package.documents is read-only.
-  // Discarding it from the update prevents
-  // serialization errors that occur when file upload fails.
-  // When file upload fails, recursive ember-file-upload document objects
-  // are inadvertently placed into the package.documents array.
-  // These recursive objects cannot be serialized, so yield serialization
-  // errors.
-  updateRecord(store, type, snapshot) {
-    if (type === PackageModel) {
-      snapshot.__attributes.documents = [];
-      snapshot._changedAttributes.documents = [];
-    }
-    const data = serializeIntoHash(store, type, snapshot);
-
-    let url = this.buildURL(type.modelName, snapshot.id, snapshot, 'updateRecord');
-
-    return this.ajax(url, 'PATCH', { data });
   }
 }
--- a/client/app/adapters/package.js
+++ b/client/app/adapters/package.js
@@ -1,0 +1,53 @@
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import { inject as service } from '@ember/service';
+import ENV from '../config/environment';
+
+// Taken from
+// https://github.com/emberjs/data/blob/v3.24.0/packages/adapter/addon/-private/utils/serialize-into-hash.js
+function serializeIntoHash(store, modelClass, snapshot, options = { includeId: true }) {
+  const serializer = store.serializerFor(modelClass.modelName);
+
+  if (typeof serializer.serializeIntoHash === 'function') {
+    const data = {};
+    serializer.serializeIntoHash(data, modelClass, snapshot, options);
+    return data;
+  }
+
+  return serializer.serialize(snapshot, options);
+}
+
+export default class PackageAdapter extends JSONAPIAdapter {
+  @service
+  session;
+
+  host = ENV.host;
+
+  get headers() {
+    if (this.session.isAuthenticated) {
+      return {
+        Authorization: `Bearer ${this.session.data.authenticated.access_token}`,
+      };
+    }
+
+    return {};
+  }
+
+  // Delete the documents array from Snapshot because
+  // package.documents is read-only.
+  // Discarding it from the update prevents
+  // serialization errors that occur when file upload fails.
+  // When file upload fails, recursive ember-file-upload document objects
+  // are inadvertently placed into the package.documents array.
+  // These recursive objects cannot be serialized, so yield serialization
+  // errors.
+  updateRecord(store, type, snapshot) {
+    snapshot.__attributes.documents = [];
+    snapshot._changedAttributes.documents = [];
+
+    const data = serializeIntoHash(store, type, snapshot);
+
+    let url = this.buildURL(type.modelName, snapshot.id, snapshot, 'updateRecord');
+
+    return this.ajax(url, 'PATCH', { data });
+  }
+}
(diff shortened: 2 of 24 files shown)


client/app/adapters/application.js
```diff
@@ -17,4 +32,24 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
 
     return {};
   }
+
+  // Delete the documents array from Snapshot because
+  // package.documents is read-only.
+  // Discarding it from the update prevents
+  // serialization errors that occur when file upload fails.
+  // When file upload fails, recursive ember-file-upload document objects
+  // are inadvertently placed into the package.documents array.
```
godfreyyeung, line 41: Actually ember-file-upload documents are placed into the package.documents array by design, by the Attachments component

client/app/adapters/application.js
```diff
@@ -17,4 +32,24 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
...
+  // These recursive objects cannot be serialized, so yield serialization
+  // errors.
+  updateRecord(store, type, snapshot) {
```
allthesignals, line 44: Can we put this in a separate adapter file for packages?

client/app/adapters/application.js
allthesignals: Interesting

client/app/models/package.js
```diff
@@ -24,6 +25,14 @@ export default class PackageModel extends Model {
     }
   }
 
+  // Since file upload doesn't perform requests through
+  // an Ember Model save() process, it doesn't automatically
+  // hydrate the package.adapterError property. When an error occurs
+  // during upload we have to manually hydrate a custom error property
+  // to trigger the error box displayed to the user.
```
allthesignals, line 32: Then why do we need to change the adapter?

client/app/models/package.js
```diff
@@ -90,30 +99,58 @@ export default class PackageModel extends Model {
...
+      await super.save();
+    } catch (e) {
+      console.log('Error saving package: ', e);
     }
-    if (this.dcpPackagetype === DCPPACKAGETYPE.DRAFT_SCOPE_OF_WORK.code) {
-      await this.saveDirtySingleCeqrInvoiceQuestionnaire();
+
+    try {
+      await this.fileManager.save();
+    } catch (e) {
```
allthesignals, line 137: Why do we need 3 separate try-catches?

allthesignals (commented):
It's true that the way doc upload works with models is confusing: we have the changeset (which doesn't update itself when the model rehydrates through successful adapter changes), the model itself, and ember-file-upload. FileUpload is triggered as its own system through the save mechanism... and when IT fails, those failures don't get communicated through the adapter system.

I don't like needing this special `fileUploadErrors` property because there's a lot of clearer error handling. You even have to manage state on it. Can you point to some precedent for this? Is there error handling in the ember-file-upload addon we use?

allthesignals:
>  For some reason, the updateRecord method on the Application Adapter is called in Mirage when performing resource GET

Are you sure this isn't a more internal method that's "updating" the in-memory record (hydrating) with data that comes back from the ajax request?

client/app/adapters/application.js
godfreyyeung: Hmm let me try

client/app/models/package.js
godfreyyeung: Changing the adapter fixes a separate issue of package.documents getting serialized and sent to the server (#2) . Whereas this just makes sure that if file upload fails, the package has a property to indicate a file upload failure. Like how the adapterError property indicates a package save error. See

client/app/models/package.js
godfreyyeung: I wanted to improve the user experience of Save. Previously if a form save() failed, the package and files would not save, and vice versa. I wanted to make it so that if a form save failed, files would still upload.  The try/catch around each save() makes sure that execution continues even if a failure occurs for a previous save().

trbmcginnis: Are you saying that if a form save fails (e.g. a user entered something into a text input and that value was NOT PATCHed to the database) that you still want files to be uploaded? Can you explain more about why we want this behavior?

If another part of the form fails to save, wouldn't we also want the document upload to fail simultaneously? Everything functioning in sync

godfreyyeung: Yeah, that is a very good counterpoint. Here are the pros and cons I can think of for allowing execution to continue after an error with one of the three saves.

(Note that with this PR, the current order of operations is.. 
     1) form saves 2) package saves 3) files save
Previously, it was 
     1) files save 2) form saves 3) package save

I re-ordered the operations to match the order in the UI)

Pros: 
   1) the user would find it less bothersome to have to re-select many files if an error occurs on form or package save. 
   2) without the try/catch, it's possible that  a partial save occurs anyway. E.g. the first save operation for forms could have succeeded, but then the second one for files fails. With the try/catch, it just means that order doesn't affect what gets saved and what doesn't.

Cons:
  1) This introduces a lot of extra error handling code, as evidenced in this PR. We have to manually prevent some operations from executing if any errors occurred.

client/app/adapters/application.js
godfreyyeung: It worked!

godfreyyeung:
> It's true that the way doc upload works with models is confusing: we have the changeset (which doesn't update itself when the model rehydrates through successful adapter changes), the model itself, and ember-file-upload. FileUpload is triggered as its own system through the save mechanism... and when IT fails, those failures don't get communicated through the adapter system.
> 
> I don't like needing this special `fileUploadErrors` property because there's a lot of clearer error handling. You even have to manage state on it. Can you point to some precedent for this? Is there error handling in the ember-file-upload addon we use?

@allthesignals  Oops sorry I missed responding to this question. I think the file-manager and attachments component itself is a precedent. It already has to manually track some state like the number of files selected for upload, to determine whether the attachments section "isDirty".

Good point about potentially using some addon error handling. In their recipes they rely on promise then/catch. However, I realized I could potentially iterate through the file queue and check if any of the files are in a "failed" state. I'm not sure if I could make this "reactive" though, because the file queue may be an opaque non-Ember object/array. It'll take me some more experimenting... 

Otherwise, I haven't thought of a better way. `*-error.hbs` components are set up to display an error if one exists on the adapterError property on the models. The `fileUploadError` property is following that pattern closely

client/app/adapters/package.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // Delete the documents array from Snapshot because
+  // package.documents is read-only.
+  // Discarding it from the update prevents
+  // serialization errors that occur when file upload fails.
+  // When file upload fails, recursive ember-file-upload document objects
+  // are inadvertently placed into the package.documents array.
+  // These recursive objects cannot be serialized, so yield serialization
+  // errors.
```
trbmcginnis, line 42: Thanks for these detailed comments!

godfreyyeung:
Per standup discussion with @allthesignals  and @trbmcginnis , since this seems to be adding more complication to code that's already a bit brittle, the ideal approach would be to take time soon to rip out ember-file-upload and redo the implementation.

For now, it's not ideal but we agreed to merge this to move forward.

PR #938, opened 2021-01-19
777: redirect city employees to nyc.id sign-in page regardless of registration status

trbmcginnis:
### Summary
If a user is a City Employee they should be redirected to the NYC.ID sign-in page regardless of their NYC ID registration status. This PR updates the rerouting logic to always send City Employees to the sign-in page. 

Just noting some extra information for documentation & research purposes based on email exchange with NYC.ID folks: 

- They confirmed that a city employee will NEVER register on the "register" page that we were originally sending non-registered city employees to. Instead they should always be redirected to the NYC.ID login page where they can select the NYC Employee option button. More information at https://www1.nyc.gov/assets/nyc4d/html/services-nycid/registration.shtml 
- They suggested that we use `GET /account/api/oauth/user.htm` instead of `GET', '/account/api/user.htm` so that we could actually receive information about whether that email existed rather than relying on an "Unauthorized" error. This issue with this is that regardless of whether the city employee's email is registered with NYC.ID or not, we get the same error of `{ accessToken: 'required' }` for both. Whereas when we do not use the `oath` method in the `GET` we're able to differentiate between the two emails' statuses -- with one having an "UnknownEmail" error and the other having an "Unauthorized" error. So while it might still be a workaround, it's better for us to use `GET', '/account/api/user.htm`.

#### Task/Bug Number
Fixes [AB#777](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/777)

### DIFF
--- a/client/app/components/auth/sign-in.js
+++ b/client/app/components/auth/sign-in.js
@@ -28,16 +28,22 @@ export default class AuthSignInComponent extends Component {
       isCityEmployee,
     } = await this.args.searchContacts(loginEmail);
 
-    if (isNycidEmailRegistered) {
-      // we can only know email validation after they logged in once.
-      // if it's null, it means we don't know because they've never logged in before (null).
-      if (isNycidValidated === true || isNycidValidated === null) {
-        this.router.transitionTo('auth.login', { queryParams: { loginEmail, isCityEmployee } });
-      } else if (isNycidValidated === false) {
-        this.router.transitionTo('auth.validate', { queryParams: { loginEmail } });
-      }
+    // We should always be redirecting nyc government employees to the NYC.ID login page
+    // (never to the register page) regardless of whether isNycidEmailRegistered is true or false
+    if (isCityEmployee) {
+      this.router.transitionTo('auth.login', { queryParams: { loginEmail, isCityEmployee } });
     } else {
-      this.router.transitionTo('auth.register', { queryParams: { loginEmail } });
+      if (isNycidEmailRegistered) { // eslint-disable-line
+        // we can only know email validation after they logged in once.
+        // if it's null, it means we don't know because they've never logged in before (null).
+        if (isNycidValidated === true || isNycidValidated === null) {
+          this.router.transitionTo('auth.login', { queryParams: { loginEmail } });
+        } else if (isNycidValidated === false) {
+          this.router.transitionTo('auth.validate', { queryParams: { loginEmail } });
+        }
+      } else {
+        this.router.transitionTo('auth.register', { queryParams: { loginEmail } });
+      }
     }
   }
 }
--- a/server/src/contact/nycid/nycid.service.ts
+++ b/server/src/contact/nycid/nycid.service.ts
@@ -29,7 +29,7 @@ export class NycidService {
       // check if the e-mail is validated in nyc.id's system
       ...isNycidValidated,
 
-      is_city_employee: isNycidEmailRegistered && isNycidValidated && email.endsWith('nyc.gov'),
+      is_city_employee: email.endsWith('nyc.gov'),
     };
   }
 


trbmcginnis:
Depending on whether goverment employees have signed in to NYC.ID and "finished creating their account", they will either have `isNycidEmailRegistered` equal to true or false. This PR updates the rerouting logic to accommodate this difference. 

nyc_id_changes

godfreyyeung:
So if an email is not registered yet but a "city employee" email, users should still be able to sign in successfully to NYCID? 

Also, I have a suspicion the root problem is the isCityEmployee value is faulty.  It looks like is_city_employee will be false if `isNycidValidated` is false. However, when there is no CRM contact, `isNycidValidated` will always be false. So even if the email has `.gov`, `is_city_employee` would be false

allthesignals:
> So is it confirmed that if an email is not registered yet but a "city employee" email, users should still be able to sign in successfully to NYCID?
> 
> Also, I have a suspicion there's still a problem with `isCityEmployee`/`is_city_employee` value being faulty. In the `getNycidStatus` function in the NycId service, the derived `is_city_employee` property depends on the `isNycidValidated` value. However, `isNycidValidated` will always be `true` since it is an object. I think this was a mistake. It should probably be
> 
> ```
> is_city_employee: isNycidEmailRegistered && isNycidValidated.is_nycid_validated && email.endsWith('nyc.gov'),
> ```

Why is `isNycidValidated` an object at all? The naming convention there suggests a boolean. The solution should be to fix `isNycidValidated` so that it works as it's semantically implied to work.

client/app/components/auth/sign-in.js
```diff
@@ -28,7 +28,9 @@ export default class AuthSignInComponent extends Component {
       isCityEmployee,
     } = await this.args.searchContacts(loginEmail);
 
-    if (isNycidEmailRegistered) {
+    // We should always be redirecting nyc government employees to the NYC.ID login page
+    // (never to the register page) regardless of whether isNycidEmailRegistered is true or false
+    if (isNycidEmailRegistered || isCityEmployee) {
```
allthesignals, line 33: I'm wondering if this is adding complexity that will bite is later. Can't we assume that all city employees are registered and validated, and route them to NYC.ID login? 

I would add something about the if: 
```
if (isCityEmployee) {
  // transition to the auth.login route.
}
```

allthesignals (approved):
I have a suggestion that I think should be considered. It would be less instrusive to the existing code. Let me know what you think. Approving for simplicity.

trbmcginnis:
> So is it confirmed that if an email is not registered yet but a "city employee" email, users should still be able to sign in successfully to NYCID?
> 
> Also, I have a suspicion there's still a problem with `isCityEmployee`/`is_city_employee` value being faulty. In the `getNycidStatus` function in the NycId service, the derived `is_city_employee` property depends on the `isNycidValidated` value. However, `isNycidValidated` will always be `true` since it is an object. I think this was a mistake. It should probably be
> 
> ```
> is_city_employee: isNycidEmailRegistered && isNycidValidated.is_nycid_validated && email.endsWith('nyc.gov'),
> ```

No, it's not officially confirmed that a "city employee" email should still be able to sign in if not registered. This is only something I've seen by testing it myself, but maybe I should reach out to DOITT to see if that's something that they can confirm.

trbmcginnis:
> > So is it confirmed that if an email is not registered yet but a "city employee" email, users should still be able to sign in successfully to NYCID?
> > Also, I have a suspicion there's still a problem with `isCityEmployee`/`is_city_employee` value being faulty. In the `getNycidStatus` function in the NycId service, the derived `is_city_employee` property depends on the `isNycidValidated` value. However, `isNycidValidated` will always be `true` since it is an object. I think this was a mistake. It should probably be
> > ```
> > is_city_employee: isNycidEmailRegistered && isNycidValidated.is_nycid_validated && email.endsWith('nyc.gov'),
> > ```
> 
> Why is `isNycidValidated` an object at all? The naming convention there suggests a boolean. The solution should be to fix `isNycidValidated` so that it works as it's semantically implied to work.

Okay cool, I can make this change

client/app/components/auth/sign-in.js
trbmcginnis: We can't assume that all city employees are registered, because one of the government emails that I tested that was a City Employee was not registered in NYC ID. And my own government email was not considered "registered" until I went to NYC ID, signed in, and then clicked "Submit" on a "finish creating account" page that I posted in a screenshot above ^^. I can reach out to DOITT to confirm these assumptions before implementing this change. I only discovered this logic by testing the process myself. 

So you're saying that if a user is a City Employee we should never send them to the `auth.validate` route but ONLY send them to the `auth.login` route?

allthesignals: Yeah — if it's an "unregistered" city employee, don't they just have to be directed to the login page?

godfreyyeung (approved):

trbmcginnis:
Just noting this for documentation purposes. When a user enters a dummy government email such as `bananas@planning.nyc.gov` they are met with the same response in regards to 1) registration 2) validation and 3) city employee email status (ends with `.nyc.gov`) that an unregistered ACTUAL city employee would be met with. So as of right now with the way we have this set up, there's no way to tell whether a user has entered a dummy email. We can try to address this in the future but as of right now this dummy email will be 1) rerouted to the nyc.id login page 2) if they try to sign in nyc.id will not let them/will tell them that they need to create an account. This feels fine for now compared to the bug that was fixed with this PR. 
dummy_email
government_email_response

PR #945, opened 2021-02-03
825: disallow users to add the same bbl twice

trbmcginnis:
### Summary
Previously, on the Land Use Form for the Project Area Tax Lot section, users were able to add duplicate BBLs. This should not be allowed. This PR makes it so that when a user tries to add a BBL that already exists on the project a warning message is displayed saying "The BBL {bbl number} has already been added to this project. Please try again."

#### Tasks/Bug Numbers
 - Fixes [AB#825](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/825)

### Technical Explanation
- There is an action that is triggered when a user selects an option from the geosearch dropdown for addresses/BBLs. This action uses `store.createRecord` to create new BBLs. If the selected option is a duplicate, this `createRecord` will not run
- Instead, when the user selects an option from the dropdown that already exists in the `store`, two `tracked` properties are updated. `bblAlreadyCreated` is set to `true` for the purpose of displaying the warning message on the template. And `duplicateBblNumber` is set to the value fo the current bbl number, which is displayed in the text of the warning message.

### DIFF
--- a/client/app/components/packages/project-geography.js
+++ b/client/app/components/packages/project-geography.js
@@ -1,6 +1,7 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
 import { bblBreakup } from '../../helpers/bbl-breakup';
 import { BOROUGHS } from '../../optionsets/bbl';
 
@@ -19,8 +20,12 @@ export default class ProjectGeographyComponent extends Component {
   @service
   store;
 
+  @tracked bblAlreadyCreated = false;
+
+  @tracked duplicateBblNumber = '';
+
   @action
-  selectSearchResult(labsSearchResult) {
+  async selectSearchResult(labsSearchResult) {
     // `labsSearchResult` is object with `label` (address) and `bbl` attributes.
     // labsSearchResult.bbl comes back as a number,
     // whereas dcpBblnumber in CRM is stored as a string.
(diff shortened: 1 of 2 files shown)


trbmcginnis:
bbl_already_added

client/app/components/packages/project-geography.js
```diff
@@ -19,35 +20,49 @@ export default class ProjectGeographyComponent extends Component {
   @service
   store;
 
+  @tracked bblAlreadyCreated = false;
```
allthesignals, line 23: Can you think of another approach that doesn't require adding more state to the component?

For example, your code in the selectSearchResult method could trigger a call to `alert` with a simple message.

allthesignals (approved):
I think this is okay but is using our form field validation setup not a possibility? IIRC, we should have code for each section that communicates a validation error.

trbmcginnis:
> I think this is okay but is using our form field validation setup not a possibility? IIRC, we should have code for each section that communicates a validation error.

I agree about reducing how much state we add to the component. I know that ember-changeset-validations has a way to create your own validation functions but I figured that since this involves checking the `store` in the component that it would be out of the scope of ember-changeset-validations. We would also have to make sure that this validation was imported into the project-geography.js in order to avoid creating a new record for the entered BBL. 

Let me investigate some more into whether we can create our own validation function for this with ember-changeset and see how extensive we can make these functions and I'll document my findings here.

allthesignals:
> > I think this is okay but is using our form field validation setup not a possibility? IIRC, we should have code for each section that communicates a validation error.
> 
> I agree about reducing how much state we add to the component. I know that ember-changeset-validations has a way to create your own validation functions but I figured that since this involves checking the `store` in the component that it would be out of the scope of ember-changeset-validations. We would also have to make sure that this validation was imported into the project-geography.js in order to avoid creating a new record for the entered BBL.
> 
> Let me investigate some more into whether we can create our own validation function for this with ember-changeset and see how extensive we can make these functions and I'll document my findings here.

Okay cool.

We validate the length of `applications` here: 

image

So, I bet there's a way to validate the uniqueness of the "newValue", which should come through as an array.

PR #1052, opened 2022-05-04
8426 added guidance copy to PAS equity form below non-res fl area 200K

horatiorosa:
### Summary
Add the below text as additional guidance under question “Increase permitted non-residential floor area by at least 200,000 square feet”
 - utilized svg and tooltip as per gilal designs in figma

Additional copy added in the `equitable-development-reporting.hbs` file

#### Tasks/Bug Numbers
 - Fixes [AB#8426](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8426)

 - Related [AB#8374](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8374)
 
Screen Shot 2022-05-06 at 4 04 02 PM

Screen Shot 2022-05-06 at 4 04 15 PM

### DIFF
--- a/client/app/styles/app.scss
+++ b/client/app/styles/app.scss
@@ -172,3 +172,13 @@ $sticky-sidebar-offset: 6rem + rem-calc(20);
 .dropdown-clear-btn-below {
   color: $red-dark;
 }
+
+// Styling for tooltip "i" icons
+// note: over the top specificity was required to overwrite the fa defaults
+span.ember-tooltip-target > svg.svg-inline--fa.fa-info-circle.fa-w-16.ember-view.ember-tooltip-target {
+  color: #828891;
+  height: 13.33px;
+  width: 13.33px;
+  margin-bottom: 5px;
+}
+
--- a/client/app/components/packages/equitable-development-reporting.hbs
+++ b/client/app/components/packages/equitable-development-reporting.hbs
@@ -74,7 +74,18 @@
       <Ui::Question @required={{false}} as |Q|>
         <Q.Legend>
           Increase permitted non-residential floor area by at least 200,000 square feet
+          <span>
+            <FaIcon @icon="info-circle" @prefix="fas" class="ember-tooltip-target"/>
+            <EmberTooltip @event="hover" @side="right">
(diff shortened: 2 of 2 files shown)


godfreyyeung:
Is this also blocked by tooltip redesign?

godfreyyeung (approved):
Nice! Thanks for bearing through all the changes we talked about. Interesting addition of ember-tooltip-target

client/app/styles/app.scss
```diff
@@ -172,3 +172,13 @@ $sticky-sidebar-offset: 6rem + rem-calc(20);
 .dropdown-clear-btn-below {
   color: $red-dark;
 }
+
+// Styling for tooltip "i" icons
+// note: over the top specifity was required to overwrite the fa defaults
+span.ember-tooltip-target > svg.svg-inline--fa.fa-info-circle.fa-w-16.ember-view.ember-tooltip-target {
```
godfreyyeung, line 178: Interesting. I would almost prefer using `!important` instead of child selector. 
I can't speak to performance of one vs the other, but `!important` seems more absolute  and less flaky.

godfreyyeung: Also I would prefer creating a class like "radio-tooltips" or something

horatiorosa: I avoid using `!important` as bad practice, reference here:  [MDN docs on specificity: !important](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#:~:text=Although%20technically%20!important%20has%20nothing,natural%20cascading%20in%20your%20stylesheets.). I will be using the tooltip in different `.hbd` files in the applicant portal. 

As for the classes used, I selected the one rendered by the `EmberTooltip` component  to achieve the specificity needed to override the default styles.

godfreyyeung: Sounds good. Thanks for sharing that. If you haven't already. I would also suggest looking into using the FaIcon `@ size` attribute first before sizing it in css 
```
@size="4x"
```
your call

horatiorosa: I tried the size attribute previously but I couldn't get it smaller than 1em/16px (used `@size="1x"`)

PR #1107, opened 2022-08-04
8125 additional logging in RWDCS form service to catch errors

horatiorosa:
### Summary
Addinal error logging in RWCDS form service
bug was caused my CMS input field that awsa too long. Attached photos have details.

### app logs with new try / catch blocks detailing error state
zrmodifyingzrtxt_length_error_ado_copy

#### form field that cause above error (after we shortened the text input)
Screen Shot 2022-08-05 at 1 45 37 PM

#### Tasks/Bug Numbers
 - Fixes [AB#8125](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8125)

### DIFF
--- a/server/src/packages/packages.service.ts
+++ b/server/src/packages/packages.service.ts
@@ -151,7 +151,6 @@ export class PackagesService {
         || firstPackage.dcp_packagetype === PACKAGE_TYPE_OPTIONSET['POST_CERT_LU'].code
       ) {
         formData = await this.fetchPackageForm(firstPackage);
-        console.log('from package.service.ts - firstPackage', firstPackage);
       }
 
       // below query filters by Racial Equity Report file type. 
--- a/server/src/packages/packages.service.ts
+++ b/server/src/packages/packages.service.ts
@@ -151,6 +151,7 @@ export class PackagesService {
         || firstPackage.dcp_packagetype === PACKAGE_TYPE_OPTIONSET['POST_CERT_LU'].code
       ) {
         formData = await this.fetchPackageForm(firstPackage);
+        console.log('from package.service.ts - firstPackage', firstPackage);
       }
 
       // below query filters by Racial Equity Report file type. 
@@ -177,7 +178,7 @@ export class PackagesService {
       try {
         firstArtifactWithDocuments = await this.artifactService.artifactWithDocuments(firstArtifactWithDocuments);
       } catch (e) {
-        console.log(e);
+        console.log('firstArtifactWithDocuments error', e);
       }
 
       dcp_project.artifact = firstArtifactWithDocuments;
@@ -196,7 +197,9 @@ export class PackagesService {
     } catch (e) {
       // relay lower-level exceptions, like from crmServce.get(),
       // or sharepoint document retrieval.
+      console.log('dcp_project.artifact error', e);
       if (e instanceof HttpException) {
+        console.log('dcp_project.artifact HttpException', e);
         throw e;
       } else {
         throw new HttpException({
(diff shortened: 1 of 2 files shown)


server/src/packages/packages.service.ts
```diff
@@ -151,6 +151,7 @@ export class PackagesService {
         || firstPackage.dcp_packagetype === PACKAGE_TYPE_OPTIONSET['POST_CERT_LU'].code
       ) {
         formData = await this.fetchPackageForm(firstPackage);
+        console.log('from package.service.ts - firstPackage', firstPackage);
```
godfreyyeung, line 154: you'll delete this eventually, right?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -106,6 +106,12 @@ export class RwcdsFormService {
   }
 
   async find(id) {
+    try {
+
```
godfreyyeung, line 110: wait why is there an empty try block?

server/src/packages/rwcds-form/rwcds-form.service.ts
```diff
@@ -166,52 +171,74 @@ export class RwcdsFormService {
...
+        // IF the zr type does not exist yet, then POST the action
+        // NOTE: Some actions that exist in dcp_projectaction SHOULD NOT be copied over.
+        // The ZONING_RESOLUTION_TYPES lookup also functions as a list of actions that SHOULD be
+        // copied over from dcp_projectaction, which is why we check that `code` exists here.
+        if (!matchingZr && code) {
+          try {
+            return this.crmService.create(`dcp_affectedzoningresolutions`, {
+              'dcp_zoningresolutiontype': code, // coded value
+              'dcp_zrsectionnumber': currentActionZoningResolution,
+              'dcp_modifiedzrsectionnumber': action.dcp_zrmodifyingzrtxt, // text was too long, need to trim to 25 char or less. need to limit text limit in crm
```
godfreyyeung, line 229: no need to include this comment. Let's see how Aline and Liz think we should handle this, possibly on the CRM side

server/src/packages/rwcds-form/rwcds-form.service.ts
horatiorosa: my bad. I didn't move the `get` into the block. fixed.

server/src/packages/rwcds-form/rwcds-form.service.ts
horatiorosa: I spoke to them in chat and on ticket.  Liz is looking into how this happened as there is a text limit in that input field. Aline believed that is a DCP entered field. This moves the focus to the CRM side of things.

server/src/packages/packages.service.ts
```diff
@@ -196,7 +196,9 @@ export class PackagesService {
     } catch (e) {
       // relay lower-level exceptions, like from crmServce.get(),
       // or sharepoint document retrieval.
+      console.log('dcp_project.artifact error', e);
```
godfreyyeung, line 199: this looks like an inaccurate message. The try/catch here wraps the entire procedure to retrieve a package

server/src/packages/packages.service.ts
```diff
@@ -196,7 +196,9 @@ export class PackagesService {
...
       if (e instanceof HttpException) {
+        console.log('dcp_project.artifact HttpException', e);
```
godfreyyeung, line 201: same comment as above

godfreyyeung (approved):

PR #1110, opened 2022-08-08
Bump node 16; upgrade CircleCI configuration

### LINKED ISSUE #1109: Applicant Portal Back End
# This issue is closed when the app builds for Node 14/16

allthesignals:
### Summary

Bump the node version and also use a newer node image in CircleCI.

#### Tasks/Bug Numbers
 - Fixes #1109 

### Technical Explanation

Specifies a node version in a few ways: using .nvmrc and using "engines" key in package.json.

Also bumps the node version in CI.

**Note: I'm not sure how this affects the production environment. If we're still using Heroku, we should check that it uses the right buildpack.**


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   nest_tests:
     docker:
-      - image: circleci/node:10.18-browsers
+      - image: cimg/node:16.15-browsers
```
TylerMatteo, line 5: Are you sure that we need this? I thought circleci files like these are leftovers from when Labs used to use it but doesn't anymore, but you would have more context into the history of that than me.

server/package.json
```diff
@@ -5,6 +5,9 @@
   "author": "",
   "private": true,
   "license": "UNLICENSED",
+  "engines": {
+    "node": ">=16"
```
TylerMatteo, line 9: Just double checking - the `>=` here limits accepted versions to 16.x, correct? As in, 18 wouldn't be excepted based on this, right?

TylerMatteo: ooof, not sure how I made that typo in one place but not the other. In that case, can you change this to just `"^16"`? That _should_ translate to "anything 16.0.0 or above but below 17.0.0". Great to hear 18 works but having a value there that doesn't at least pin the major version seems to be asking for problems down the road.

.circleci/config.yml
TylerMatteo: Hmm interesting. Do you happen to know how where I might find login credentials for circleci? I asked around about it but didn't get any answers.

TylerMatteo: Hmm not sure which GH account that would be but I'm fine with this change in the mean time.

PR #1119, opened 2022-08-30
Merge develop to QA

godfreyyeung:
- node bump
- getcartkey endpoint


server/src/citypay/citypay.service.ts
```diff
@@ -1,120 +1,135 @@
...
-        await page.type('#Password', this.config.get('MS_APPLICANT_PORTAL_PASSWORD'));
-      } catch (e) {
-        await page.type('#PasswordValue', this.config.get('MS_APPLICANT_PORTAL_PASSWORD'));
-      }
+    try {
+      let citypayResponse = null;
 
-      // click signin, but await for the response with cookies
-      page.click('#submit-signin-local');
+      citypayResponse = await axios.post(`${this.config.get('PAYMENT_BASE_URL')}/${this.config.get('PAYMENT_STEP1_URL')}`, params.toString(), {
```
TylerMatteo, line 102: Can you consolidate `PAYMENT_BASE_URL` AND `CITYPAY_CUSTOMER_LINK` into one variable? It looks like they are the same for most of the full path. Related, can you rename `PAYMENT_STEP1_URL`? I know that's what i was called in the C# code but that's not a very descriptive variable name.

server/src/citypay/citypay.service.ts
```diff
@@ -1,120 +1,135 @@
...
-      } else {
-        throw new Error(`Could not create cookie, maybe something is wrong with the username/password? ${e.toString()}`);
-      }
-    } finally {
-      const browser = await browserPromise;
-
-      browser.close();
+      return cartKey;
+    } catch(e) {
+      console.log("error: ", e);
```
TylerMatteo, line 132: Not going to hold up this PR over this but think about how we can integrate Sentry into this work. I'm going to ask that we have any endpoints that we're rewriting hooked up to Sentry before we release this.

server/src/citypay/citypay.service.ts
```diff
@@ -1,120 +1,135 @@
...
 
-      // Something changed MCS-portal side for this selector
-      // the selector changed from Password -> PasswordValue
-      // this code will try both just in case.
-      try {
-        await page.type('#Password', this.config.get('MS_APPLICANT_PORTAL_PASSWORD'));
-      } catch (e) {
-        await page.type('#PasswordValue', this.config.get('MS_APPLICANT_PORTAL_PASSWORD'));
-      }
+    try {
```
TylerMatteo, line 99: Do you think you could break up this try/catch into 2 or 3 smaller ones? I've had trouble in the past debugging Applicant Portal because we do a bunch of stuff in one big try block and it's hard to tell what threw an exception. You could then have each of the smaller catches throw a more description error object with a `code` like we do elsewhere in the app.

godfreyyeung: I decided to do one try/catch here because function call in this try/catch depends on the previous to succeed. If I split it up into different try/catch, then the an exception in the first function calls would not stop execution in subsequent function calls. 

The Applicant Portal troubles with descriptive error message is because I previously set up a poor design of re-wrapping exceptions in nest exceptions and not logging out the original error. 

In this case, the error should be logged with the line of code or specific function that yielded the error.

godfreyyeung: actually now that you mention it... i should move the try even further up...

TylerMatteo (approved):

server/src/citypay/citypay.service.ts
godfreyyeung: I'm going to test this in a future PR because currently I can't test locally due to the node bump

PR #1218, opened 2024-02-22
Add and use new optionset for Type II eligibility question

bmarchena:
This PR solves a bug where users could not save the PAS or RWDCS forms when selecting an answer other than "Unsure" for the Type II eligibility question. It seems our server was expecting different codes than those assigned to the answers in the YES_NO_UNSURE_SMALLINT optionset. This solution adds a new optionset and instead uses those for the question in both forms.

### DIFF
--- a/client/app/optionsets/common.js
+++ b/client/app/optionsets/common.js
@@ -64,6 +64,21 @@ export const YES_NO_UNSURE_SMALLINT = {
   },
 };
 
+export const YES_NO_UNSURE_LONGINT = {
+  YES: {
+    code: 717170000,
+    label: 'Yes',
+  },
+  NO: {
+    code: 717170001,
+    label: 'No',
+  },
+  UNSURE: {
+    code: null,
+    label: 'Unsure at this time',
+  },
+};
+
 export const YES_NO_DONT_KNOW = {
   YES: {
     code: 717170000,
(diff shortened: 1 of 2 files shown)


client/app/optionsets/common.js
```diff
@@ -64,6 +64,21 @@ export const YES_NO_UNSURE_SMALLINT = {
...
+  NO: {
+    code: 717170001,
+    label: 'No',
+  },
+  UNSURE: {
+    code: null,
+    label: 'Unsure at this time',
+  },
+};
+
```
horatiorosa, line 81: @bmarchena  I am wondering why we need to make a new optionset `YES_NO_UNSURE_LONGINT` instead of reusing the `YES_NO_UNSURE` on line 37. They seem to share the same `Yes.code`, `No.code` values the CRM expects in the response.

bmarchena: The problem is that the CRM is expecting a 'null' response for unsure on the eligibility question, so I added the new set rather than modify that existing one for fear of breaking functionality elsewhere

horatiorosa: Oh, my B. I wasn't aware that the CRM required a `null` response  for `UNSURE` rather than the  `YES_NO_UNSURE.UNSURE`value of 71717001 response.

horatiorosa (approved):
🚀

PR #1258, opened 2024-09-13
configure workflow based on zap search

### LINKED ISSUE #1244: Implement deployments and development from main
# Description
Move from `git-flow` to `main based` development. This should streamline feature integrations and releases

## Acceptance criteria
- Client deploys to Netlify
   - [ ] production
   - [ ] staging
   - [ ] qa 
- Server deploys to Heroku
   - [ ] production
   - [ ] staging
   - [ ] qa

### Code reference
Zap Search follows a similar deployment architecture and most of the required configurations are available in [a feature branch](https://github.com/NYCPlanning/labs-zap-search/tree/main/.github/workflows). 

**Note:** *From work done in other workflows since writing the zap-search one, `echo`ing the variables into an `.env` does not seem necessary*

horatiorosa:
#### Tasks/Bug Numbers
 - Closes #1244

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -1,0 +1,31 @@
+name: 🚧 Run code quality checks
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  test-client:
+    name:  🧪 Test client code
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: client
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: install dependencies
+        run: yarn install --frozen-lockfile --non-interactive
+        working-directory: client
+      - name: build
+        env:
+          DISABLE_SOURCE_MAPS: true
+          BROCCOLI_ENV: production
+        run: yarn build
+        working-directory: client
+      - name: test
+        run: yarn test
+        working-directory: client
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,0 +1,100 @@
+name: 🌠 Deploy to production
+
+on: workflow_dispatch
+
+jobs:
+  deploy-server:
+    name: 🚀 Deploy server
+    environment: 
+      name: production
+      url: https://applicants-api.nycplanningdigital.com
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: server
+      - uses: akhileshns/heroku-deploy@v3.13.15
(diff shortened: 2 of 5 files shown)


.github/workflows/production.yml
```diff
@@ -0,0 +1,100 @@
...
+      - uses: akhileshns/heroku-deploy@v3.13.15
+        name: Deploy server to Heroku
+        with:
+          heroku_email: ${{secrets.HEROKU_EMAIL}}
+          heroku_api_key: ${{secrets.HEROKU_API_KEY}}
+          heroku_app_name: ${{ vars.HEROKU_APP_NAME }}
+          team: ${{secrets.HEROKU_TEAM}}
+          appdir: server
+        env:
+          HD_ADO_PRINCIPAL: ${{ secrets.ADO_PRINCIPAL }}
```
TylerMatteo, line 25: What does the `HD` stand for?

.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,31 @@
+name: 🚧 Run code quality checks
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
```
TylerMatteo, line 9: @TangoYankee @horatiorosa Maybe this is something we visit but would it be difficult to move these code quality checks into the workflows for deploying the environments? I much prefer things like code quality and tests being in the same workflow as deployments, as separate "jobs". That way you have the option of blocking the deployment if the code quality checks fail. We use this approach in [EDDE's workflows](https://github.com/NYCPlanning/equity-tool/blob/main/.github/workflows/feature_branch.yml).

.github/workflows/production.yml
TangoYankee: Heroku Deployment I think? It's from the action we're using

https://github.com/marketplace/actions/deploy-to-heroku#environment-variables

.github/workflows/code_quality.yml
TangoYankee: yes and no to the difficulty question. We can just copy and paste the code quality checks into each environment workflow.

Long term, we should make the jobs reusable, otherwise we end up needing to update code across several workflows every time we make an updated. Of course, this goes for code quality and deployment jobs (we're mostly reusing code across the deployment workflows, too)

For now, would you like to go forward with copying and pasting code quality checks into each deployment workflow?

TylerMatteo: Yes please.

TylerMatteo (dismissed):

TangoYankee (changes requested):
A safety measure to make sure we don't merge

TangoYankee:
Closing because we put these changes on the "default" branch by making `main` the "default"

PR #1263, opened 2024-09-23
Feature flag for Self Service portal

### LINKED ISSUE #1242: Feature Flag for New Projects
## Acceptance criteria
- [x] A feature flag is created to toggle "New project" feature
- [x] Feature is opt-in
- [x] Feature is activated with Environment Variable

TangoYankee:
Protect server and client self service feature with flag

closes #1242

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -45,6 +45,7 @@ jobs:
           team: ${{secrets.HEROKU_TEAM}}
           appdir: server
         env:
+          HD_FEATURE_FLAG_SELF_SERVICE: ${{ vars.FEATURE_FLAG_SELF_SERVICE }}
           HD_ADO_PRINCIPAL: ${{ secrets.ADO_PRINCIPAL }}
           HD_AUTHORITY_HOST_URL: ${{ secrets.AUTHORITY_HOST_URL }}
           HD_CITYPAY_AGENCYID: ${{ secrets.CITYPAY_AGENCYID }}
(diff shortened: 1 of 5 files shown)


.github/workflows/production.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           team: ${{secrets.HEROKU_TEAM}}
           appdir: server
         env:
+          HD_FEATURE_FLAG_SELF_SERVICE: ${{ vars.FEATURE_FLAG_SELF_SERVICE }}
```
pratishta, line 48: Is this environment variable meant to be set in `development.env`?

TangoYankee: It depends. We use the same `FEATURE_FLAG_SELF_SERVICE` variable in both the client and server. When running locally, the server's instance of `FEATURE_FLAG_SELF_SERVICE` is set in `server/development.env`. However, the client's instance would be set in a `client/.env` file or by prefixing the run command with the variable, a la:
```sh
FEATURE_FLAG_SELF_SERVICE=ON yarn run start:ssl
```

This particular file is for running GitHub actions. For actions, we set the environment variables as repository settings. The `workflows/production.yml` file pulls these variables from the repository settings and passes them to the action environments where the client and server are being built and deployed. 

*Ignore the `HD_` prefix. It's a quirk of the Heroku action and the `HD_` is dropped when actually passed to the application.*

horatiorosa (approved):
great job. kudos to you sir.

PR #1359, opened 2025-05-05
Sync ZAP and NYC ID email addresses

TangoYankee:
Sync email addresses when logging in

closes NYCPlanning/ae-private#81


TylerMatteo (approved):
Lgtm but we should think about how to test this thoroughly. We need to make sure it won't cause any regressions.

TangoYankee:
> Lgtm but we should think about how to test this thoroughly. We need to make sure it won't cause any regressions.

If we feel more tested is required, I'd like to deploy this feature branch to the QA environment and test it before merging to main. Should we recruit the QA team to look it over.

TylerMatteo (approved):

